PR #1, opened 2022-05-04
Add CRC + download_tools helper and make targets to created service instances

stuggi:
splits the repo in devsetup and os (OpenStack) part. devsetup
helps to setup CRC environment and download_tools to install all
required operator dev tools with the correct versions.

Also adds support to create instances of mariadb and keystone.


os/scripts/gen-keystone-kustomize.sh
```diff
@@ -0,0 +1,94 @@
...
+
+CR_DIR=out/${NAMESPACE}/keystone/cr
+
+if [ ! -d ${CR_DIR} ]; then
+      mkdir -p ${CR_DIR}
+fi
+
+pushd ${CR_DIR}
+
+cat <<EOF >kustomization.yaml
```
dprince, line 46: These files are quite verbose. I've been using a combined k8s secret for devel (essentially sharing one secret for multiple services). Not recommended for production of course but having a HERE doc that spits out one k8s secret for all the things is much simpler for dev environments.

And with regards to kustomize I think the way I was thinking about using it was more of a we have a set of defaults everyone uses. But if I want to have my own defaults I create my own branch in my fork, and can go do whatever I want and layer over things with kustimize. Things like custom images, or whatever. That way we don't have to maintain shell scripts with options for all the things. Just a thought on the pattern we develop for this. Essentially trying to keep the set of scripts to something quite minimal

stuggi: yes we could do that. I did something similar to create multiple secrets, but for the password they e.g default all to the same password passed in via the Makefile. I have not checked the cnosp operators to what they expect to be in that secret. we might have to extend that this single secret holds all the information a controller expects. like rabbit, passwords to connect to another OSP service. 

I was thinking for the CMs/Secrets the operators require for input, that it is probably better to use labels to get the object instead of the CM/secret name itself. with this its possible to have a combined object holding information for multiple controllers.

stuggi:
closing in favor of https://github.com/openstack-k8s-operators/install_yamls/pull/3

PR #3, opened 2022-05-13
Deploy services using service operator samples

stuggi:
Checks out the operator source and use the sample
combined with kustomize to create an instance of
the service.

A single secret gets created to have all the passwords
neede by the operators.

Kustomize is used to only set the secret and namespace.

All other changes can be done via a branch of the
operator.

For each service there is <service>_deploy_prep,
<service>_deploy and <service>_deploy_cleanup.

In addition a global `make cleanup` got added to
remove all service objects.

### DIFF
--- a/scripts/gen-service-kustomize.sh
+++ b/scripts/gen-service-kustomize.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+#
+# Copyright 2022 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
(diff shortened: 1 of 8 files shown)


scripts/gen-service-kustomize.sh
```diff
@@ -0,0 +1,64 @@
...
+pushd ${DEPLOY_DIR}
+
+cat <<EOF >kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+patches:
+- patch: |-
+    - op: replace
+      path: /metadata/namespace
```
dprince, line 53: Couldn't we just use kustomize's --namespace option to do this?

stuggi: yes thats possible, let me update it

PR #7, opened 2022-06-27
Install latest go 1.17.x

olliewalsh:
keystone-operator requires 1.17

### DIFF
--- a/devsetup/roles/download_tools/defaults/main.yaml
+++ b/devsetup/roles/download_tools/defaults/main.yaml
@@ -10,7 +10,7 @@ opm_version: latest
 sdk_version: v1.14.0
 
 # golang version
-go_version: 1.16.9
+go_version: 1.17.11
 
 # kustomize version to use (must be specific version)
 kustomize_version: v4.0.1


devsetup/roles/download_tools/defaults/main.yaml
```diff
@@ -10,7 +10,7 @@ opm_version: latest
 sdk_version: v1.14.0
 
 # golang version
-go_version: 1.16.9
+go_version: 1.17.11
```
stuggi, line 13: should update in var/defaults.yaml https://github.com/openstack-k8s-operators/install_yamls/blob/master/devsetup/vars/default.yaml#L9

olliewalsh: hrmm, <scratches_head/> for some reason it it didn't install 1.17.9 for me

stuggi: I mean its not wrong to update here the default of the role to some 1.17 release, as that's the version we want to use, but for having the latest we should also do it in the var file

PR #12, opened 2022-07-12
Make sure the PV node is a worker

ccamacho:
This commit makes sure the
oc get nodes command get only
one worker node, otherwise the
oc debug command will fail.


stuggi (approved):
/lgtm

PR #13, opened 2022-07-15
Allow setting service containerImage via Makefile

stuggi:
Adds env variables to the Makefile to set the containerImage for
the OpenStack service to use for service instance created.

Defaults those images now to quay.io/tripleowallabycentos9/<svc>

PR #14, opened 2022-07-19
Add support for installing/deploying Glance

abays:


Makefile
```diff
@@ -24,6 +24,12 @@ PLACEMENT_REPO      ?= https://github.com/openstack-k8s-operators/placement-oper
 PLACEMENT_BRANCH    ?= master
 PLACEMENTAPI        ?= config/samples/placement_v1beta1_placementapi.yaml
 
+# Sir Glancealot
+GLANCE_IMG       ?= quay.io/openstack-k8s-operators/glance-operator-index:latest
+GLANCE_REPO      ?= https://github.com/openstack-k8s-operators/glance-operator.git
+GLANCE_BRANCH    ?= master
+GLANCEAPI        ?= config/samples/glance_v1beta1_glanceapi.yaml
```
stuggi, line 31: should add an equivalent for glance for the service image what we introduced with https://github.com/openstack-k8s-operators/install_yamls/pull/13

abays: Should we also update the `config/samples` in the respective operator repos to point to `quay.io/tripleowallabycentos9` instead of `quay.io/tripleotraincentos8`?

stuggi (approved):
/lgtm

Makefile
stuggi: yes I think that probably makes sense

PR #18, opened 2022-08-09
Set CRC CPU/Memory defaults

fultonj:
Add "crc config set {cpus,memory}" commands to make it easier to increase the defaults. Default values to the `CRC` defaults as per the docs [1].

[1] https://access.redhat.com/documentation/en-us/red_hat_openshift_local/2.5/html/getting_started_guide/configuring_gsg


abays (approved):
/lgtm

PR #20, opened 2022-08-10
Allow CPUS/MEMORY change by setting env vars

stuggi:
To change the default memory and/or cpus for the VM use e.g.
`CPUS=12 MEMORY=20000 make crc`


abays (approved):
/lgtm

PR #21, opened 2022-08-23
Bump CRC storage to 10Gi to align with rabbit

dprince:
The rabbitmq/cluster-operator defaults are 10Gi so lets
bump these to align with that.


abays (approved):
/lgtm

PR #23, opened 2022-08-30
Add RWO,ROX,RWX to local-storage PVs

fmount:
During the glance-operator development, we hit a few issues related to
pvc(s) stuck in Pending mode, while controller wasn't able to schedule
any pod on the single node OCP.
It turned out that [1] requests the PVC with RWX, which makes sense in
a multinode environment with storage classes different from "local-storage".
This patch adds all the potential combinations needed to address any kind
of operator development.

[1] https://github.com/openstack-k8s-operators/glance-operator/blame/master/pkg/glance/pvc.go#L20

Signed-off-by: Francesco Pantano <fpantano@redhat.com>
Co-authored-by: John Fulton <fulton@redhat.com>

### DIFF
--- a/crc/storage.yaml
+++ b/crc/storage.yaml
@@ -15,6 +15,8 @@ spec:
     storage: 10Gi
   accessModes:
     - ReadWriteOnce
+    - ReadWriteMany
+    - ReadOnlyMany
   persistentVolumeReclaimPolicy: Delete
   local:
     path: "/mnt/openstack/pv001"
@@ -39,6 +41,8 @@ spec:
     storage: 10Gi
   accessModes:
     - ReadWriteOnce
+    - ReadWriteMany
+    - ReadOnlyMany
   persistentVolumeReclaimPolicy: Delete
   local:
     path: "/mnt/openstack/pv002"
@@ -63,6 +67,8 @@ spec:
     storage: 10Gi
   accessModes:
     - ReadWriteOnce
+    - ReadWriteMany
+    - ReadOnlyMany
   persistentVolumeReclaimPolicy: Delete
   local:
     path: "/mnt/openstack/pv003"
@@ -87,6 +93,8 @@ spec:
     storage: 10Gi
   accessModes:
     - ReadWriteOnce
+    - ReadWriteMany
+    - ReadOnlyMany
   persistentVolumeReclaimPolicy: Delete
   local:
     path: "/mnt/openstack/pv004"
@@ -111,6 +119,8 @@ spec:
     storage: 10Gi
   accessModes:
     - ReadWriteOnce
+    - ReadWriteMany
+    - ReadOnlyMany
   persistentVolumeReclaimPolicy: Delete
   local:
     path: "/mnt/openstack/pv005"
@@ -135,6 +145,8 @@ spec:
     storage: 10Gi
   accessModes:
     - ReadWriteOnce
+    - ReadWriteMany
+    - ReadOnlyMany
   persistentVolumeReclaimPolicy: Delete
   local:
     path: "/mnt/openstack/pv006"


fmount:
@abays I'd like to hear your feedback on this, can you check when you have the chance?

crc/storage.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: storage.k8s.io/v1
 metadata:
   name: local-storage
 provisioner: kubernetes.io/no-provisioner
-volumeBindingMode: WaitForFirstConsumer
+volumeBindingMode: Immediate
```
abays, line 6: `Immediate` is fine for our single-node CRC environments, but there is talk of changing the `make crc_storage` to something non-CRC specific.  The initial `install_yamls` context actually did not even include/involve CRC, but over time we sort of injected CRC into this repo.  I believe the plan is to eventually extricate it and refactor `install_yamls` to work with "any" OCP cluster (with certain constraints, of course).

Anyhow, this change is probably fine for now.  Just wanted to document the aforementioned concern for potential reconsideration later on.

crc/storage.yaml
```diff
@@ -15,6 +15,8 @@ spec:
     storage: 10Gi
   accessModes:
     - ReadWriteOnce
+    - ReadWriteMany
+    - ReadOnlyMany
```
abays, line 19: I'm not fully-versed on these `accessModes`, but I wonder if there is a scenario in which it would be desirable to have PVs of `local-storage` storage class with mutually exclusive `accessModes`.  i.e. say, 2 with `RWO`, 2 with `ROX` and 2 with `RWX` (as opposed to having all `accessModes` on every PV)?  @dprince do you have thoughts on this?

crc/storage.yaml
dprince: Yeah, could we leave the volumeBindingMode alone for now and just merge the changes you have below for accessModes?

fmount: ack, works for me, the issue I hit was a combination of `accessModes` and `volumeBindingMode`, but having the right access mode is the blocker

crc/storage.yaml
fmount: Yeah, I see your point, and I'm not sure about the potential combination required by different types of operators.
We can try to update the patch and try with a subset of PVs w/ `RWO` and `RWX` (mutually exclusive) and let the scheduler do its job, but in theory, having both `RWX` and `RWO` allows to use the `PV` for both use cases. 
Any particular reason we'd like to have mutually exclusive `accessModes` ?

abays: I couldn't think of any off the top of my head, but I just thought I'd throw it out there.  I'm good with the change as-is.

crc/storage.yaml
abays: Ah yeah, sorry, I didn't mean to imply only CRC is ever single-node.  SNO is definitely out there and in-use regularly.

fultonj:
It's fine by me to not change volumeBindingMode to Immediate from WaitForFirstConsumer in this PR. I just wanted to share why it was initially asked for.

I tested this patch in its current state on CRC by building my own version of the glance operator using WaitForFirstConsumer and the manager got stuck 'Waiting for GlanceAPI PVC to bind'. Running `oc get pv | grep local` showed plenty of available volumes and `oc get pvc` showed a PVC that was Pending for glance. After 15 minutes of not getting anything I recreated the CRC environment but with volumeBindingMode Immediate. After building and using the operator a run of `oc get pvc` showed that the volume was bound to the glance pod.

Next step is probably to figure out why WaitForFirstConsumer didn't result in getting past a Pending PVC. If we get more info maybe there can be a follow up PR. I'll probably use Immediate as a workaround in my own environment until then.

abays:
> It's fine by me to not change volumeBindingMode to Immediate from WaitForFirstConsumer in this PR. I just wanted to share why it was initially asked for.
> 
> I tested this patch in its current state on CRC by building my own version of the glance operator using WaitForFirstConsumer and the manager got stuck 'Waiting for GlanceAPI PVC to bind'. Running `oc get pv | grep local` showed plenty of available volumes and `oc get pvc` showed a PVC that was Pending for glance. After 15 minutes of not getting anything I recreated the CRC environment but with volumeBindingMode Immediate. After building and using the operator a run of `oc get pvc` showed that the volume was bound to the glance pod.
> 
> Next step is probably to figure out why WaitForFirstConsumer didn't result in getting past a Pending PVC. If we get more info maybe there can be a follow up PR. I'll probably use Immediate as a workaround in my own environment until then.

The `WaitForFirstConsumer` might not have been the problem for you.  If you only have the 6 PVs that `install_yamls` creates, those are all of `local-storage` storage class.  Currently, however, the Glance [sample config](https://github.com/openstack-k8s-operators/glance-operator/blob/master/config/samples/glance_v1beta1_glanceapi.yaml) that we use requests empty-string storage class.  So it might be worth trying to add `storageClass: local-storage` to the Glance CR if you haven't already given that a shot.

fmount:
> > It's fine by me to not change volumeBindingMode to Immediate from WaitForFirstConsumer in this PR. I just wanted to share why it was initially asked for.
> > I tested this patch in its current state on CRC by building my own version of the glance operator using WaitForFirstConsumer and the manager got stuck 'Waiting for GlanceAPI PVC to bind'. Running `oc get pv | grep local` showed plenty of available volumes and `oc get pvc` showed a PVC that was Pending for glance. After 15 minutes of not getting anything I recreated the CRC environment but with volumeBindingMode Immediate. After building and using the operator a run of `oc get pvc` showed that the volume was bound to the glance pod.
> > Next step is probably to figure out why WaitForFirstConsumer didn't result in getting past a Pending PVC. If we get more info maybe there can be a follow up PR. I'll probably use Immediate as a workaround in my own environment until then.
> 
> The `WaitForFirstConsumer` might not have been the problem for you. If you only have the 6 PVs that `install_yamls` creates, those are all of `local-storage` storage class. Currently, however, the Glance [sample config](https://github.com/openstack-k8s-operators/glance-operator/blob/master/config/samples/glance_v1beta1_glanceapi.yaml) that we use requests empty-string storage class. So it might be worth trying to add `storageClass: local-storage` to the Glance CR if you haven't already given that a shot.

I hit the same issue even with `StorageClass: local-storage`, and I fixed it by manually editing the `glance pv` adding `volumeName: local-storage00x`: as long as I "manually" bound it to a specific PV, the controller flow moved on.

fultonj:
> > It's fine by me to not change volumeBindingMode to Immediate from WaitForFirstConsumer in this PR. I just wanted to share why it was initially asked for.
> > I tested this patch in its current state on CRC by building my own version of the glance operator using WaitForFirstConsumer and the manager got stuck 'Waiting for GlanceAPI PVC to bind'. Running `oc get pv | grep local` showed plenty of available volumes and `oc get pvc` showed a PVC that was Pending for glance. After 15 minutes of not getting anything I recreated the CRC environment but with volumeBindingMode Immediate. After building and using the operator a run of `oc get pvc` showed that the volume was bound to the glance pod.
> > Next step is probably to figure out why WaitForFirstConsumer didn't result in getting past a Pending PVC. If we get more info maybe there can be a follow up PR. I'll probably use Immediate as a workaround in my own environment until then.
> 
> The `WaitForFirstConsumer` might not have been the problem for you. If you only have the 6 PVs that `install_yamls` creates, those are all of `local-storage` storage class. Currently, however, the Glance [sample config](https://github.com/openstack-k8s-operators/glance-operator/blob/master/config/samples/glance_v1beta1_glanceapi.yaml) that we use requests empty-string storage class. So it might be worth trying to add `storageClass: local-storage` to the Glance CR if you haven't already given that a shot.

I should have mentioned that I already had `storageClass: local-storage` in `glance_v1beta1_glanceapi.yaml` when I encountered this behavior.

abays (approved):
/lgtm

fultonj:
> It's fine by me to not change volumeBindingMode to Immediate from WaitForFirstConsumer in this PR. I just wanted to share why it was initially asked for.
> 
> I tested this patch in its current state on CRC by building my own version of the glance operator using WaitForFirstConsumer and the manager got stuck 'Waiting for GlanceAPI PVC to bind'. Running `oc get pv | grep local` showed plenty of available volumes and `oc get pvc` showed a PVC that was Pending for glance. After 15 minutes of not getting anything I recreated the CRC environment but with volumeBindingMode Immediate. After building and using the operator a run of `oc get pvc` showed that the volume was bound to the glance pod.

The issue described above no longer reproduces for me when I use the following:

  https://github.com/openstack-k8s-operators/glance-operator/pull/30/

PR #24, opened 2022-09-13
Adoption: don't overwrite existing secrets

jistr:
When doing data plane adoption, we set secrets to the ones in the existing deployment. However, they would get overwritten to the default during operator deployment (like `make keystone`), as the operator targets pull in the `input` target, and until now it always overwrote the passwords to their install_yamls default.

We will only create the osp-secret when it doesn't exist yet, enabling the adoption scenario. The one situation when this could create an issue is when a new service (new password) is being added. In that case, the developer will have to explicitly run `make input_cleanup` before deploying their operator or calling `make input` again.


dprince (approved):

jistr:
Re doing this for other Makefile targets - we could but it may not matter as much.

For the passwords it is needed because the rule is not being treated as a "leaf" rule executed directly by the developer. Typically it's being pulled in automatically as a dependency of some other rule, so it's re-executed often without the developer explicitly requesting that.

PR #25, opened 2022-09-15
Support absolute patch YAMLs in *API

Akrog:
There are times when we want to use this project to deploy things but we want to do so with our own CR manifest files.

For example to deploy Glance with a custom manifest to use our Ceph cluster.

This is currently not possible, since `GLANCEAPI` is always expected to be a relative path within the `GLANCE_REPO` repository.

This patch adds the possibility, to all the operators, of setting absolute path using the `*API_CR` environmental variables:

- KEYSTONEAPI_CR
- MARIADB_CR
- PLACEMENTAPI_CR
- GLANCEAPI_CR
- NEUTRONAPI_CR
- CINDER_CR
- RABBITMQ_CR
- IRONIC_CR

An alternative would be to try to figure out if the `*API` environmental variables are relative to the `*_REPO` location with something like:
```
GLANCEAPI_CR      = ${OPERATOR_BASE_DIR}/glance-operator/${GLANCEAPI}
ifeq ($(realpath $(GLANCEAPI_CR)),)
  GLANCEAPI_CR = $(GLANCEAPI)
endif
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ KEYSTONE_IMG        ?= quay.io/openstack-k8s-operators/keystone-operator-index:l
 KEYSTONE_REPO       ?= https://github.com/openstack-k8s-operators/keystone-operator.git
 KEYSTONE_BRANCH     ?= master
 KEYSTONEAPI         ?= config/samples/keystone_v1beta1_keystoneapi.yaml
+KEYSTONEAPI_CR      ?= ${OPERATOR_BASE_DIR}/keystone-operator/${KEYSTONEAPI}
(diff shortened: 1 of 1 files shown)


stuggi:
looks god to me, just for the naming, should we call it _CR instead of _YAML? its a yaml which has the specifics of the custom resource.

Akrog:
I was having trouble deciding on the ending, and I didn't like the `_YAML_ ending much.
I like your suggestion though, so I will update it now.

abays (approved):
/lgtm

stuggi (approved):
/lgtm thanks!

PR #26, opened 2022-09-16
Bump Operator SDK to 1.23.0

abays:
We're currently using `1.23.0` in the operators themselves

### DIFF
--- a/devsetup/vars/default.yaml
+++ b/devsetup/vars/default.yaml
@@ -3,10 +3,10 @@
 opm_version: latest
 
 # operator-sdk version to use (must be specific version)
-sdk_version: v1.22.2
+sdk_version: v1.23.0
 
 # golang version
-go_version: 1.18.5
+go_version: 1.18.6
 
 # kustomize version to use (must be specific version)
-kustomize_version: v4.5.4
+kustomize_version: v4.5.7


devsetup/vars/default.yaml
```diff
@@ -3,7 +3,7 @@
 opm_version: latest
 
 # operator-sdk version to use (must be specific version)
-sdk_version: v1.22.2
+sdk_version: v1.23.0
 
 # golang version
 go_version: 1.18.5
```
stuggi, line 9: should we also update go_version to latest 1.18.6 and kustomize_version to v4.5.7

abays: Sure.  What's the best method to find the corresponding Go version?  I didn't see it indicated in the Operator SDK repo.

stuggi: maybe we do not update the go version, though not sure if a minor version diff is that critical
~~~
$ operator-sdk version
operator-sdk version: "v1.23.0", commit: "1eaeb5adb56be05fe8cc6dd70517e441696846a4", kubernetes version: "1.24.2", go version: "go1.18.5", GOOS: "linux", GOARCH: "amd64"
~~~

abays: Well, for this one time I will try the new values you suggest and then, if they seem to work, let's just update them now in this one PR

stuggi (approved):
/lgtm

dprince (approved):

PR #31, opened 2022-10-10
Allow storage class selection via Kustomize

abays:
We are seeing deployment issues occur in CI related to the fact that it utilizes different storage classes in its non-CRC environment than we do in CRC.  This patch exposes an environment variable knob to allow the user to choose a storage class to use with the operators (any operator top-level CRD that has `storageClass` at the root-level of its `spec`).  The variable defaults to `local-storage`, as this is what we explicitly provide via `make crc_storage`.  CI will set it to whatever they need for their environments.

Depends on: https://github.com/openstack-k8s-operators/openstack-operator/pull/28

### DIFF
--- a/scripts/gen-service-kustomize.sh
+++ b/scripts/gen-service-kustomize.sh
@@ -60,6 +60,9 @@ patches:
     - op: replace
       path: /spec/containerImage
       value: ${IMAGE}
+    - op: replace
+      path: /spec/storageClass
+      value: ${STORAGE_CLASS}
   target:
     kind: ${KIND}
 EOF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,9 @@ OPERATOR_BASE_DIR   ?= ${OUT}/operator
 SERVICE_REGISTRY    ?= quay.io
 SERVICE_ORG         ?= tripleowallabycentos9
 
+# storage (used by some operators)
+STORAGE_CLASS       ?= "local-storage"
+
 # OpenStack Operator
 OPENSTACK_IMG        ?= quay.io/openstack-k8s-operators/openstack-operator-index:latest
 OPENSTACK_REPO       ?= https://github.com/openstack-k8s-operators/openstack-operator.git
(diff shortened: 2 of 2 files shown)


scripts/gen-service-kustomize.sh
```diff
@@ -47,6 +47,23 @@ fi
...
+      path: /spec/storageClass
+      value: ${STORAGE_CLASS}"
+elif [ "$KIND" == "OpenStackControlPlane" ]; then
+      STORAGE_CLASS_REPLACE="
+    - op: replace
+      path: /spec/storageClass
+      value: ${STORAGE_CLASS}
+    - op: replace
+      path: /spec/glanceTemplate/storageClass
+      value: ${STORAGE_CLASS}"
```
stuggi, line 64: do we need to set this with https://github.com/openstack-k8s-operators/openstack-operator/blob/fb0de2a0805580ddd928cfacb3e38b44f171d035/pkg/openstack/glance.go#L36 ? if not, we could just blind set /spec/storageClass for all service CRs, if they don't have it, it gets ignored by the api server and we'd not have to do this service condition kind check here?

stuggi: just thinking about not adding service specific things here which requires updates when we add additional services.

abays: That's a good point actually.  Now that we have that `OpenStackControlPlane` `storageClass` propagation patch [1] merged for Glance, we can just do a blind `replace` here with Kustomize without checking anything else.  Really the `KIND` comparison was just to handle the longer paths, because Kustomize throws an error if an intermediate dir is missing in the path (i.e. if you wanted to replace `spec/glanceTemplate/storageClass` and the CRD did not have `glanceTemplate` in the spec).  But if we're always doing `/spec/storageClass` we should be fine.

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/29

abays: Changed to use the blind `replace`

stuggi (approved):
/lgtm

PR #35, opened 2022-10-13
Makefile: Fix cinder_deploy_prep

Akrog:
Running `make cinder_deploy` fails with:
```
  + echo 'Please set IMAGE'
  Please set IMAGE
  + exit 1
  make: *** [Makefile:423: cinder_deploy_prep] Error 1
```

This also makes the `make cinder_deploy_cleanup` fail.

The issue is in `cinder_deploy_prep` that is missing the export of the IMAGE env var that we have in other `_deploy_prep` targets.

The reason why it was missing is because cinder has 4 different images,
one for each of its services.

This patch exports the `IMAGE` variable to avoid failures with `make`,
but it doesn't actually do anything with that variable, as indicated by
the assigned value.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -413,6 +413,7 @@ cinder_cleanup: ## deletes the operator, but does not cleanup the service resour
 
 .PHONY: cinder_deploy_prep
 cinder_deploy_prep: export KIND=Cinder
+cinder_deploy_prep: export IMAGE=unused
 cinder_deploy_prep: cinder_deploy_cleanup ## prepares the CR to install the service based on the service sample file CINDER
 	$(eval $(call vars,$@,cinder))
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR} ${DEPLOY_DIR}


abays (approved):
/lgtm

Makefile
```diff
@@ -413,6 +413,7 @@ cinder_cleanup: ## deletes the operator, but does not cleanup the service resour
 
 .PHONY: cinder_deploy_prep
 cinder_deploy_prep: export KIND=Cinder
+cinder_deploy_prep: export IMAGE="${CINDER_IMG}"
```
stuggi, line 416: the IMAGE is usually the one to be used by the service (api/scheduler/volume). with cinder having multiple nested services, this is right now not supported here to set. it works as it kustomize won't fail with setting this here, but right now its to wrong image and won't have any affect and uses the images set in the sample cr. maybe we just set it to a string which shows that it does not result in a used setting?

Akrog: That sounds like a good idea, I'll update the patch.

stuggi (approved):
/lgtm

Makefile
abays: Yes, good point.  Elsewhere for other operators we set it to something to indicate that it is not used: https://github.com/openstack-k8s-operators/install_yamls/blob/259ac56930a128ea9ee99727b47cb99c5987b422/Makefile#L183

abays (approved):
/lgtm

PR #36, opened 2022-10-13
Makefile: Support absolute path CR for OpenStack

Akrog:
In a previous patch we added support for absolute paths in the CR for the different CRs, but not for the OPENSTACK one.

This patch adds the `OPENSTACK_CR` env variable that makes it possible to provide a manifest that's not from the openstack-operator repository.

This is useful to do a full deployment of MariaDB + RabbitMQ + Keystone
+ Glance + Cinder from a custom CR.

This change is backward compatible as the existing OPENSTACK_CTRLPLANE env var still works.


abays (approved):
/lgtm

PR #37, opened 2022-10-27
Add proper DB cleanup to nova_deploy_cleanup

gibizer:
So far we only dropped nova_api and nova_cell0 DBs but we need to handle any number of cells.


abays (approved):
/lgtm

stuggi (approved):
/lgtm

PR #40, opened 2022-11-10
Create presistent volumes on all worker nodes

dciabrin:
On Openshift environments with multiple workers, create the local paths on all nodes and configure the affinity of persistent volumes to allow them to be created on all nodes.

### DIFF
--- a/scripts/create-pv.sh
+++ b/scripts/create-pv.sh
@@ -16,9 +16,11 @@
 set -ex
 PV_NUM=${PV_NUM:-12}
 
-NODE_NAME=$(oc get node -o name -l node-role.kubernetes.io/worker | head -n 1)
-if [ -z "$NODE_NAME" ]; then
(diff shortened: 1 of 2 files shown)


stuggi:
While this helps to schedule pods to all workers, when having multiple worker nodes wouldn't it be better to have e.g. an nfs storage + storageclass, like in when using osp-director-dev-tools and set the STORAGE_CLASS for install_yamls to whatever is the storage class, e.g. host-nfs-storageclass [1]. As soon testing moving pods to a different worker I think this is what we need, right?

[1] https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/master/ansible/vars/default.yaml#L286

dprince (approved):

dprince:
@stuggi I'm good with what @dciabrin is doing here. If its useful in some "simpler dev setups" I don't think it hurts anything to do this on all workers.

PR #41, opened 2022-11-14
Align rabbitmq CR name with openstack-operator

dprince:
This aligns the name of the RabbitmqCluster resource which currently gets its name from the upstream
github.com/rabbitmq/cluster-operator example with the name that we use in the openstack-operator.

With this patch both resources should be named
simply 'rabbitmq'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -532,8 +532,8 @@ rabbitmq_deploy_prep: rabbitmq_deploy_cleanup ## prepares the CR to install the 
 .PHONY: rabbitmq_deploy
 rabbitmq_deploy: input rabbitmq_deploy_prep ## installs the service instance using kustomize. Runs prep step in advance. Set RABBITMQ_REPO and RABBITMQ_BRANCH to deploy from a custom repo.
 	$(eval $(call vars,$@,rabbitmq))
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

Makefile
```diff
@@ -532,13 +532,13 @@ rabbitmq_deploy_prep: rabbitmq_deploy_cleanup ## prepares the CR to install the
 .PHONY: rabbitmq_deploy
 rabbitmq_deploy: input rabbitmq_deploy_prep ## installs the service instance using kustomize. Runs prep step in advance. Set RABBITMQ_REPO and RABBITMQ_BRANCH to deploy from a custom repo.
 	$(eval $(call vars,$@,rabbitmq))
-	#oc kustomize ${DEPLOY_DIR} | oc apply -f -
-	oc apply -f ${DEPLOY_DIR}/rabbitmq.yaml
+	KIND=RabbitmqCluster NAME=rabbitmq bash scripts/gen-name-kustomize.sh
```
Akrog, line 535: Not thrilled about requiring a new bash script just for this, but I understand you are doing it this way to leverage it in the future.

Akrog (commented):
/lgtm

PR #43, opened 2022-11-30
Add targets to install service operator crds only

stuggi:
To develop the openstack-operator without installing it, there is the need that all crds of the openstack services it can handle to be installed. To not have to install them this adds per service target to install it and a global `make crds` to install all of them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,9 @@ all: namespace keystone mariadb placement neutron
 help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
+.PHONY: crds
+crds: nova_crd octavia_crd neutron_crd ovn_crd ironic_crd cinder_crd glance_crd placement_crd keystone_crd mariadb_crd rabbitmq_crd ## installs all service operator crds
+
 .PHONY: cleanup
 cleanup: nova_cleanup octavia_cleanup neutron_cleanup ovn_cleanup ironic_cleanup cinder_cleanup glance_cleanup placement_cleanup keystone_cleanup mariadb_cleanup ## Delete all operators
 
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

dprince:
What I've been doing is installing with OLM first. Setting replicas to 0 and then proceeding. That way you get permissions, etc. Are folks really testing without OLM at all and thus want to manually install all the CRDs and permissions. There is also a 'make install' target in each operator repo

stuggi:
> What I've been doing is installing with OLM first. Setting replicas to 0 and then proceeding. That way you get permissions, etc. Are folks really testing without OLM at all and thus want to manually install all the CRDs and permissions. There is also a 'make install' target in each operator repo

thats a possible workflow, in the way I uses it right now I don't use olm at all. Also do not use the service operators. I add new CRDs to the openstack-operator which are independent from services. So only the crds are required to be installed. right, could use make install in each operator when checked out, instead of the oc apply.

SeanMooney:
until today when i merged the rabbit mq support into nova the nova operator has never need the OpenStack operator to be installed so i have never installed it until now

i generally only deploy mariadb keystone and placement with isntall_yamls
and then install nova with make install in the nova-operator repo then run it in a debugger.

it would be nice if we could continue to do that.

unfortunately since the transport URL crd is provided by the oepnstack meta operator currently i can no longer do that. it would be nice if we could eventually move that to a different repo so that the depencies are one way. meta operator to service operators.

dprince:
@stuggi consider this implementation https://github.com/dprince/install_yamls/pull/new/crd_installer

stuggi:
> until today when i merged the rabbit mq support into nova the nova operator has never need the OpenStack operator to be installed so i have never installed it until now
> 
> i generally only deploy mariadb keystone and placement with isntall_yamls and then install nova with make install in the nova-operator repo then run it in a debugger.
> 
> it would be nice if we could continue to do that.
> 
> unfortunately since the transport URL crd is provided by the oepnstack meta operator currently i can no longer do that. it would be nice if we could eventually move that to a different repo so that the depencies are one way. meta operator to service operators.

right, with the adding the transport url into the openstack-operator we removed this one way imports. we might want to move the transporturl and coming networking functionality into their own repos or an combined infra-operator (or whatever the name is). @dprince thoughts on this?

stuggi:
done via different approach in https://github.com/openstack-k8s-operators/install_yamls/pull/44

PR #45, opened 2022-12-02
Fix make crc_storage

gibizer:
After https://github.com/openstack-k8s-operators/install_yamls/pull/40 make crc_storage started failing with:

  /usr/bin/bash: -c: line 1: syntax error near unexpected token `02'
  /usr/bin/bash: -c: line 1: `02'

That patch added seq to generate the PV id sequence. It seems that the issue is caused by the default '\n' separator of seq. So this patch adds -s ' ' to restore the functionality.


stuggi (approved):
/lgtm

PR #46, opened 2022-12-05
Use Zed for container images

abays:
I'm assuming we should be defaulting to Zed now, given we do in the `OpenStackControlPlane` example CR [1].

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/config/samples/core_v1beta1_openstackcontrolplane.yaml


stuggi (approved):
/lgtm

PR #48, opened 2022-12-09
Indicate devsetup's Makefile path

4383:
Some given examples ask to execute `make` commands, however, this repository comes with multiple Makefile so its not clear where Makefile target are stored at first glance.

These changes propose to indicate the path to go to before executing these examples, hence, the apropriate Makefile will be at hand for users.


stuggi (approved):
/lgtm . we planned to move the devsetup out of install_yamls into its own repo ... but so far had not time to do it :)

PR #52, opened 2023-01-13
Initial EDPM integration

slagle:
- Add ansibleee-operator to Makefile
- Add edpm-compute to devsetup/Makefile
- Add edpm-play.yaml

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -53,3 +53,11 @@ edpm_compute_cleanup:
 	-sudo virsh destroy edpm-compute-${EDPM_COMPUTE_SUFFIX}
 	-sudo virsh undefine edpm-compute-${EDPM_COMPUTE_SUFFIX}
 	rm -f ${HOME}/.crc/machines/crc/edpm-compute-${EDPM_COMPUTE_SUFFIX}.qcow2
+
+.PHONY: edpm_play
+edpm_play:
+	oc create -f edpm/edpm-play.yaml
+
+.PHONY: edpm_play_cleanup
+edpm_play_cleanup:
+	-oc delete ansibleee deploy-external-dataplane-compute
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -1,6 +1,7 @@
 CRC_URL ?= 'https://developers.redhat.com/content-gateway/rest/mirror/pub/openshift-v4/clients/crc/latest/crc-linux-amd64.tar.xz'
 KUBEADMIN_PWD ?= 12345678
 PULL_SECRET  ?= ${PWD}/pull-secret.txt
+EDPM_COMPUTE_SUFFIX ?= 0
 
 ##@ General
 
@@ -37,3 +38,18 @@ download_tools: ## Runs an ansible playbook to install required tools with the v
 	ANSIBLE_FORCE_COLOR=true ansible-playbook \
 	-v -i hosts \
 	download_tools.yaml
+
+.PHONY: crc_attach_default_interface
+crc_attach_default_interface:
+	sudo virsh attach-interface crc --source default --type network --model virtio --config --persistent
+
+.PHONY: edpm_compute
+edpm_compute:
+	scripts/gen-ansibleee-ssh-key.sh
+	scripts/gen-edpm-compute-node.sh ${EDPM_COMPUTE_SUFFIX}
+
+.PHONY: edpm_compute_cleanup
+edpm_compute_cleanup:
+	-sudo virsh destroy edpm-compute-${EDPM_COMPUTE_SUFFIX}
+	-sudo virsh undefine edpm-compute-${EDPM_COMPUTE_SUFFIX}
+	rm -f ${HOME}/.crc/machines/crc/edpm-compute-${EDPM_COMPUTE_SUFFIX}.qcow2
--- a/devsetup/edpm/edpm-play.yaml
+++ b/devsetup/edpm/edpm-play.yaml
@@ -1,0 +1,217 @@
+apiVersion: redhat.com/v1alpha1
+kind: AnsibleEE
+metadata:
+  name: deploy-external-dataplane-compute
+  namespace: openstack
+spec:
+  # We can specify either playbook, which will run with default ansible-runner options,
+  # or args, which allows specify the whole command that we want to execute
+  play: |
+
+    - hosts: all
+      tasks:
+        - name: Enable tripleo-repos
+          shell: |
+            rpm -q git || sudo yum -y install git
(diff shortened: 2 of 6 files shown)


devsetup/edpm/edpm-play.yaml
```diff
@@ -0,0 +1,217 @@
...
+                      region_name: regionOne
+                      send_service_user_token: 'True'
+                      user_domain_name: Default
+                      username: nova
+                    vnc:
+                      novncproxy_base_url: http://192.168.24.3:6080
+                  tripleo_nova_compute_neutron_auth_type: v3password
+                  tripleo_nova_compute_neutron_auth_url: http://192.168.24.3:5000/v3
+                  tripleo_nova_compute_neutron_password: viaNg9cyAqS8N1ydQe7n3XkuM
+                  tripleo_nova_compute_neutron_project_domain_name: Default
```
dprince, line 157: would it be reasonable to file a Jira to rename all these parameters to *not* use tripleo_ in the names?

devsetup/edpm/edpm-play.yaml
```diff
@@ -0,0 +1,217 @@
+apiVersion: redhat.com/v1alpha1
```
dprince, line 1: Would it work better to host this file withing the openstack-ansible-operator (or somewhere else) so that the parameters, examples are more close to the codebase? Many of the operators here have the samples they instantiate live in the operators themselves. Like: https://github.com/openstack-k8s-operators/openstack-operator/blob/master/config/samples/core_v1beta1_openstackcontrolplane.yaml

devsetup/edpm/edpm-play.yaml
slagle: Yes, we're going to be migrating the ansible code from tripleo-ansible to edpm-ansible (under openstack-k8s-operators ideally). As part of that migration, all the tripleo naming/branding will be removed.

devsetup/edpm/edpm-play.yaml
slagle: Well, openstack-ansibleee-operator is generic. It is not EDPM specific. A sample likely makes sense in openstack-operator, as we make progress on the dataplane CRD/controller. I put it here for the time being as this is purely dev/test setup.

dprince: I'm good with it.

dprince (approved):

devsetup/Makefile
```diff
@@ -37,3 +38,26 @@ download_tools: ## Runs an ansible playbook to install required tools with the v
...
+.PHONY: edpm_compute
+edpm_compute:
+	scripts/gen-ansibleee-ssh-key.sh
+	scripts/gen-edpm-compute-node.sh ${EDPM_COMPUTE_SUFFIX}
+
+.PHONY: edpm_compute_cleanup
+edpm_compute_cleanup:
+	-sudo virsh destroy edpm-compute-${EDPM_COMPUTE_SUFFIX}
+	-sudo virsh undefine edpm-compute-${EDPM_COMPUTE_SUFFIX}
+	rm -f ${HOME}/.crc/machines/crc/edpm-compute-${EDPM_COMPUTE_SUFFIX}.qcow2
```
raukadah, line 55: All the interfaces works fine. I have few suggestions which we can do in the new pull request.
I think we can improve the edpm_compute_cleanup to
- also cleanup out/edpm/edpm-compute-0-id_rsa.pub to avoid following issues
```
+ sudo virt-copy-out -d edpm-compute-0 /root/.ssh/id_rsa.pub ../out/edpm
+ mv ../out/edpm/id_rsa.pub ../out/edpm/edpm-compute-0-id_rsa.pub
mv: replace '../out/edpm/edpm-compute-0-id_rsa.pub', overriding mode 0644 (rw-r--r--)? 
```
 and also delete ansibleee-ssh-key-secret secret.
```
oc delete -f ../out/edpm/ansibleee-ssh-key-secret.yaml
```
Rest looks good.

raukadah (approved):
Nice work James o/

devsetup/Makefile
```diff
@@ -37,3 +38,26 @@ download_tools: ## Runs an ansible playbook to install required tools with the v
...
+
+.PHONY: edpm_play
+edpm_play:
+	oc create -f edpm/edpm-play.yaml
+
+.PHONY: edpm_play_cleanup
+edpm_play_cleanup:
+	-oc delete ansibleee deploy-external-dataplane-compute
```
bogdando, line 63: typo?

raukadah: The target actually executes the command and it works successfully. I have removed dash in this pr https://github.com/openstack-k8s-operators/install_yamls/pull/56

devsetup/Makefile
```diff
@@ -37,3 +38,26 @@ download_tools: ## Runs an ansible playbook to install required tools with the v
 	ANSIBLE_FORCE_COLOR=true ansible-playbook \
 	-v -i hosts \
 	download_tools.yaml
+
+.PHONY: crc_attach_default_interface
+crc_attach_default_interface:
+	sudo virsh attach-interface crc --source default --type network --model virtio --config --persistent
+
+.PHONY: edpm_compute
+edpm_compute:
```
SeanMooney, line 47: so this is really jsut creating the vms which is fine for testing with CRC

we might want to explore using CNV for this for CI/dev in the future instead.
but this is fine for now

devsetup/Makefile
```diff
@@ -37,3 +38,26 @@ download_tools: ## Runs an ansible playbook to install required tools with the v
...
+
+.PHONY: edpm_compute_cleanup
+edpm_compute_cleanup:
+	-sudo virsh destroy edpm-compute-${EDPM_COMPUTE_SUFFIX}
+	-sudo virsh undefine edpm-compute-${EDPM_COMPUTE_SUFFIX}
+	rm -f ${HOME}/.crc/machines/crc/edpm-compute-${EDPM_COMPUTE_SUFFIX}.qcow2
+
+.PHONY: edpm_play
+edpm_play:
+	oc create -f edpm/edpm-play.yaml
```
SeanMooney, line 59: This however will likely go away or change quite a lot when the OpenStack operator and service operators gain support for EDPM

we will not be running AEE palys here for long.
we will instead be creating instance of the OpenstackDataPlaneNode CRs which will be reconsiled by the openstack/dataplane operator and then the Plays will be run by the infra and service operators via creating the AnsibleEE CRs.

devsetup/edpm/edpm-play.yaml
slagle: yes

PR #53, opened 2023-01-16
Ensure operator is deleted by cleanup

kajinamit:
The current cleanup script does not delete the operator. This ensures the operator is actually deleted from the deployment. Note that crds should be deleted in advance, otherwise the operator is automatically restored.

### DIFF
--- a/scripts/operator-cleanup.sh
+++ b/scripts/operator-cleanup.sh
@@ -30,3 +30,8 @@ if [ -n "$CSV" ]; then
 fi
 oc delete -n ${NAMESPACE} subscription ${OPERATOR_NAME}-operator --ignore-not-found=true
 oc delete -n ${NAMESPACE} catalogsource ${OPERATOR_NAME}-operator-index --ignore-not-found=true
+
+for crd in `oc get crd -o name | grep "\.${OPERATOR_NAME}\.openstack\.org$"`; do
+  oc delete $crd
+done
+oc delete operators ${OPERATOR_NAME}-operator.openstack


scripts/operator-cleanup.sh
```diff
@@ -30,3 +30,8 @@ if [ -n "$CSV" ]; then
 fi
 oc delete -n ${NAMESPACE} subscription ${OPERATOR_NAME}-operator --ignore-not-found=true
 oc delete -n ${NAMESPACE} catalogsource ${OPERATOR_NAME}-operator-index --ignore-not-found=true
+
+for crd in `oc get crd -o name | grep "\.${OPERATOR_NAME}\.openstack\.org$"`; do
```
stuggi, line 34: while this works for the service operators, it does not for the CRDs coming with the openstack-operator which are right now
```
openstackcontrolplanes.core.openstack.org
transporturls.rabbitmq.openstack.org
```

It might be better to delete the CRDs from the operator csv `oc get csv openstack-operator.v0.0.1 -o json | jq -r ".spec.customresourcedefinitions.owned[].name"` ?

bshephar: I thought maybe this is something that should be handled by OLM when the operator is removed. But it appears there has been some extensive conversation on this topic previously:
https://github.com/operator-framework/operator-lifecycle-manager/issues/1326

dprince: Also, by design you might not want to delete the CRDs because doing so removes resources that are potentially using them.

Say I deploy a running OpenStackControlplane, and want to upgrade the operator codebase without destroying my running deployment. I would want the CRDs to persist.

dprince: This is an area where it is quite possible one developers workflow (always deleting CRDs) breaks another developers if we automate it too much. Might a more generic CRD cleanup work be more explicit and work for you just as well. make crd_cleanup -> (runs) oc get crd | grep openstack | cut -f 1 -d ' ' | xargs oc delete crd

kajinamit: I agree that we should check the owner of that CRD. I'll look into that.

Regarding the workflow, maybe making this optional (or adding a new command such as keystone_cleanup_all) would make sense ? I see operator sdk provides a similar option to purge also CRD (or operator group).

https://sdk.operatorframework.io/docs/cli/operator-sdk_cleanup/

Having the generic command might work but as we have options to manage individual operators, having separate commands for specific commands would be useful.

stuggi: right as Dan said its intentional that the CRDs stay, so that uninstall an operator to e.g. install a new version would not affect any created resources. It sounds good to me to have a DELETE_CRDS env var and if set the cleanup would also deplete the owned CRDs.

rabi: I came across this PR while looking for the operator uninstall behavior. The strange thing is it allows you to uninstall a operator (i.e delete csv, subscription, catalogsources) and leaves the CRs and CRDs. Who would manage those CRs and CRDs?  Also, it does not allow you to delete the CRs without clearing the finalizers on them.   I guess the right thing from OLM pov is to allow upgrade but not allow to uninstall the operator with existing CRs/CRDs.  Maybe there is better way to handle all these that I'm missing.

dprince: The suggestion here is to either have an ENV variable we can set (behavior should default to not cleaning them up I think), or a separate Makefile target for this action. Either of those would work to accommodate the concern here.

dprince (changes requested):
either add an ENV variable to control this or implement as a new Makefile target

kajinamit:
I'll close this now and rethink of the approach if needed.

PR #54, opened 2023-01-17
Introduce Ceph target in the Makefile

fmount:
This change allows to deploy a toy `Ceph` cluster made by one single pod and no real `OSD`.
After the cluster is up `pools` and `secrets` are created and it can be consumed by other
operators (e.g. `cinder` and `glance`).
`HostNetworking` is turned off by default, and the `Ceph` pod will be started on the `openshift`
cluster network.

However, running:

`HOSTNETWORK=true make ceph`

we can deploy a `Ceph` Pod listening on the `host network` and it can be also consumed by the
`edpm` hosts (provided they're bridged with `crc` or `SNO`).

We have 4 modes to deploy `Ceph`:

1. `make ceph` # the pod is bound to the `openshift-sdn`
2. `HOSTNETWORK=true make ceph` # the pod uses `hostNetworking` but the container relies on `$NET_AUTODISCOVERY` to find the right ip address
3. `HOSTNETWORK=true POD_IP=true make ceph` # the pod is bound to the `Pod InternalIP` 
4. `HOSTNETWORK=true MON_IP=<YOUR_HOST_IP_ADDRESS> NET_DISCOVERY=0 make ceph` # the pod uses `hostNetworking`and the container will be bound to the specified ip address
5. `NETWORKS_ANNOTATION="[{\"Name\":\"storage\",\"Namespace\":\"openstack\"}]" make ceph`  # attach the NAD to the POD provided it's precreated.
6. `NET_DISCOVERY=0 NETWORKS_ANNOTATION=\'[{\"Name\":\"storage\",\"Namespace\":\"openstack\",\"ips\":[\"172\.18\.0\.51\/24\"]}]\' MON_IP="172.18.0.51" make ceph`  # example of binding the Ceph Pod to the storage NAD

Signed-off-by: Francesco Pantano <fpantano@redhat.com>

### DIFF
--- a/scripts/gen-ceph-kustomize.sh
+++ b/scripts/gen-ceph-kustomize.sh
@@ -1,0 +1,213 @@
+#!/bin/bash
+#
+# Copyright 2022 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+# set -ex
+
+# expect that the common.sh is in the same dir as the calling script
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+. ${SCRIPTPATH}/common.sh --source-only
+
+if [ -z "$IMAGE" ]; then
+    echo "Unable to determine ceph image."
+    exit 1
+fi
+
+if [ -z "$NAMESPACE" ]; then
+    echo "Please set NAMESPACE."
+    exit 1
+fi
+
+if [ ! -d ${DEPLOY_DIR} ]; then
+    mkdir -p ${DEPLOY_DIR}
+fi
+
+pushd ${DEPLOY_DIR}
+
+TIMEOUT=${TIMEOUT:-30}
+POD_IP=${POD_IP:-false}
+HOSTNETWORK=${HOSTNETWORK:-false}
+NET_DISCOVERY=${NET_DISCOVERY:-"2"}
+POOLS=("volumes" "images" "backups")
+
+function add_ceph_pod {
+cat <<EOF >ceph-pod.yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  annotations:
+    k8s.v1.cni.cncf.io/networks: ""
+  name: ceph
+  namespace: openstack
+  labels:
+    app.kubernetes.io/name: ceph
+    app: ceph
+spec:
+  hostNetwork: $HOSTNETWORK
+  containers:
+   - image: quay.io/ceph/daemon:latest-quincy
+     name: ceph
+     env:
+     - name: MON_IP
+       value: "$MON_IP"
+     - name: NETWORK_AUTO_DETECT
+       value: "$NET_DISCOVERY"
+     - name: CEPH_DAEMON
+       value: demo
+     - name: CEPH_PUBLIC_NETWORK
+       value: "0.0.0.0/0"
+     - name: DEMO_DAEMONS
+       value: "osd"
+     volumeMounts:
+      - mountPath: /var/lib/ceph
+        name: data
+      - mountPath: /var/log/ceph
+        name: log
+      - mountPath: /run/ceph
+        name: run
+  volumes:
+  - name: data
+    emptyDir:
+      sizeLimit: 500Mi
+  - name: run
+    emptyDir:
+      sizeLimit: 500Mi
+  - name: log
+    emptyDir:
+      sizeLimit: 500Mi
+  securityContext:
+    runAsUser: 0
+    seccompProfile:
+      type: Unconfined
+EOF
+}
+
+function ceph_kustomize {
+cat <<EOF >kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ./ceph-pod.yaml
+namespace: openstack
+patches:
+- target:
+    kind: Pod
+  patch: |-
+    - op: replace
+      path: /metadata/namespace
+      value: $NAMESPACE
+    - op: replace
+      path: /spec/containers/0/image
+      value: $IMAGE
+    - op: replace
+      path: /metadata/annotations/k8s.v1.cni.cncf.io~1networks
+      value: $NETWORKS_ANNOTATION
+EOF
+}
+
+function get_pod_ip {
+    if [[ "$HOSTNETWORK" == "false" ]]; then
+        echo "HOSTNETWORK must be set to true!"
+        exit 1
+    fi
+    MON_IP=$(oc get nodes -o 'jsonpath={.items[*].status.addresses[?(.type=="InternalIP")].address}')
+    NET_DISCOVERY="0"
+}
+
+function ceph_is_ready {
+    echo "Waiting the cluster to be up"
+    until oc rsh ceph ls /etc/ceph/I_AM_A_DEMO &> /dev/null
+    do
+        sleep 1
+        echo -n .
+        (( TIMEOUT-- ))
+        [[ "$TIMEOUT" -eq 0 ]] && exit 1
+    done
+    echo
+}
+
+function create_pool {
+
+    [ "${#POOLS[@]}" -eq 0 ] && return;
+
+    for pool in "${POOLS[@]}"; do
+        oc rsh ceph ceph osd pool create $pool 4
+        oc rsh ceph ceph osd pool application enable $pool rbd
+    done
+}
+
+function build_caps {
+    local CAPS=""
+    for pool in "${POOLS[@]}"; do
+        caps="allow rwx pool="$pool
+        CAPS+=$caps,
+    done
+    echo "${CAPS::-1}"
+}
+
+function create_key {
+    local client=$1
+    local caps
+    local osd_caps
+
+    if [ "${#POOLS[@]}" -eq 0 ]; then
+        osd_caps="allow *"
+    else
+        caps=$(build_caps)
+        osd_caps="allow class-read object_prefix rbd_children, $caps"
+    fi
+    # do not log the key if exists
+    oc rsh ceph ceph auth get-or-create "$client" mgr "allow rw" mon "allow r" osd "$osd_caps" >/dev/null
+}
+
+function create_secret {
+
+    SECRET_NAME="$1"
+
+    TEMPDIR=`mktemp -d`
+    local client="client.openstack"
+    trap 'rm -rf -- "$TEMPDIR"' EXIT
+    echo 'Copying Ceph config files from the container to $TEMPDIR'
+    oc rsync ceph:/etc/ceph/ceph.conf $TEMPDIR
+    echo 'Create OpenStack keyring'
+    # we build the cephx openstack key
+    create_key "$client"
+    # do not log the exported key
+    echo 'Copying OpenStack keyring from the container to $TEMPDIR'
+    oc rsh ceph ceph auth export "$client" -o /etc/ceph/ceph.$client.keyring >/dev/null
+    oc rsync ceph:/etc/ceph/ceph.$client.keyring $TEMPDIR
+
+    echo "Replacing openshift secret $SECRET_NAME"
+    oc delete secret "$SECRET_NAME" -n $NAMESPACE 2>/dev/null || true
+    oc create secret generic $SECRET_NAME --from-file=$TEMPDIR/ceph.conf --from-file=$TEMPDIR/ceph.$client.keyring -n $NAMESPACE
+}
+
+## MAIN
+case "$1" in
+    "build")
+        [ "$POD_IP" == "true" ] && get_pod_ip;
+        add_ceph_pod
+        ceph_kustomize
+        kustomization_add_resources
+        ;;
+    "secret")
+        create_secret "ceph-conf-files"
+        ;;
+    "pools")
+        create_pool
+        ;;
+    "isready")
+        ceph_is_ready
+        ;;
+esac
--- a/Makefile
+++ b/Makefile
@@ -121,6 +121,9 @@ ANSIBLEEE_BRANCH    ?= main
 ANSIBLEEE           ?= config/samples/_v1alpha1_ansibleee.yaml
 ANSIBLEEE_CR        ?= ${OPERATOR_BASE_DIR}/ansibleee-operator/${ANSIBLEEE}
 
+# Ceph
+CEPH_IMG       ?= quay.io/ceph/daemon:latest-quincy
+
 # target vars for generic operator install info 1: target name , 2: operator name
 define vars
 ${1}: export NAMESPACE=${NAMESPACE}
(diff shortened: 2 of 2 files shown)


scripts/gen-ceph-kustomize.sh
```diff
@@ -0,0 +1,159 @@
...
+  namespace: openstack
+  labels:
+    app.kubernetes.io/name: ceph
+    app: ceph
+spec:
+  hostNetwork: $HOSTNETWORK
+  containers:
+   - image: quay.io/ceph/daemon:latest-quincy
+     name: ceph
+     env:
```
fmount, line 56: @Akrog I didn't find anything to push/force the FSID, and we use that value in the Cinder CR. I'm trying to think about a potential development flow, but I suspect we need something to build the cinder config snippet and inject it to the CR

fmount: long term I guess we can try adding the `--fsid` option to `ceph-mon` [1] when [2] is run and propose a change for the demo pod.
I'll follow up on that part too

[1] https://docs.ceph.com/en/quincy/man/8/ceph-mon/
[2] https://github.com/ceph/ceph-container/blob/main/src/daemon/start_mon.sh

scripts/gen-ceph-kustomize.sh
```diff
@@ -0,0 +1,174 @@
...
+function get_pod_ip {
+    if [[ "$HOSTNETWORK" == "false" ]]; then
+        echo "HOSTNETWORK must be set to true!"
+        exit 1
+    fi
+    MON_IP=$(oc get nodes -o 'jsonpath={.items[*].status.addresses[?(.type=="InternalIP")].address}')
+    NET_DISCOVERY="0"
+}
+
+function ceph_is_ready {
```
viroel, line 122: Can we consider adding a timeout for this function? we could fail earlier in CI with a timeout here.
One idea is to add timeout when calling it: "timeout 300 ceph_is_ready"

fmount: Makes sense to me, I'll rebase and add it!

viroel (commented):
I did a quick test in my SNO env, and seems to work, both 'ceph' and 'ceph_cleanup' targets.
I was able to add a ceph backend to both cinder-vol and cinder-bkp, which solved the issue o container crashing.

scripts/gen-ceph-kustomize.sh
```diff
@@ -0,0 +1,177 @@
...
+    echo "Waiting the cluster to be up"
+    until oc rsh ceph ls /etc/ceph/I_AM_A_DEMO &> /dev/null
+    do
+        sleep 1
+        echo -n .
+        (( TIMEOUT-- ))
```
fmount, line 129: @viroel so if you use `timeout <int> <function>`, it's not going to work because <function> isn't an exported command.
The following solution:

```
export -f ceph_is_ready
timeout <int> bash -c "ceph_is_ready" 
```
doesn't seem clean enough because of the subshell, so I went with a simple, easy way: a counter that is decremented each second and you can override the provided parameter.
I tested in my environment and it looks good, let me know if you have better solutions to this.

viroel: Agree, the proper solution is to add the the control var into the loop itself.

viroel (commented):
thanks for addint the timeout on wait function

fultonj (commented):
I think merging this is a good idea. It gives install_yamls users an easy way to have a ceph backend as well as get a Ceph cluster for CI. I don't see it as introducing any trouble since if someone doesn't want to use ceph they can simply not type `make ceph`. I'd like to make sure @dprince is OK with it before anyone merges it though.

I tested it and found it easy to `make ceph` and then update my glance pod so that it could easily write an image to the toy ceph cluster. E.g. since the `ceph-conf-files` secret existed already I just had to add an `extraMounts` and `customServiceConfig` to my glance CR.

One possible improvement to this PR is to include an update to the README showing how to use `make ceph`. E.g. run it and update the CR of a service to use the new backend with an example https://paste.opendev.org/show/b23BDFSo3Q1207nmfa91/

stuggi:
works fine for me, I'd like to extend it to be able to attach that pod to additional  networks via network attachment definition (needs to be created before). Using this right now for my isolated networks testing

```
diff --git a/scripts/gen-ceph-kustomize.sh b/scripts/gen-ceph-kustomize.sh
index f6c1b1c..2145e55 100644
--- a/scripts/gen-ceph-kustomize.sh
+++ b/scripts/gen-ceph-kustomize.sh
@@ -46,6 +46,8 @@ cat <<EOF >ceph-pod.yaml
 apiVersion: v1
 kind: Pod
 metadata:
+  annotations:
+    k8s.v1.cni.cncf.io/networks: ""
   name: ceph
   namespace: openstack
   labels:
@@ -108,6 +110,9 @@ patches:
     - op: replace
       path: /spec/containers/0/image
       value: $IMAGE
+    - op: replace
+      path: /metadata/annotations/k8s.v1.cni.cncf.io~1networks
+      value: '$NETWORKS_ANNOTATION'
 EOF
 }
```

So that you could run it with the following to reference a network attachement definition and get it attached to another network:

`NETWORKS_ANNOTATION='[{"Name":"storage","Namespace":"openstack"}]' make ceph`

@fmount let me know if you'd be ok with such an enhancement and if you want to do it as part of this PR or if I should submit a follow up

stuggi:
> works fine for me, I'd like to extend it to be able to attach that pod to additional networks via network attachment definition (needs to be created before). Using this right now for my isolated networks testing
> 
> ```
> diff --git a/scripts/gen-ceph-kustomize.sh b/scripts/gen-ceph-kustomize.sh
> index f6c1b1c..2145e55 100644
> --- a/scripts/gen-ceph-kustomize.sh
> +++ b/scripts/gen-ceph-kustomize.sh
> @@ -46,6 +46,8 @@ cat <<EOF >ceph-pod.yaml
>  apiVersion: v1
>  kind: Pod
>  metadata:
> +  annotations:
> +    k8s.v1.cni.cncf.io/networks: ""
>    name: ceph
>    namespace: openstack
>    labels:
> @@ -108,6 +110,9 @@ patches:
>      - op: replace
>        path: /spec/containers/0/image
>        value: $IMAGE
> +    - op: replace
> +      path: /metadata/annotations/k8s.v1.cni.cncf.io~1networks
> +      value: '$NETWORKS_ANNOTATION'
>  EOF
>  }
> ```
> 
> So that you could run it with the following to reference a network attachement definition and get it attached to another network:
> 
> `NETWORKS_ANNOTATION='[{"Name":"storage","Namespace":"openstack"}]' make ceph`
> 
> @fmount let me know if you'd be ok with such an enhancement and if you want to do it as part of this PR or if I should submit a follow up

well that is not complete. the ceph secret also needs to be updated.  lets get this in as is and follow up.

stuggi (approved):
/lgtm

fmount:
> works fine for me, I'd like to extend it to be able to attach that pod to additional networks via network attachment definition (needs to be created before). Using this right now for my isolated networks testing
> 
> ```
> diff --git a/scripts/gen-ceph-kustomize.sh b/scripts/gen-ceph-kustomize.sh
> index f6c1b1c..2145e55 100644
> --- a/scripts/gen-ceph-kustomize.sh
> +++ b/scripts/gen-ceph-kustomize.sh
> @@ -46,6 +46,8 @@ cat <<EOF >ceph-pod.yaml
>  apiVersion: v1
>  kind: Pod
>  metadata:
> +  annotations:
> +    k8s.v1.cni.cncf.io/networks: ""
>    name: ceph
>    namespace: openstack
>    labels:
> @@ -108,6 +110,9 @@ patches:
>      - op: replace
>        path: /spec/containers/0/image
>        value: $IMAGE
> +    - op: replace
> +      path: /metadata/annotations/k8s.v1.cni.cncf.io~1networks
> +      value: '$NETWORKS_ANNOTATION'
>  EOF
>  }
> ```
> 
> So that you could run it with the following to reference a network attachement definition and get it attached to another network:
> 
> `NETWORKS_ANNOTATION='[{"Name":"storage","Namespace":"openstack"}]' make ceph`
> 
> @fmount let me know if you'd be ok with such an enhancement and if you want to do it as part of this PR or if I should submit a follow up

Oh that's a good idea, we definitely want to do that!.
Let me update the patch with your suggestion!

fmount:
> > works fine for me, I'd like to extend it to be able to attach that pod to additional networks via network attachment definition (needs to be created before). Using this right now for my isolated networks testing
> > ```
> > diff --git a/scripts/gen-ceph-kustomize.sh b/scripts/gen-ceph-kustomize.sh
> > index f6c1b1c..2145e55 100644
> > --- a/scripts/gen-ceph-kustomize.sh
> > +++ b/scripts/gen-ceph-kustomize.sh
> > @@ -46,6 +46,8 @@ cat <<EOF >ceph-pod.yaml
> >  apiVersion: v1
> >  kind: Pod
> >  metadata:
> > +  annotations:
> > +    k8s.v1.cni.cncf.io/networks: ""
> >    name: ceph
> >    namespace: openstack
> >    labels:
> > @@ -108,6 +110,9 @@ patches:
> >      - op: replace
> >        path: /spec/containers/0/image
> >        value: $IMAGE
> > +    - op: replace
> > +      path: /metadata/annotations/k8s.v1.cni.cncf.io~1networks
> > +      value: '$NETWORKS_ANNOTATION'
> >  EOF
> >  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So that you could run it with the following to reference a network attachement definition and get it attached to another network:
> > `NETWORKS_ANNOTATION='[{"Name":"storage","Namespace":"openstack"}]' make ceph`
> > @fmount let me know if you'd be ok with such an enhancement and if you want to do it as part of this PR or if I should submit a follow up
> 
> well that is not complete. the ceph secret also needs to be updated. lets get this in as is and follow up.

Not sure I follow here, I mean, if you cleanup ceph and redeploy the secret is recreated, which is an easy pattern to follow.
If we're ok with this change I propose to merge and follow up on any interesting work we can do!

fmount:
> > > works fine for me, I'd like to extend it to be able to attach that pod to additional networks via network attachment definition (needs to be created before). Using this right now for my isolated networks testing
> > > ```
> > > diff --git a/scripts/gen-ceph-kustomize.sh b/scripts/gen-ceph-kustomize.sh
> > > index f6c1b1c..2145e55 100644
> > > --- a/scripts/gen-ceph-kustomize.sh
> > > +++ b/scripts/gen-ceph-kustomize.sh
> > > @@ -46,6 +46,8 @@ cat <<EOF >ceph-pod.yaml
> > >  apiVersion: v1
> > >  kind: Pod
> > >  metadata:
> > > +  annotations:
> > > +    k8s.v1.cni.cncf.io/networks: ""
> > >    name: ceph
> > >    namespace: openstack
> > >    labels:
> > > @@ -108,6 +110,9 @@ patches:
> > >      - op: replace
> > >        path: /spec/containers/0/image
> > >        value: $IMAGE
> > > +    - op: replace
> > > +      path: /metadata/annotations/k8s.v1.cni.cncf.io~1networks
> > > +      value: '$NETWORKS_ANNOTATION'
> > >  EOF
> > >  }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > So that you could run it with the following to reference a network attachement definition and get it attached to another network:
> > > `NETWORKS_ANNOTATION='[{"Name":"storage","Namespace":"openstack"}]' make ceph`
> > > @fmount let me know if you'd be ok with such an enhancement and if you want to do it as part of this PR or if I should submit a follow up
> > 
> > 
> > well that is not complete. the ceph secret also needs to be updated. lets get this in as is and follow up.
> 
> Not sure I follow here, I mean, if you cleanup ceph and redeploy the secret is recreated, which is an easy pattern to follow. If we're ok with this change I propose to merge and follow up on any interesting work we can do!

The good news is that provided `NADs` exist:

```
> oc get network-attachment-definitions
NAME          AGE
internalapi   7m19s
storage       7m17s
```

the following:

```
> NETWORKS_ANNOTATION=\'[{\"Name\":\"storage\",\"Namespace\":\"openstack\"}]\' make ceph
bash scripts/gen-namespace.sh

namespace/openstack unchanged
sleep 2
oc project openstack

oc kustomize /home/fmount/devnull/osp-k8s-operators/dev_operator/install_yamls/out/openstack/ceph/cr | oc apply -f -
pod/ceph created
bash scripts/gen-ceph-kustomize.sh "isready"

Waiting the cluster to be up
.............

bash scripts/gen-ceph-kustomize.sh "pools"
pool 'volumes' created
enabled application 'rbd' on pool 'volumes'
pool 'images' created
enabled application 'rbd' on pool 'images'
pool 'backups' created
enabled application 'rbd' on pool 'backups'

bash scripts/gen-ceph-kustomize.sh "secret"
Copying Ceph config files from the container to $TEMPDIR
ceph.conf
ceph.client.admin.keyring

Replacing openshift secret ceph-conf-files
secret "ceph-conf-files" deleted
secret/ceph-conf-files created
```

sees the `storage network` attached (via multus) to the Ceph Pod:

```
Events:
  Type    Reason          Age   From               Message
  ----    ------          ----  ----               -------
  Normal  Scheduled       14s   default-scheduler  Successfully assigned openstack/ceph to crc-pbwlw-master-0 by crc-pbwlw-master-0
  Normal  AddedInterface  13s   multus             Add eth0 [10.217.0.130/23] from openshift-sdn
  Normal  AddedInterface  13s   multus             Add net1 [172.18.0.50/24] from openstack/storage
  Normal  Pulled          13s   kubelet            Container image "quay.io/ceph/daemon:latest-quincy" already present on machine
  Normal  Created         12s   kubelet            Created container ceph
  Normal  Started         12s   kubelet            Started container ceph
```

which means this patch actually works with the recent changes we introduced.
@stuggi If we're able to force a `static ip` we can actually bind the pod to the `multus network`, but we'll sync up on this and follow up. 
@dprince would you mind reviewing it? It should be good to go, including the multus support.

stuggi:
@fmount yes it works like you did to set a fixed IP. in your example, have you also set it as the `POD_IP= <YOUR_HOST_IP_ADDRESS>` ?
```
      - name: MON_IP
       value: "172.18.0.51"
```

fmount:
> @fmount yes it works like you did to set a fixed IP. in your example, have you also set it as the `POD_IP= <YOUR_HOST_IP_ADDRESS>` ?
> 
> ```
>       - name: MON_IP
>        value: "172.18.0.51"
> ```

```
> NETWORKS_ANNOTATION=\'[{\"Name\":\"storage\",\"Namespace\":\"openstack\",\"ips\":[\"172\.18\.0\.51\/24\"]}]\' make ceph
```

> @fmount yes it works like you did to set a fixed IP. in your example, have you also set it as the `POD_IP= <YOUR_HOST_IP_ADDRESS>` ?
> 
> ```
>       - name: MON_IP
>        value: "172.18.0.51"
> ```

This is the command:

`NET_DISCOVERY=0 NETWORKS_ANNOTATION=\'[{\"Name\":\"storage\",\"Namespace\":\"openstack\",\"ips\":[\"172\.18\.0\.51\/24\"]}]\' MON_IP="172.18.0.51" make ceph`

`POD_IP=true` is meant for `hostNetworking` and it's used to get the host internal network w/o specifying any ip.
The command above, instead, just bind the `Ceph Pod` to the `IP` provided as input.

ASBishop:
I haven't studied the pr yet, but will we support an "openstack" cephx client? The earlier PoC stuff I had seen used the admin client, which I think we should avoid.

fmount:
> I haven't studied the pr yet, but will we support an "openstack" cephx client? The earlier PoC stuff I had seen used the admin client, which I think we should avoid.

We can improve the script and add it: we can then export the `client.ceph.openstack.keyring` instead of using the `admin` one.
Given that people are not reviewing this PR yet, I think it's reasonable adding that part too!
@ASBishop I'll point you to the code in my next update!

fultonj:
> > I haven't studied the pr yet, but will we support an "openstack" cephx client? The earlier PoC stuff I had seen used the admin client, which I think we should avoid.
> 
> We can improve the script and add it: we can then export the `client.ceph.openstack.keyring` instead of using the `admin` one. Given that people are not reviewing this PR yet, I think it's reasonable adding that part too! @ASBishop I'll point you to the code in my next update!

Thanks Fracnesco. Yes, I think it's a good idea to switch this to `client.ceph.openstack.keyring`.

scripts/gen-ceph-kustomize.sh
```diff
@@ -0,0 +1,213 @@
...
+    SECRET_NAME="$1"
+
+    TEMPDIR=`mktemp -d`
+    local client="client.openstack"
+    trap 'rm -rf -- "$TEMPDIR"' EXIT
+    echo 'Copying Ceph config files from the container to $TEMPDIR'
+    oc rsync ceph:/etc/ceph/ceph.conf $TEMPDIR
+    echo 'Create OpenStack keyring'
+    # we build the cephx openstack key
+    create_key "$client"
```
fmount, line 185: @ASBishop @fultonj instead of using the admin keyring, we're building a new `client.openstack` key and updating the secret accordingly.
We can expose more parameters, but in order to keep things simple I'd just use the `openstack` user. In case we need more we can follow up.

fultonj (commented):
/lgtm thanks Francesco

dprince (approved):

PR #56, opened 2023-01-19
Improved edpm_compute_cleanup target

raukadah:
It also removes:
- compute pub keys
- secrets
- It removes '-' from command to keep consistent.

Since the ansibleee-operator is renamed to openstack-ansibleee-operator. It changes the same.

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -50,8 +50,8 @@ edpm_compute:
 
 .PHONY: edpm_compute_cleanup
 edpm_compute_cleanup:
-	sudo virsh destroy edpm-compute-${EDPM_COMPUTE_SUFFIX}
-	sudo virsh undefine edpm-compute-${EDPM_COMPUTE_SUFFIX}
+	-sudo virsh destroy edpm-compute-${EDPM_COMPUTE_SUFFIX}
+	-sudo virsh undefine edpm-compute-${EDPM_COMPUTE_SUFFIX}
 	rm -f ${HOME}/.crc/machines/crc/edpm-compute-${EDPM_COMPUTE_SUFFIX}.qcow2
 	rm -f ../out/edpm/edpm-compute-*-id_rsa.pub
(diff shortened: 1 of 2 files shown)


rebtoor (approved):

abays (approved):
/lgtm

devsetup/Makefile
```diff
@@ -50,14 +50,16 @@ edpm_compute:
 
 .PHONY: edpm_compute_cleanup
 edpm_compute_cleanup:
-	-sudo virsh destroy edpm-compute-${EDPM_COMPUTE_SUFFIX}
-	-sudo virsh undefine edpm-compute-${EDPM_COMPUTE_SUFFIX}
+	sudo virsh destroy edpm-compute-${EDPM_COMPUTE_SUFFIX}
+	sudo virsh undefine edpm-compute-${EDPM_COMPUTE_SUFFIX}
```
slagle, line 54: The dash ("-") in front of those commands meant they could fail (non-zero exit) and make wouldn't complain. I think we still need that, or at least a check to see if the vm's are running before we destroy and undefine them.

devsetup/Makefile
```diff
@@ -50,14 +50,16 @@ edpm_compute:
...
+	oc delete -f ../out/edpm/ansibleee-ssh-key-secret.yaml
 
 .PHONY: edpm_play
 edpm_play:
 	oc create -f edpm/edpm-play.yaml
 
 .PHONY: edpm_play_cleanup
 edpm_play_cleanup:
-	-oc delete ansibleee deploy-external-dataplane-compute
+	oc delete ansibleee deploy-external-dataplane-compute
```
slagle, line 65: Ditto.

slagle (changes requested):

devsetup/Makefile
raukadah: Thank you for highlighting the use of dash here. I have added it back in the new commit.

PR #57, opened 2023-01-19
Removed emulator section in the vm definition since it's pointless

rebtoor:

Path specified with `command -v qemu-kvm` doesn't work on CentOS since `qemu-kvm` binary is located under `/usr/libexec`. BTW, if we avoid to insert that line, the `virsh define` command seems to take care of generate emulator entry with the default binary path.


rebtoor:
Just deployed the edpm-compute-0 on a CentOS-9 hypervisor, that line is ignored.
`[stack@localhost devsetup]$ sudo virsh dumpxml edpm-compute-0 |grep emulator
    <emulator>/usr/libexec/qemu-kvm</emulator>
`

PR #58, opened 2023-01-19
Sync delete-pv.sh with changes to create-pv.sh

jistr:
This brings delete-pv.sh up to date with changes to create-pv.sh from these PRs:

https://github.com/openstack-k8s-operators/install_yamls/pull/40 https://github.com/openstack-k8s-operators/install_yamls/pull/45

Most importantly the directories in create-pv are now named using format like `pv01` while delete-pv was still using the previous `pv1` format, so the delete-pv script wasnt cleaning up some of the PV directories previously. This is now fixed.

### DIFF
--- a/scripts/delete-pv.sh
+++ b/scripts/delete-pv.sh
@@ -16,9 +16,11 @@
 set -ex
 PV_NUM=${PV_NUM:-12}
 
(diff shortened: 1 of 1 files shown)


stuggi (approved):
/lgtm

PR #59, opened 2023-01-20
Add cleanup after running the kuttl tests

Sandeepyadav93:
Currently, We are leaving behind Openstack Operators and Openstack deployment which we creates during setup for kuttl.

ex.
When we execute `make keystone_kuttl` we install mariadb/keystone operator and mariadb service which are leftovers once keystone kuttl tests finishes.

This change make sure we cleanup Openstack services and Operatos once kuttl tests finishes.


SeanMooney:
so i think this si wrong in general
the kuttl target should not install or clean up any operators

it should run the test in a new namespace and then after that is complete the namespace should be removed
ideally install yamls would just invoke a kuttl make target in the operator to do that and you would not need to orchestrate cleaning up this via install yamls

the reason for that is we would idelly be able to use the kuttl test in a dev context

i.e. as a develop of keysonte i should be able to deploy vai the meta operator and disable keysone operator in the csv.

i should then be able to run my operator under a debugger and run the kuttl test with make kuttl and test that the change i am making do not break anything.

so the kuttl make target should be the interface to running the tests and cleaning up

the install yaml bit if it exist shoudl be a thin wrapper over that but the kuttl envs should not depend on deploying or cleaning up the operators or crs.

in ci in particular we will need to build the operator under test and use that version.

so i think this make target is currently doing to much and adding better cleanup is just a symptom of that.

the kuttel tests in repo shoudl be cleaning up and if we want an extra level of safety we shoudl delete the tempory namespace created for running it so that k8s cleans up any dangling resources.


to me as a developer of a service operator its important to me to be able to run the kuttle test without it impacting my dev environment and potentially doing this sharing a single CRC or real openshfit cluster between them 

if we include deploy_cleanup in the depencies i cannot use this target form install_yamls

maybe that is ok and as an operator-developer i should just use make kuttl in my repo but it will mean we have different workflows. as a developer i want to maintain one way of running the kuttl test and use make files as that interface so i don't want to have to have a second way of doing that with install yamls if we can avoid that.

cescgina:
As discussed on slack, I added a new commit that splits the kuttl targets in two: one with the same name (e.g. `keystone_kuttl`), that has no dependencies nor cleanup and assumes that user will have deployed everything needed beforehand, and a new ci one (e.g. `keystone_kuttl_ci_run`) which has all the dependencies and clenaup added to be used in ci jobs or when the user does only want to run the tests and does not want to maintain the previous state of the cluster

marios:
was there a different pull request opened for the namespace change across the kuttl tests i recall @Sandeepyadav93 mentioned that on yesterdays ci scrum

Makefile
```diff
@@ -697,14 +697,22 @@ nova_deploy_cleanup: ## cleans up the service instance, Does not affect the oper
 ##@ KUTTL tests
 
 .PHONY: mariadb_kuttl
-mariadb_kuttl: namespace input openstack_crds deploy_cleanup mariadb_deploy_prep mariadb  ## runs kuttl tests for the mariadb operator. Installs openstack crds and keystone operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
+mariadb_kuttl: input ## runs kuttl tests for the mariadb operator, assumes that everything needed for running the test was deployed beforehand.
```
marios, line 700: lgtm. just a thought on the name, perhaps it is better to keep mariadb_kuttl as the 'do everything target' and have a special one for devs that just want the tests like "mariadb_kuttl_only"  so it is more descriptive

cescgina:
> was there a different pull request opened for the namespace change across the kuttl tests i recall @Sandeepyadav93 mentioned that on yesterdays ci scrum

@marios not that I'm aware of, maybe @Sandeepyadav93 can tell you more

Makefile
cescgina: Yeah, maybe that makes more sense, changing

Sandeepyadav93:
@marios this was the patch which I was talking about, I pushed the initial commit and @cescgina took over after comments from Sean.

Makefile
```diff
@@ -696,14 +696,27 @@ nova_deploy_cleanup: ## cleans up the service instance, Does not affect the oper
 
 ##@ KUTTL tests
 
-.PHONY: mariadb_kuttl
-mariadb_kuttl: namespace input openstack_crds deploy_cleanup mariadb_deploy_prep mariadb  ## runs kuttl tests for the mariadb operator. Installs openstack crds and keystone operators and cleans up previous deployments before running the tests.
+.PHONY: mariadb_kuttl_run
+mariadb_kuttl_run: input ## runs kuttl tests for the mariadb operator, assumes that everything needed for running the test was deployed beforehand.
```
gibizer, line 700: Is there a specific reason we kept `input` as dependency here?

Makefile
```diff
@@ -696,14 +696,27 @@ nova_deploy_cleanup: ## cleans up the service instance, Does not affect the oper
...
-.PHONY: keystone_kuttl
-keystone_kuttl: namespace input openstack_crds deploy_cleanup mariadb mariadb_deploy mariadb_deploy_validate keystone_deploy_prep keystone ## runs kuttl tests for the keystone operator. Installs openstack crds and keystone operators and cleans up previous deployments before running the tests.
+.PHONY: mariadb_kuttl
+mariadb_kuttl: namespace input openstack_crds deploy_cleanup mariadb_deploy_prep mariadb  ## runs kuttl tests for the mariadb operator. Installs openstack crds and keystone operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
+	make mariadb_kuttl_run
+	make deploy_cleanup
+	make mariadb_cleanup
+
+.PHONY: keystone_kuttl_run
+keystone_kuttl_run: input ## runs kuttl tests for the keystone operator, assumes that everything needed for running the test was deployed beforehand.
```
gibizer, line 710: ditto

Makefile
cescgina: I kept it because I saw that it's the only dependency that is set for all *_deploy targets, so I thought it would be more consistent, but can be removed.

gibizer: I think we can remove it. In the case when the user directly running `mariadb_kuttl_run` and therefore installed all the prerequisites I assume the secret is also created by the user.

cescgina: Makes sense, done.

gibizer (commented):
Looks good to me.  I haven't approved it yet to let others reviewed before to check the current version.

abays (approved):
/lgtm

PR #62, opened 2023-01-24
Add labels to default namespace

bshephar:
Add labels to default namespace
  
  This change is necessary due to new admission controller standards in kubernetes:
  https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-namespace-labels/
  
  We need to ensure we can still run workloads within our default namespace. So this change adds
  two labels to the default namespace that will allow us to continue working within the new standards.
  
  See the following issue for more information:
  https://issues.redhat.com/browse/OSP-21710
  
  Impact of the k8s change on OpenShift and related workloads:
  https://github.com/redhat-openshift-ecosystem/community-operators-prod/discussions/1417

### DIFF
--- a/scripts/gen-namespace.sh
+++ b/scripts/gen-namespace.sh
@@ -35,4 +35,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
     name: ${NAMESPACE}
+    labels:
+      pod-security.kubernetes.io/enforce: privileged
+      security.openshift.io/scc.podSecurityLabelSync: "false"
 EOF_CAT


karelyatin (commented):
looks fine as a workaround but need to consider the recommended ways for deployments on 4.12.

bshephar:
Yeah, we realistically need to address a few things related to default SCC's within all Operators. For example, we run all of our containers with UID 0, which requires anyuid SCC, or privileged. So I think the scope will be fairly big for us to completely comply with OCP SCC standards. Maybe we can create a story with sub-tasks for each operator to assess the required changes.

For example, we need to see if any operators create pods directly. But I think it requires more research before making any changes to the operators directly. For now, I think it's best to keep the status quo, allow privileged on our namespace and we can take our time to make the right design decisions for each operator.

I've added a note to discuss in our next meeting.

dprince (approved):

PR #63, opened 2023-01-25
Ansible role wrapper around install_yamls Makefile

raukadah:
It contains:
- Add phony target for crc_storage and crc_storage cleanup
- Adds the ansible role wrapper around install_yamls Makefile

In order to use the same interface irrespective of consumer like CI/DEV/QE. This patch uses the Install_yamls makefile defined variables and phony target to generate roles default vars and template around that using python code.

If a consumer adds a new interface to the Makefile, he can run the scripts and get the updated role.

Once the ansible role execution finishes, It generates the script and CR file which can be used in any supported OpenShift cluster to re-run the same steps executed in CI.

The same role is consumed in Zuul Based job:
https://review.rdoproject.org/r/c/testproject/+/46569/76/playbooks/deploy_podified_control_plane.yaml#11

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -1,0 +1,117 @@
+# Ansible role wrapper around install_yamls Makefile
+
+Install_yamls makefile can be used by Dev/CI/QE as a single interface to
+deploy NextGen OpenStack.
+
+use_install_yamls role is a ansible wrapper around the install_yamls Makefile.
+It uses scripts/makefile_to_install_yamls.py script to generate the following
+ansible role specific files.
+
+    * defaults/main.yaml
+    * templates/install_yamls.sh.j2
+    * vars/command_after_make_target.yaml
+
+## Why we need a wrapper?
+
+Developer, CI and QE are the three consumers of install_yamls tool.
+Their OpenStack deployment environment might be different based on their needs.
+Each of the consumers might be exporting different variables, running different
+make commands and other commands in between multiple make commands to setup
+the openstack deployment.
+
+If something went wrong, we need a way to reproduce the same environment easily.
+Since Ansible is the widely used everywhere to drive the deployment. It is easy to use.
+Having a wrapper around the tool allows each of the consumers to use the same role
+and create a playbook to achieve the desired outcome.
+
+When the playbook will be called, It is going to generate `install_yamls.sh` script
+containing all the exported vars, commands.
+
+One can use the same script to reproduce in the enviroment.
+
+## What is the requirements in order to consume it?
+
+We need a working openshift cluster, kubeconfig file and ansible installed.
+
+## How we keep the wrapper role up to date?
+
+Once any consumers adds a new interface to the makefile, they can run following
+commands to sync the wrapper role.
+```
+$ cd <path_to_install_yamls>/devsetup/
+$ make ansible_role_sync
+cd ../ci/roles/use_install_yamls; rm -f defaults/main.yaml vars/command_after_make_target.yaml templates/install_yamls.sh.j2;
+cd ../ci/scripts; python makefile_to_install_yamls.py ../../Makefile;
+tree ../ci/roles/use_install_yamls;
+../ci/roles/use_install_yamls
+├── defaults
+│   └── main.yaml
+├── tasks
+│   └── main.yaml
+├── templates
+│   └── install_yamls.sh.j2
+└── vars
+    └── command_after_make_target.yaml
+
+5 directories, 4 files
+```
+
+## How the wrapper role works?
+
+The wrapper role works on following assumptions.
+
+* All the Install_yamls Makefile vars are defined in smallcase under defaults/main.yaml.
+* In order to run any `make <target>` command, we need add `run_` as a prefix to the target name
+  and set the var as `run_<target>: true` in the playbook.
+
+  For example: We want to run `make openstack` command, then the ansible var would be
+  `run_openstack: true`.
+
+* This wrapper role also provides functionality to run custom commands after a make command.
+  We need to add `command_after_make_<target>` as a var and add the command below that.
+
+  For Example: We want to run 'oc get csv' after `make openstack` command. Then we need to
+  define following var.
+  command_after_make_openstack: |
+    oc get csv
+
+* If multiple cleanup commands are called, then order will be reversed based on the sequence
+  defined in the Makefile
+
+## Example playbook to consume the role
+
+```
+❯ cat example_playbook.yaml
+---
+- hosts: localhost
+  tasks:
+    - include_role:
+        name: use_install_yamls
+      vars:
+        namespace: openstack
+        openstack_img: quay.rdoproject.org/openstack-k8s-operators/openstack-operator-index:latest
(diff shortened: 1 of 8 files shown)


ci/README.md
```diff
@@ -0,0 +1,54 @@
...
+Install_yamls makefile can be used by Dev/CI/QE as a single interface to
+deploy NextGen OpenStack.
+
+use_install_yamls role is a ansible wrapper around the install_yamls Makefile.
+It uses scripts/makefile_to_install_yamls.py script to generate the default vars
+and template file for use_install_yamls role.
+
+Run the following command to see it in action
+```
+$ rm ../roles/use_install_yamls/defaults/main.yaml ../roles/use_install_yamls/templates/install_yamls.sh.j2
```
cjeanner, line 12: where are we supposed be in the path in order to run this? imho, the "../" is wrong, seeing the L15, apparently running from ci/scripts.

pinikomarov:
hi @raukadah ,I up for anything that simplifies and consolidates templates, so a couple of questions about this : 

1. Isn't the regular [current install yamls Makefile](https://github.com/openstack-k8s-operators/install_yamls/blob/master/Makefile#L66) + a few lines of code + make all command enough ? do we need a wrapper around something which is not that complex? 
2. Can you give a user use case which demonstrates this usage is better than install yamls current , as stated in 1) ?
3. You are parsing a WIP dev Makefile , while it's possible that and other files/templates may change, are you accounting for that? 
4. This is creating a seperate role in Install_yamls , which may put it onto your team (no the devs) to maintain. Are you prepared to take care - maintain this role going forward (bugs , user support) etc..?
5. Why not integrate into install yamls ? why not take this opinionated template consolidating logic of this template and change the way install yamls deploys - integrate this into install yamls ?

raukadah:
Hello @pinikomarov,

Thank you for taking the time to look at the pull request.
Based on feedback received, I have cleaned up the wrapper role by making following changes:
- Improve crc_storage command to create custom storage - https://github.com/openstack-k8s-operators/install_yamls/pull/79
- Manipulate openstack-services containers urls in openstack control plane CRD - https://github.com/openstack-k8s-operators/install_yamls/pull/81

Now the current pull request contains the code which generates the wrapper around the install_yamls Makefile with proper
README: https://github.com/raukadah/install_yamls/blob/ci_code/ci/README.md

Thank you for these questions. below are the inline answers for the same.

> hi @raukadah ,I up for anything that simplifies and consolidates templates, so a couple of questions about this :
> 
> 1. Isn't the regular [current install yamls Makefile](https://github.com/openstack-k8s-operators/install_yamls/blob/master/Makefile#L66) + a few lines of code + make all command enough ? do we need a wrapper around something which is not that complex?

The purpose of creating this ansible wrapper around install_yamls makefile is to provide a single role which can be used to create playbook/task contains all the vars needed to finish the OSP next gen deployment. After the playbook execution finishes, It will generate
the script with all the vars, make commands and other commands runned between that.

Any user can take that script and reproduce it in any given openshift cluster. One can use the same playbook in downstream/upstream CI also for the same job
without duplicating the code.

By thinking from reusability and reproducing the environment point of view, We need a wrapper around the install_yamls.
I tried to explain in detail in README: https://github.com/raukadah/install_yamls/blob/ci_code/ci/README.md

> 2. Can you give a user use case which demonstrates this usage is better than install yamls current , as stated in 1) ?

When we started creating the CI job for podified and EDPM testing, I was writting multiple shell tasks to run various commands.
Because of this wrapper, we can simply call the role and pass the vars and I have podified deployment up. In the end if deployment
fail, we can simply run the install_yamls.sh script to recreate the env.

If We want to reduplicate the same job downstream, we donot need to write multiple shell tasks, we simply use the wrapper role to achieve
the same. It will be true for other jobs as we extend our testing scope.

> 3. You are parsing a WIP dev Makefile , while it's possible that and other files/templates may change, are you accounting for that?

if any interface in Makefile changes, the developer or any change proposer can run 'make ansible_role_sync' command, it will update the wrapper role.
We do not need to add or update anything in the role.

> 4. This is creating a seperate role in Install_yamls , which may put it onto your team (no the devs) to maintain. Are you prepared to take care - maintain this role going forward (bugs , user support) etc..?

We are not creating a separate role. It is a wrapper around the install_yamls makefile which can be automatically updated based on the interface added or removed from
the makefile by running 'make ansible_role_sync' can update it. We donot need an extra effort to maintain it.

However each of the service operator and their integration in install_yamls have their own owners. The CI framework can take care of it.

> 5. Why not integrate into install yamls ? why not take this opinionated template consolidating logic of this template and change the way install yamls deploys - integrate this into install yamls ?

it is a wrapper around install_yamls which is tightly integrated with install_yamls Makefile. Everything can be controlled via Makefile vars (converted into ansible vars by wrapper.) . Apart from that custom commands can be also added between make targets, can be stored in playbook. One can use the same playbook to avoid duplicacy.

I hope it answers the above queries.

ci/scripts/makefile_to_install_yamls.py
```diff
@@ -0,0 +1,105 @@
+# Python script to automatically generate roles content for install_yamls
```
cjeanner, line 1: While I fully support the concept of wrapper, I have some concerns:

since the goal is to call the various "make" from within ansible, via a single generated file, we'll get a hanging task for the whole time of the run. We'll therefore lose track of what's going on, and may lost important pieces in the process.

What about:
- generate an env-file with the needed ```export``` of the various environment variables we can set
- generate per-make call scripts, sourcing the generated env-file, maybe with some order prefix (1-foo, 2-bar, ...) in a dedicated directory
- call each of those script within ansible (we can loop on the directory content, no problem here)

That way, we'll get a more agile approach, more modular, and would allow to, maybe generate things but not run all of them (think "breakpoints support"). We'll also be able to properly manage the needed pauses, calling the related ```oc wait --for ... <resource>``` instead of static "sleep" in the script.

WDYT?

raukadah: Thank you @cjeanner for taking a look at the code. I liked the idea of moving stuffs in different files like:
* Moving the user exported vars in a single file
* Generating one file per make target
     * Then we can easily club it together based on user's need
It will make the role more modular and maintainable.

All these above stuffs can be achieved by the wrapper script. It will generate the role for us. We can also add custom logic where ever needed. Let me go ahead and modify my script and see how it goes.

pinikomarov: @raukadah Thanks for the answers , I'm +1 for @cjeanner's comment on running the individual generated makes from ansible and not bash.
Other than that I understand that some complexity in maintenance is necessary for this wrapper to be usable and reproducible across different envs. So I say let's go for it and see how it goes , nice job:)

I'm ok with this , please ask the core guys to approve this PR (Dan,Andrew,etc..)

abays: I'm cool with whatever you guys agree upon here, as long as it doesn't break the existing direct usage of the Makefile.

raukadah:
Thank you everyone for the feedback, we are moving this work here https://github.com/openstack-k8s-operators/ci-framework/pull/65 . I am closing this pr.

PR #66, opened 2023-01-30
Bump golang/operator-sdk

rabi:
We've started to use latest stable golang and sdk in operators[1].

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/149

### DIFF
--- a/devsetup/roles/download_tools/defaults/main.yaml
+++ b/devsetup/roles/download_tools/defaults/main.yaml
@@ -1,17 +1,17 @@
 ---
 # kuttl version to use (must be specific version)
-kuttl_version: 0.9.0
+kuttl_version: 0.15.0
 
 # Released version of the opm package (can be set to 'latest')
 opm_version: latest
 
 # operator-sdk version to use (must be specific version)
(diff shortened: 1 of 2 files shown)


devsetup/roles/download_tools/defaults/main.yaml
```diff
@@ -7,10 +7,10 @@ opm_version: latest
 
 # operator-sdk version to use (must be specific version)
 #sdk_version: v0.19.2 - cnosp is right now based on that version
-sdk_version: v1.14.0
+sdk_version: v1.26.0
 
 # golang version
-go_version: 1.16.9
+go_version: 1.19.0
```
abays, line 13: Should we have this match what is in `vars/default.yaml`?  I know the `vars` setting actually "wins", but perhaps they should be identical.

rabi: We probably should not have it multiple places, but updated.

abays (approved):
/lgtm

dprince (approved):

PR #71, opened 2023-01-31
crc: Make disk size tunable

kajinamit:
The default size (31GiB) is sometimes small. This allows tuning root disk size using an environment variable, similarly to cpus and memory.


stuggi (approved):
/lgtm

PR #72, opened 2023-01-31
Remove unused cache options for nova-compute

kajinamit:
The nova-compute service does not use caching backed by memcached. This removes the unused variables from edpm playbook to avoid further confusion.


slagle (approved):

SeanMooney:
ya the compute shoudl not be using the cache backend.

I'm hoping wee can remove all the nova related part form this soon and instead just create the novaexternalCR via the dataplane operator.

SeanMooney (approved):

gibizer (commented):
Users of install_yamls is encouraged to deploy the compute via `make edpm_deploy` instead of `make edpm_play`. The latter is kept temporary to allow CI to catch up with the work we did in the dataplane-operator and nova-operator. The edpm-play.yaml is expected to go away very soon.

PR #75, opened 2023-01-31
Fix the installation of the operator openstack-ansibleee

rebtoor:

Since we're changed the name of the operator from `ansibleee-operator` to `openstack-ansibleee-operator`, the `vars` function was invocated with the old name, so when the gen-olm.sh script was invocated during the `ansibleee_prep` target, it created the CatalogSource and Subscription resources with wrong naming, so the operator installation was stuck with the error:

`constraints not satisfiable: no operators found in package ansibleee-operator`


fao89 (approved):
Thank you!

fmount (approved):
+1 thanks

PR #77, opened 2023-02-01
Add provisioned-by annotation for crc local storage

rabi:
Local storage pvs created for crc go to "Failed" status after deleting service instances using them with "Error getting deleter volume plugin for volume "xxx": no deletable volume plugin". This is because deleting pvc tries to delete the pvs.

This would also allow for the pvs to be reused after patching.

$oc patch pv local-storage12 -p '{"spec":{"claimRef": null}}'

### DIFF
--- a/scripts/gen-crc-pv-kustomize.sh
+++ b/scripts/gen-crc-pv-kustomize.sh
@@ -42,6 +42,8 @@ kind: PersistentVolume
 apiVersion: v1
 metadata:
   name: local-storage$i
+  annotations:
+    pv.kubernetes.io/provisioned-by: crc-devsetup
 spec:
   storageClassName: local-storage
   capacity:


abays:
Will that patch obviate the need for this?

`for i in $(oc get pv | egrep "Failed|Released" | awk {'print $1'}); do oc patch pv $i --type='json' -p='[{"op": "remove", "path": "/spec/claimRef"}]'; done`

I find myself running that a lot.  It would be great if the PVs didn't fail on `make <service>_deploy_cleanup`.

rabi:
Probably not completely as the claimRef would still stay with the pv (and needs to be patched/cleared), but they would be in 'Released' status and not Failed with error like before.

abays (approved):
/lgtm

PR #81, opened 2023-02-02
Use Kustomize to update openstack containers in CR

raukadah:
In order to use containers from different registeries with different
hashes, we needed a way to update the OpenstackControlPlane CR file.

This patch uses kustomize to achieve the same.

For example: By just doing
```
export SERVICE_REGISTRY=quay.rdoproject.org
```
The target registry will be changed to quay.rdoproject.org in OpenstackControlPlane file
when make openstack_deploy command is invoked.

It also users from not creating their own CRs.


raukadah:
@stuggi @abays @cjeanner @pinikomarov please have a look, thank you :-)

Makefile
```diff
@@ -241,6 +267,7 @@ openstack_deploy_prep: openstack_deploy_cleanup ## prepares the CR to install th
 	pushd ${OPERATOR_BASE_DIR} && git clone -b ${OPENSTACK_BRANCH} ${OPENSTACK_REPO} && popd
 	cp ${OPENSTACK_CR} ${DEPLOY_DIR}
 	bash scripts/gen-service-kustomize.sh
+	bash scripts/gen-service-container-kustomize.sh
```
abays, line 270: One thing I would point out is that calling `gen-service-container-kustomize.sh` after `gen-service-kustomize.sh` will overwrite the `kustomization.yaml` from the former.  That YAML was already consumed and is not technically needed anymore, but perhaps someone might want to look at its contents for debugging purposes after the fact.  There are a few different ways to avoid this if we care here, but I just thought I'd mention it.

raukadah: Thank you @abays for pointing out, yes kustomization.yaml is getting overwritten. I agree with you to keep the content for debugging purpose. Please suggest the changes in order to avoid this.

abays: One suggestion might be to merge `gen-service-container-kustomize.sh` into `gen-service-kustomize.sh`, unless you think there might be scenarios where we would be executing the logic in one but would not necessarily want to use the logic of the other.  I think it makes semantic sense to have both kustomization processes under one `gen-service-kustomize.sh` script, because it's all ultimately targeting the OpenStack service we're deploying.

chkumar246: That sounds good. I have made the changes. Please have a look.

raukadah: That sounds good. I have made the changes. Please have a look.

Makefile
```diff
@@ -146,6 +157,21 @@ ${1}: export OUT=${OUT}
...
+export CINDERAPI_IMG=${CINDERAPI_IMG}
+export CINDERSCHEDULER_IMG=${CINDERSCHEDULER_IMG}
+export CINDERBACKUP_IMG=${CINDERBACKUP_IMG}
+export CINDERVOLUME_IMG=${CINDERVOLUME_IMG}
+export OVNBDS_IMG=${OVNBDS_IMG}
+export OVSBDS_IMG=${OVSBDS_IMG}
+export OVNNORTHD_IMG=${OVNNORTHD_IMG}
+export OVSSERVICE_IMG=${OVSSERVICE_IMG}
+export OVNCONTROLLER_IMG=${OVNCONTROLLER_IMG}
+export NEUTRONSERVER_IMG=${NEUTRONSERVER_IMG}
```
abays, line 174: Should we perhaps follow the pattern used in the existing vars definitions, i.e. `${1}: export VAR=${VAR}`.  I'm not sure off the top of my head what the functional difference is between that and what we've added here, but I just thought I'd ask.

raukadah: Done

raukadah:
Thank you @gibizer @abays for reviewing it, I have made the necessary changes. Please have a look. :-)

raukadah:
Added depends-on: https://github.com/openstack-k8s-operators/openstack-operator/pull/173 otherwise we will hit 
```
error: replace operation does not apply: doc is missing path: /spec/nova/template/apiServiceTemplate/containerImage: missing value
```

Makefile
```diff
@@ -87,14 +94,18 @@ NEUTRON_BRANCH     ?= master
...
 CINDER_IMG       ?= quay.io/openstack-k8s-operators/cinder-operator-index:latest
 CINDER_REPO      ?= https://github.com/openstack-k8s-operators/cinder-operator.git
 CINDER_BRANCH    ?= master
 CINDER           ?= config/samples/cinder_v1beta1_cinder.yaml
 CINDER_CR        ?= ${OPERATOR_BASE_DIR}/cinder-operator/${CINDER}
-# TODO: Image customizations for all Cinder services
+CINDERAPI_IMG	 ?= ${SERVICE_REGISTRY}/${SERVICE_ORG}/${PREFIX}-cinder-api:${CONTAINER_TAG}
+CINDERSCHEDULER_IMG ?= ${SERVICE_REGISTRY}/${SERVICE_ORG}/${PREFIX}-cinder-schedular:${CONTAINER_TAG}
+CINDERBACKUP_IMG ?= ${SERVICE_REGISTRY}/${SERVICE_ORG}/${PREFIX}-cinder-backup:${CONTAINER_TAG}
+CINDERVOLUME_IMG ?= ${SERVICE_REGISTRY}/${SERVICE_ORG}/${PREFIX}-cinder-volume:${CONTAINER_TAG}
```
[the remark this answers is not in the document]
raukadah: I have added the nova related vars in Makefile.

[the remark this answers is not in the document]
gibizer: thanks

gibizer (commented):
Looks OK from nova perspective. I will let @abays approve it as he had other comments too

abays (approved):
I see a few indentation nits that could be cleaned-up in the `Makefile`, but no big deal

/lgtm

PR #82, opened 2023-02-03
Make compute ip configurable

kajinamit:
... instead of requiring users to edit the playbook file.

Signed-off-by: Takashi Kajinami <tkajinam@redhat.com>

### DIFF
--- a/devsetup/README.md
+++ b/devsetup/README.md
@@ -92,31 +92,38 @@ After completing the devsetup, attach the crc VM to the default network:
 make crc_attach_default_interface
 ```
 
-Deploy additional VM's for compute nodes:
+Deploy a compute node VM:
 ```
 # Creates edpm-compute-0:
 make edpm_compute
-# Set $EDPM_COMPUTE_SUFFIX to create additional VM's beyond 0:
-make edpm_compute EDPM_COMPUTE_SUFFIX=1
 ```
 
-Edit edpm/edpm-play.yaml and set the compute node VM IP addresses in the
-inventory. The IP address in the inventory (192.168.122.139) needs to be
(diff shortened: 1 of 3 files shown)


kajinamit:
We might want to allow users to inject the whole inventory, but I think this can be a good improvement for users playing edpm with a single compute node.

raukadah:
+1 to this improvement, It will help me to drop a complete role from edpm-ansible job.

raukadah (approved):

rebtoor:
@kajinamit ++

devsetup/README.md
```diff
@@ -92,31 +92,38 @@ After completing the devsetup, attach the crc VM to the default network:
...
+Execute the ansible to configure the compute node:
 ```
-make edpm_play
+make edpm_play EDPM_COMPUTE_IP=<IP>
 ```
 
+You can also deploy additional compute node VMs:
+```
+# Set $EDPM_COMPUTE_SUFFIX to create additional VM's beyond 0:
+make edpm_compute EDPM_COMPUTE_SUFFIX=1
```
rebtoor, line 114: we should also use `EDPM_COMPUTE_IP` variable to deploy multiple vms.

rebtoor (changes requested):

devsetup/README.md
kajinamit: EDPM_COMPUTE_IP is not used by edpm_compute.
Also, it does not work because users should update a few more items in inventory (such as ips). That's why I suggested modifying the playbook directly as automating generation of inventory is quite complicated.

rebtoor (approved):

PR #83, opened 2023-02-03
enable ssh traffic from the ansibleee container to the compute node

rebtoor:

Since we're going to [import](https://github.com/openstack-k8s-operators/edpm-ansible/pull/29) and enable the nftables role for the compute host, we should enable ssh traffic to the compute node from the ansibleee container.

### DIFF
--- a/devsetup/edpm/edpm-play.yaml
+++ b/devsetup/edpm/edpm-play.yaml
@@ -172,6 +172,9 @@ spec:
                 edpm_ovn_controller_image: quay.io/tripleomastercentos9/openstack-ovn-controller:current-tripleo
                 gather_facts: false
                 enable_debug: false
+                # edpm firewall, change the allowed CIDR if needed
+                edpm_sshd_configure_firewall: true
+                edpm_sshd_allowed_ranges: [ '192.168.122.0/24' ]
                 # SELinux module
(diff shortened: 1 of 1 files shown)


devsetup/edpm/edpm-play.yaml
```diff
@@ -172,6 +172,9 @@ spec:
                 edpm_ovn_controller_image: quay.io/tripleomastercentos9/openstack-ovn-controller:current-tripleo
                 gather_facts: false
                 enable_debug: false
+                # edpm firewall, change the allowed CIDR if needed
+                edpm_sshd_configure_firewall: true
```
kajinamit, line 176: Where are these variables implemented ? I don't see these in the files we have in this repo.

kajinamit (changes requested):

devsetup/edpm/edpm-play.yaml
rebtoor: They are needed by the nftables module that we want to import, see [this](https://github.com/openstack-k8s-operators/edpm-ansible/pull/29) PR.

kajinamit: Sorry I totally missed the link.

kajinamit (approved):

PR #88, opened 2023-02-06
Add new parameters to gen-ansibleee-ssh-key.sh

cjeanner:
In order to get rid of some ansible role in RDO, we need to be able to reuse an existing SSH key.

This patch adds the possibility to set the existing key name, its algorithm and key size.

The defaults values are the ones (save the default key size, sets to 4096 instead of 3072).

### DIFF
--- a/devsetup/scripts/gen-ansibleee-ssh-key.sh
+++ b/devsetup/scripts/gen-ansibleee-ssh-key.sh
@@ -18,7 +18,11 @@ set -ex
 # expect that the common.sh is in the same dir as the calling script
 SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 NAMESPACE=${NAMESPACE:-"openstack"}
+METADATA_NAME=${METADATA_NAME:-"ansibleee-ssh-key-secret"}
 OUTPUT_DIR=${OUTPUT_DIR:-"../out/edpm"}
+SSH_ALGORITHM=${SSH_ALGORITHM:-"rsa"}
+SSH_KEY_FILE=${SSH_KEY_FILE:-"ansibleee-ssh-key-id_rsa"}
+SSH_KEY_SIZE=${SSH_KEY_SIZE:-"4096"}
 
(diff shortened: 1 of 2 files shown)


devsetup/scripts/gen-ansibleee-ssh-key.sh
```diff
@@ -18,7 +18,11 @@ set -ex
 # expect that the common.sh is in the same dir as the calling script
 SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 NAMESPACE=${NAMESPACE:-"openstack"}
+METADATA_NAME=${METADATA_NAME:-"ansibleee-ssh-key-secret"}
```
raukadah, line 21: It is not used with in the script, May be we can add at L44.

raukadah (changes requested):
Looks good, left few comments!

devsetup/scripts/gen-ansibleee-ssh-key.sh
cjeanner: ah, right, forgot that one.

devsetup/scripts/gen-ansibleee-ssh-key.sh
```diff
@@ -33,8 +37,8 @@ if oc get secret ansibleee-ssh-key-secret 2>&1 1>/dev/null; then
     exit 0
 fi
 
-if [ ! -f ansibleee-ssh-key-id_rsa ]; then
-    ssh-keygen -f ansibleee-ssh-key-id_rsa -N ""
+if [ ! -f $SSH_KEY_FILE ]; then
+    ssh-keygen -f SSH_KEY_FILE -N "" -t ${SSH_ALGORITHM} -b ${SSH_KEY_SIZE}
```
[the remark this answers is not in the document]
cjeanner: hmm. yeah. we may do this. lemme correct it for the sake of consistency.

raukadah:
@slagle @rebtoor please have a look, thank you :-)

rebtoor (approved):

PR #91, opened 2023-02-07
Fix 'make crc_storage' generation of PV YAMLs

abays:
The existing generation now errors-out because it generates PVs with `metadata.name` with an invalid format like so: `name: "local-storage"01`.  Also: perhaps there is a cleaner way to approach this fix than what I have proposed here.

### DIFF
--- a/scripts/gen-crc-pv-kustomize.sh
+++ b/scripts/gen-crc-pv-kustomize.sh
@@ -45,7 +45,7 @@ cat >> out/crc/storage.yaml <<EOF_CAT
 kind: PersistentVolume
 apiVersion: v1
 metadata:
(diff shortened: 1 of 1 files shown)


jparoly (approved):
I have tested out this change locally and it works great.  Thank you!

gibizer:
@abays thanks for fixing this. I guess we could drop the `"` from https://github.com/openstack-k8s-operators/install_yamls/blob/be8095d964e21c54484ab34df4568939d09086ec/Makefile#L17 but the current fix is OK as well. 

I'm more interested in getting a deploy job running on install_yamls repo as a bug in this repo can affect a lot of other jobs / people and this is not the first time we merged broken code in install_yamls.

PR #96, opened 2023-02-09
Start all operations from the script directory

slagle:
In order to be able to execute the script standalone, from any initial
working directory, make sure to switch to the script directory first.


devsetup/scripts/gen-ansibleee-ssh-key.sh
```diff
@@ -57,3 +59,4 @@ EOF
 oc create -f ansibleee-ssh-key-secret.yaml
 
 popd
+popd
```
dprince, line 62: @slagle nested, popd's? Maybe we should get rid of pushd/popd and just use a clean 'cd' to whatever directory approach instead

PR #97, opened 2023-02-10
Add targets for ovs-operator kuttl tests

cescgina:
Add targets to run the kuttl tests from the ovs operator.

Depends-on: https://github.com/openstack-k8s-operators/ovs-operator/pull/60


Makefile
```diff
@@ -81,6 +81,8 @@ OVS_REPO            ?= https://github.com/openstack-k8s-operators/ovs-operator.g
 OVS_BRANCH          ?= main
```
marios, line 81: would be nice to point to https://github.com/openstack-k8s-operators/ovs-operator/pull/60 in the commit message

i know we don't have depends-on here but it would be useful even for the reviewer

marios (commented):
looks good didn't spot something & follows the same pattern as the others (_kuttl_run)

karelyatin (approved):
/lgtm

PR #99, opened 2023-02-13
Add targets for neutron-operator kuttl tests

cescgina:
Add targets to run the kuttl tests from the neutron operator.

Tests can be found in: https://github.com/openstack-k8s-operators/neutron-operator/pull/119

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,8 @@ NEUTRON_BRANCH     ?= master
 NEUTRONAPI         ?= config/samples/neutron_v1beta1_neutronapi.yaml
 NEUTRONAPI_CR      ?= ${OPERATOR_BASE_DIR}/neutron-operator/${NEUTRONAPI}
 NEUTRONAPI_IMG     ?= ${SERVICE_REGISTRY}/${SERVICE_ORG}/openstack-neutron-server:current-tripleo
+NEUTRON_KUTTL_CONF ?= ${OPERATOR_BASE_DIR}/neutron-operator/kuttl-test.yaml
+NEUTRON_KUTTL_DIR  ?= ${OPERATOR_BASE_DIR}/neutron-operator/test/kuttl/tests
 
 # Cinder
 CINDER_IMG       ?= quay.io/openstack-k8s-operators/cinder-operator-index:latest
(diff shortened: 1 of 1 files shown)


abays (commented):
Are there any Neutron KUTTL tests yet in the Neutron operator repo?

cescgina:
> Are there any Neutron KUTTL tests yet in the Neutron operator repo?

Just pushed them and added a link in the PR body, you were too fast :)

abays:
One thing I've noticed that is not specific to this PR, but instead seems to be a concern for all KUTTL tests, is that we assume `local-storage` is the storage class being used.  Is there a way we could support variable storage classes?  I noticed this problem when I tried running this Neutron KUTTL test and noticed that the `MariaDB` CR assertion was looking for `storageClass: local-storage`.

cescgina:
> One thing I've noticed that is not specific to this PR, but instead seems to be a concern for all KUTTL tests, is that we assume `local-storage` is the storage class being used. Is there a way we could support variable storage classes? I noticed this problem when I tried running this Neutron KUTTL test and noticed that the `MariaDB` CR assertion was looking for `storageClass: local-storage`.

Kuttl gives no flexibility when checking, we will need an exact match, use an script or we can avoid the check for `storageClass`. As far as I recall, we only check for it in MariaDB, and those tests were quickly added, so maybe it's not really that important?

cescgina:
> > > One thing I've noticed that is not specific to this PR, but instead seems to be a concern for all KUTTL tests, is that we assume `local-storage` is the storage class being used. Is there a way we could support variable storage classes? I noticed this problem when I tried running this Neutron KUTTL test and noticed that the `MariaDB` CR assertion was looking for `storageClass: local-storage`.
> > 
> > 
> > Kuttl gives no flexibility when checking, we will need an exact match, use an script or we can avoid the check for `storageClass`. As far as I recall, we only check for it in MariaDB, and those tests were quickly added, so maybe it's not really that important?
> 
> My real concern would be regarding CI environments and what storage class they would be using. If it's always just `local-storage`, then I suppose it's not a big deal.

I've opened a PR to remove the line: https://github.com/openstack-k8s-operators/mariadb-operator/pull/70

abays (approved):
/lgtm

PR #103, opened 2023-02-14
use patchesStrategicMerge to update default containers value

raukadah:
Signed-off-by: Chandan Kumar <raukadah@gmail.com>

PR #104, opened 2023-02-14
Reverting PR #81

gibizer:
This PR reverts #81 as it turned out that change prevents the usage of non fully populated sample files. For example if a sample file set a service disabled then it should not be required to list the containerImage for that service.

The issue is caused by the limitation of the `replace` strategy of kustomize that cannot handle more than one level of missing keys in the input. A possible way forward is to use the patchesStrategicMerge strategy instead.


gibizer:
lets test #103 first if it works then we can close this

gibizer:
#103 has some issues so we did the revert instead.

PR #105, opened 2023-02-14
Create secret for webhooks to work

rabi:


dprince:
@rabi hey, so I think we should be able to solve this by merging the webhookdefinitions in our CSV. As OLM can automate the creation of these I don't think we should need to have this code here right?

See: https://github.com/openstack-k8s-operators/openstack-operator/pull/181

PR #113, opened 2023-02-21
Update secret generation for dataplane-operator

slagle:
Depends on
https://github.com/openstack-k8s-operators/dataplane-operator/pull/30

The dataplane-operator now takes a Secret name that contains the private
key contents. This change updates install_yamls to match the sample
Secret name from the config/samples in dataplane-operator, and to create
the Secret with the correct format.

This change also updates the devsetup/scripts/gen-ansibleee-ssh-key.sh
script to change to the script directory intially, so that it can be run
directly from the git checkout, no matter the working directory.

### DIFF
--- a/devsetup/scripts/gen-ansibleee-ssh-key.sh
+++ b/devsetup/scripts/gen-ansibleee-ssh-key.sh
@@ -18,7 +18,7 @@ set -ex
 # expect that the common.sh is in the same dir as the calling script
 SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 NAMESPACE=${NAMESPACE:-"openstack"}
-METADATA_NAME=${METADATA_NAME:-"ansibleee-ssh-key-secret"}
+METADATA_NAME=${METADATA_NAME:-"dataplane-ansible-ssh-private-key-secret"}
 OUTPUT_DIR=${OUTPUT_DIR:-"../../out/edpm"}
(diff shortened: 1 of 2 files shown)


fao89 (approved):

devsetup/scripts/gen-ansibleee-ssh-key.sh
```diff
@@ -18,42 +18,45 @@ set -ex
...
 kind: Secret
 namespace: ${NAMESPACE}
 metadata:
     name: ${METADATA_NAME}
 data:
-    public_ssh_key: |
+    ssh_public_key: |
 $(cat ${SSH_KEY_FILE}.pub | base64 | sed 's/^/        /')
-    private_ssh_key: |
+    ssh_private_key: |
```
fao89, line 55: if https://github.com/openstack-k8s-operators/dataplane-operator/pull/54 is merged first, we would need to change
```suggestion
    ssh-privatekey: |
```

slagle: thanks, we can hold off on this PR until https://github.com/openstack-k8s-operators/dataplane-operator/pull/54 merges.

fao89:
need to update:
https://github.com/openstack-k8s-operators/install_yamls/blob/master/devsetup/edpm/edpm-play.yaml#L205-L207

dprince (approved):

fao89 (approved):

PR #117, opened 2023-02-22
Install virt-customize correctly on Cent/RHEL8

fultonj:
Update gen-edpm-compute-node.sh so that it can install virt-customize correclty on either RHEL8 or RHEL9. For RHEL9 virt-customize is provided by guestfs-tools. For RHEL8 virt-customize is provided by libguestfs-tools-c. Without this patch the script fails when run on RHEL8.

### DIFF
--- a/devsetup/scripts/gen-edpm-compute-node.sh
+++ b/devsetup/scripts/gen-edpm-compute-node.sh
@@ -141,8 +141,13 @@ if [ ! -f ${DISK_FILEPATH} ]; then
         popd
     fi
     qemu-img create -f qcow2 -F qcow2 -b centos-9-stream-base.qcow2 ${DISK_FILEPATH} 20G
-    if ! rpm -q guestfs-tools; then
-        sudo dnf -y install guestfs-tools
+    if [[ ! -e /usr/bin/virt-customize ]]; then
+        if [[ $(awk '{print $6}' /etc/redhat-release) =~ ^8.* ]]; then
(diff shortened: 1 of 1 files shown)


devsetup/scripts/gen-edpm-compute-node.sh
```diff
@@ -141,8 +141,13 @@ if [ ! -f ${DISK_FILEPATH} ]; then
         popd
     fi
     qemu-img create -f qcow2 -F qcow2 -b centos-9-stream-base.qcow2 ${DISK_FILEPATH} 20G
-    if ! rpm -q guestfs-tools; then
-        sudo dnf -y install guestfs-tools
+    if [[ ! -e /usr/bin/virt-customize ]]; then
+        if [[ $(awk '{print $6}' /etc/redhat-release) =~ ^8.* ]]; then
+            sudo dnf -y install hexedit libguestfs-tools-c
```
fultonj, line 146: `hexedit` is in the list of packages to install as a workaround because `libguestfs-tools-c` alone produced the following error.
```
[fultonj@hamfast install_yamls]$ sudo dnf install libguestfs-tools-c
Updating Subscription Management repositories.
Last metadata expiration check: 2:58:20 ago on Wed 22 Feb 2023 06:21:29 AM EST.
Error: 
 Problem: cannot install the best candidate for the job
  - nothing provides /usr/bin/hexedit needed by libguestfs-tools-c-1:1.44.0-3.el8s.x86_64
```

PR #124, opened 2023-03-02
Add check to EDPM compute DHCP Reservation

fultonj:
Add condition so that `virsh net-update default add-last ip-dhcp-host` is only run if the MAC address does not already have an entry.

Running `make edpm_compute_cleanup; make edpm_compute` more than once on the same hypervisor produces the error below unless this patch is in place.
```
+ sudo virsh net-update default add-last ip-dhcp-host \ --xml '<host mac='\''52:54:00:C1:A3:7F'\'' \ name='\''edpm-compute-0'\'' ip='\''192.168.122.100'\''/>'\ --config --live

error: Failed to update network default
error: Requested operation is not valid: there is an existing dhcp host entry in network 'default' that matches "<host mac='52:54:00:C1:A3:7F' name='edpm-compute-0'
  ip='192.168.122.100'/>"
```

### DIFF
--- a/devsetup/scripts/gen-edpm-compute-node.sh
+++ b/devsetup/scripts/gen-edpm-compute-node.sh
@@ -179,8 +179,9 @@ if [ ! -f ${DISK_FILEPATH} ]; then
         exit 1
     fi
 fi
-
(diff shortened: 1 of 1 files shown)


fao89 (approved):

fao89:
I just hit this issue!
Thank you for the fix

PR #128, opened 2023-03-06
Pass all the edpm container images as var

raukadah:
The CI team is working on designing the periodic line for EDPM and podified testing. There openstack services containers will be built and pushed to different registry with different namespace and dlrn container tag.

In order to consume the same, We need to break the container image url into multiple var so that we can override it.


rebtoor (commented):
+1 for me

PR #129, opened 2023-03-06
add support for microshfit

SeanMooney:
micoroshift does not support the openshift user and project extensions to the k8s API.
to workaround this we implement the missing functionality when the namespace is created.


Makefile
```diff
@@ -233,7 +233,8 @@ namespace: ## creates the namespace specified via NAMESPACE env var (defaults to
 	bash scripts/gen-namespace.sh
 	oc apply -f ${OUT}/${NAMESPACE}/namespace.yaml
 	sleep 2
-	oc project ${NAMESPACE}
+	oc project ${NAMESPACE} || oc config set-context --current --namespace=${NAMESPACE}
+	oc adm policy add-scc-to-user privileged -z default --namespace ${NAMESPACE}
```
dprince, line 237: won't this mean we always allow (and test with) this policy change/addition. Seems like if it is required for microshift you'd only want to do it there.

dprince (dismissed):

Makefile
SeanMooney: ya so i was just commiting the experiment i did over the weekend
What i was thinking of doing was adding a microshift varible that is set to false by default and putting this in an if so that we would only do this if set to true.

oc adm policy add-scc-to-user privileged -z default --namespace ${NAMESPACE}

is required for olm to be able to install the operators as the install job uses users that are not in the default allowed set.

oc config set-context --current --namespace=${NAMESPACE}
does the equivalent of oc project ${NAMESPACE}

although i don't know if either are actually required since the namespace is set in the env it does not hurt anything to do this.

https://github.com/openstack-k8s-operators/install_yamls/blob/master/Makefile#L176-L186

ill respin this making 
oc adm policy add-scc-to-user privileged -z default --namespace ${NAMESPACE}
conditional but I'm only really looking at this in sparetime so i don't plan on working on it for a few days

danpawlik: something like this ?
`oc project ${NAMESPACE} || oc config set-context --current --namespace=${NAMESPACE}
if oc project microshift; then oc adm policy add-scc-to-user privileged -z default --namespace ${NAMESPACE}; fi`

SeanMooney: i was thinking more like 
```
if $MICROSHIFT; oc adm policy add-scc-to-user privileged -z default --namespace ${NAMESPACE} ; fi
```

and then you would set MICROSHIFT=True when you run make

SeanMooney: i just need to lookup the make if syntax and test it 

i also consider having it do 

 if $micorshift:
   oc config set-context --current --namespace=${NAMESPACE}
   oc adm policy add-scc-to-user privileged -z default --namespace ${NAMESPACE} 
 else:
   oc project ${NAMESPACE} 

i just need to lookup the make file if/else syntax and test it.

SeanMooney: apparently make is really annoying an indention changes the meaning of ifeq

if its dedented its parsed as a make directive if i indent it its treated as bash.
that took way to much time to figure out yesterday because i did not see that documented anywhere. anyway i think i have adressed this feedback so ill resolve this commeent theread.

danpawlik (approved):

raukadah (approved):
Tested here: https://github.com/openstack-k8s-operators/install_yamls/pull/135, works for microshift

arxcruz (approved):

stuggi (approved):
/lgtm

PR #130, opened 2023-03-06
Ability to deploy a control plane with galera or mariadb

dciabrin:
Deploy galera as a database with:

   make DBSERVICE=galera openstack openstack_deploy

This defaults to a 1-node galera, but it can be
configured to deploy a 3-node cluster with:

   make DBSERVICE=galera DBREPLICAS=3 make openstack openstack_deploy

By default, the control plane is still deployed with mariadb


scripts/gen-service-kustomize.sh
```diff
@@ -71,6 +72,23 @@ cat <<EOF >>kustomization.yaml
...
+      path: /spec/replicas
+      value: ${DBREPLICAS}
+EOF
+fi
+
+if [ "$NAME" = "openstackcontrolplane" ] && [ -n "${DBREPLICAS:-}" ]; then
+cat <<EOF >>kustomization.yaml
+    - op: replace
+      # only update the main galera DB for the time being
+      path: /spec/galera/templates/openstack/replicas
```
abays, line 87: This will cause a `kustomize` error if for some reason `DBREPLICAS` is set and yet all or part of the `path` is missing in the source `OpenStackControlPlane` YAML.  Though I suppose that would mostly be user error since they would have to be explicitly providing the former and excluding the latter.

dciabrin: Right... I don't see a fast and simple means to ensure this path is always present in the resource that is getting built by 'oc kustomize' without running it in gen-service-kustomize.
What about hardening the check by explicitly requiring DBSERVICE=galera and live with that until we deprecate MariaDB in favour of Galera?

dprince: would it be better to just have an alternate config/sample in openstack-operator that configures Galara instead? Also, I think as much as possible having the default we set via openstack-operator (code can live in mariadb operator and get applied at creation time) would be ideal

dciabrin: So you mean no tweak of DBREPLICAS, and instead a single config for 3-node galera? And eventually when we deprecate mariadb in favor of galera we update all mariadb config files to use 1-node galera instead?

I'm good with that, I just like to have a simple means to deploy the openstack-operator with galera, so the DBSERVICE env var seems reasonable until we deprecate mariadb.

raukadah:
recheck

raukadah:
recheck

raukadah:
recheck

scripts/gen-service-kustomize.sh
dciabrin: > would it be better to just have an alternate config/sample in openstack-operator that configures Galara instead? Also, I think as much as possible having the default we set via openstack-operator (code can live in mariadb operator and get applied at creation time) would be ideal

I just updated the PR select different config samples for 1-node galera (default) or 3-node galera. It depends on https://github.com/openstack-k8s-operators/openstack-operator/pull/279.

@dprince is that what you had in mind?

dciabrin:
recheck

raukadah:
recheck

dciabrin:
rebased to fix merge conflicts

danpawlik:
recheck

danpawlik:
checking pull requests on new CRC image base on version 2.17

danpawlik:
recheck

danpawlik:
recheck

Makefile
```diff
@@ -23,11 +29,27 @@ NETWORK_ISOLATION   ?= true
...
+GALERA_REPLICAS         ?=
+
 # OpenStack Operator
 OPENSTACK_IMG                ?= quay.io/openstack-k8s-operators/openstack-operator-index:latest
 OPENSTACK_REPO               ?= https://github.com/openstack-k8s-operators/openstack-operator.git
 OPENSTACK_BRANCH             ?= main
-OPENSTACK_CTLPLANE           ?= $(if $(findstring true,$(NETWORK_ISOLATION)),config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml,config/samples/core_v1beta1_openstackcontrolplane.yaml)
+ifeq ($(NETWORK_ISOLATION), true)
+ifeq ($(DBSERVICE), galera)
+OPENSTACK_CTLPLANE           ?= core_v1beta1_openstackcontrolplane_galera_network_isolation.yaml
```
abays, line 41: This needs to be prepended with `config/samples/`

Makefile
```diff
@@ -23,11 +29,27 @@ NETWORK_ISOLATION   ?= true
...
 
+# set to 3 to use a 3-node galera sample
+GALERA_REPLICAS         ?=
+
 # OpenStack Operator
 OPENSTACK_IMG                ?= quay.io/openstack-k8s-operators/openstack-operator-index:latest
 OPENSTACK_REPO               ?= https://github.com/openstack-k8s-operators/openstack-operator.git
 OPENSTACK_BRANCH             ?= main
-OPENSTACK_CTLPLANE           ?= $(if $(findstring true,$(NETWORK_ISOLATION)),config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml,config/samples/core_v1beta1_openstackcontrolplane.yaml)
+ifeq ($(NETWORK_ISOLATION), true)
```
abays, line 39: We were previously using `if $(findstring true,$(NETWORK_ISOLATION))` instead of `ifeq`.  Just wondering whether the former choice was deliberate to avoid a certain problem.

Makefile
dciabrin: Oops, thanks! Fixed.

Makefile
dciabrin: I moved the condition to top-level for clarity, and I believe GNU make has no if statements for top-level conditionals, only ifeq/ifneq/ifdef variants. The $(if $(findstring ) ...) idiom is for inline conditional expansion.

dciabrin:
> LGTM. Eventually I wonder if we should just get rid of MariaDB entirely and just always use Galera

Let's deprecate mariadb once galera is merged and is known to work for everyone?

Makefile
abays: Ack

abays (approved):
/lgtm

PR #137, opened 2023-03-10
Add make edpm_compute_repos

fultonj:
After running `make edpm_compute` it is useful to be able to run `make edpm_compute_repos` which installs https://git.openstack.org/openstack/tripleo-repos on the new node and runs it with current-tripleo-dev.

The edpm-ansible roles, which run on an edpm-compute, expect to be able to `dnf install os-net-config` and similar. This new option is a quick way to satisfy that requirement.


raukadah (approved):
Thank you @fultonj for adding this change. Let me fix the same in CI side.

PR #142, opened 2023-03-13
Allow to point to arbitrary CDR for edpm deploy

cjeanner:
With the ci-framework, we will be generating a custom CRD. In order to still be able to leverage `make edpm_play`, we have to make the actual CRD a parameter to `make`.

This patch allows to call either `make edpm_play`, using the default CRD, or `make edpm_play EDPM_PLAY_CRD=~/my-custom-crd.yml`.

Note that the `sed` call doesn't change, so we expect the custom CRD to have the same kind of parameters.


raukadah (approved):
Thank you for mentioning about sed call in the commit message.

rebtoor (approved):

PR #143, opened 2023-03-13
Add support for ceilometer-operator

vyzigold:
This adds the necessary make targets to deploy ceilometer. Part of this PR is also a modified gen-service-kustomize.sh script, because the ceilometer uses 3 container images instead of just one like all of the other operators.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,6 +183,16 @@ CEPH_IMG       ?= quay.io/ceph/demo:latest
 # NNCP
 NNCP_INTERFACE ?= enp6s0
 
+# Ceilometer
+CEILOMETER_IMG                 ?= quay.io/openstack-k8s-operators/ceilometer-operator-index:main-latest
+CEILOMETER_REPO                ?= https://github.com/openstack-k8s-operators/ceilometer-operator.git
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -183,6 +183,16 @@ CEPH_IMG       ?= quay.io/ceph/demo:latest
 # NNCP
 NNCP_INTERFACE ?= enp6s0
 
+# Ceilometer
+CEILOMETER_IMG                 ?= quay.io/openstack-k8s-operators/ceilometer-operator-index:main-latest
```
jlarriba, line 187: :latest was pushed to the repository as of yesterday, so I think we can change it to match it

jlarriba: Oh no, the -index:latest was not generated, only the ceilometer-operator. So :latest does not exist for this image and currently it is called :main-latest

jlarriba (changes requested):

Makefile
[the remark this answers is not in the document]
vyzigold: I changed it to :latest and verified, that it works. Thanks

PR #144, opened 2023-03-14
Replace TripleO images by TCIB images

kajinamit:
We imported container build tools from TripleO as TCIB[1]. Now we are retiring TripleO and should complete the migration.

[1] https://github.com/openstack-k8s-operators/tcib


kajinamit:
Probably https://github.com/openstack-k8s-operators/install_yamls/pull/145 is a better approach

PR #145, opened 2023-03-14
Remove capability to override images

kajinamit:
The current behavior to override the image by default is can result in an expected behavior, especially when someone wants to test a change which updates the image name in sample CR definition.

Some components (eg. nova) do not support customizing images which is also confusing.

In addition, the current implementation requires that we maintain the image names in Makefile to keep these consistent with sample files in individual repo, but this requirement is redundant.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,10 +9,6 @@ INSTALL_YAMLS       ?= ${PWD}  # used for kuttl tests
 # operators gets cloned here
 OPERATOR_BASE_DIR   ?= ${OUT}/operator
 
-# default registry and org to pull service images from
-SERVICE_REGISTRY    ?= quay.io
-SERVICE_ORG         ?= tripleozedcentos9
-
 # storage (used by some operators)
 STORAGE_CLASS       ?= "local-storage"
 
@@ -32,7 +28,6 @@ INFRA_BRANCH     ?= master
 # Memcached
 MEMCACHED           ?= config/samples/memcached_v1beta1_memcached.yaml
 MEMCACHED_CR        ?= ${OPERATOR_BASE_DIR}/infra-operator/${MEMCACHED}
-MEMCACHED_IMG       ?= ${SERVICE_REGISTRY}/${SERVICE_ORG}/openstack-memcached:current-tripleo
 
 # Keystone
 KEYSTONE_IMG        ?= quay.io/openstack-k8s-operators/keystone-operator-index:latest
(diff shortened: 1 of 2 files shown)


dprince:
@kajinamit Is there another PR where you are hitting difficulties that motivated you to create this patch? Also, how is nova-operator behaving differently here (it shouldn't be!?)

Makefile
```diff
@@ -32,15 +28,13 @@ INFRA_BRANCH     ?= master
...
 MEMCACHED_CR        ?= ${OPERATOR_BASE_DIR}/infra-operator/${MEMCACHED}
-MEMCACHED_IMG       ?= ${SERVICE_REGISTRY}/${SERVICE_ORG}/openstack-memcached:current-tripleo
 
 # Keystone
 KEYSTONE_IMG        ?= quay.io/openstack-k8s-operators/keystone-operator-index:latest
 KEYSTONE_REPO       ?= https://github.com/openstack-k8s-operators/keystone-operator.git
 KEYSTONE_BRANCH     ?= master
 KEYSTONEAPI         ?= config/samples/keystone_v1beta1_keystoneapi.yaml
 KEYSTONEAPI_CR      ?= ${OPERATOR_BASE_DIR}/keystone-operator/${KEYSTONEAPI}
-KEYSTONEAPI_IMG     ?= ${SERVICE_REGISTRY}/${SERVICE_ORG}/openstack-keystone:current-tripleo
```
cescgina, line 43: Maybe instead of removing them, the default value should be `unused`? That seems to be a null value in the `gen-service-kustomize.sh`. Also, do all operators have the image set in the sample files? If not, this could cause problems

bogdando: kustomize should use sample files as resources for custom images data.
we should stop overloading env vars use in makefiles

kajinamit:
> 
I was updating the default service images used by operators but install_yaml "reverts" the change in sample files and deploys CR with old images. This is causing the failure in https://github.com/openstack-k8s-operators/keystone-operator/pull/215 .

We can workaround the problem by merging the change in install_yamls first but in general defining defaults in mutliple layers causes tricky dependencies when updating defaults so I'm trying to fix it.

kajinamit:
Probably https://github.com/openstack-k8s-operators/install_yamls/pull/148 (just removing the hard-coded defaults) would be a better short-term solution for now. I'm still not sure how we can implement support for customizing multiple container images using current implementation but that can be revisited later.

PR #146, opened 2023-03-14
Switch from tripleo-repos to repo-setup

cjeanner:
repo-setup is the way to go from now on.

### DIFF
--- a/devsetup/edpm/edpm-play.yaml
+++ b/devsetup/edpm/edpm-play.yaml
@@ -15,11 +15,11 @@ spec:
           shell: |
             rpm -q git || sudo yum -y install git
             sudo yum -y install python-setuptools python-requests python3-pip
-            git clone https://git.openstack.org/openstack/tripleo-repos
-            pushd tripleo-repos
+            git clone https://github.com/openstack-k8s-operators/repo-setup
+            pushd repo-setup
             sudo python3 setup.py install
             popd
-            sudo /usr/local/bin/tripleo-repos current-tripleo-dev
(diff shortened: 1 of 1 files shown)


devsetup/edpm/edpm-play.yaml
```diff
@@ -15,11 +15,11 @@ spec:
...
             rpm -q git || sudo yum -y install git
             sudo yum -y install python-setuptools python-requests python3-pip
-            git clone https://git.openstack.org/openstack/tripleo-repos
-            pushd tripleo-repos
+            git clone https://github.com/openstack-k8s-operators/repo-setup
+            pushd repo-setup
             sudo python3 setup.py install
             popd
-            sudo /usr/local/bin/tripleo-repos current-tripleo-dev
+            sudo /usr/local/bin/repo-setup current-tripleo-dev
```
raukadah, line 22: we can replace current-tripleo-dev with current-podified-dev

raukadah (changes requested):

devsetup/edpm/edpm-play.yaml
steveb: Agreed, that is the only required change I see here

raukadah (approved):

kajinamit:
Ugh. we should have removed the reference to this PR from https://github.com/openstack-k8s-operators/install_yamls/pull/152 .

I think we no longer need this because we put the repo setup to a separate make command ?
https://github.com/openstack-k8s-operators/install_yamls/pull/137

We have to update the CI script to run that command before edpm_play.

PR #148, opened 2023-03-15
Stop overriding image by default

kajinamit:
Currently install_yaml defines its own defaults about the container images used to deploy services (eg. placement-api) but this causes the redundant requirement that we have to maintain default images in two layers (install_yamls and each operators).

This removes the hard-coded defaults in install_yamls and let is use the default images defined in individual operator repos so that we can maintain default images only in operator repos.

This also adds TODO comments for operators which do not yet have proper interfaces for image customization, so that we can revisit these later easily.

Finally, some lines were using tabs while the others are not. This fixes that format inconsistency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,14 +38,14 @@ MEMCACHED_CR        ?= ${OPERATOR_BASE_DIR}/infra-operator/${MEMCACHED}
 MEMCACHED_DEPL_IMG  ?= unused
 
 # Keystone
-KEYSTONE_IMG        ?= quay.io/openstack-k8s-operators/keystone-operator-index:latest
-KEYSTONE_REPO       ?= https://github.com/openstack-k8s-operators/keystone-operator.git
-KEYSTONE_BRANCH     ?= master
-KEYSTONEAPI         ?= config/samples/keystone_v1beta1_keystoneapi.yaml
-KEYSTONEAPI_CR      ?= ${OPERATOR_BASE_DIR}/keystone-operator/${KEYSTONEAPI}
-KEYSTONEAPI_IMG     ?= unused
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -31,22 +27,23 @@ OPENSTACK_CR         ?= ${OPERATOR_BASE_DIR}/openstack-operator/${OPENSTACK_CTLP
...
+INFRA_IMG           ?= quay.io/openstack-k8s-operators/infra-operator-index:latest
+INFRA_REPO          ?= https://github.com/openstack-k8s-operators/infra-operator.git
+INFRA_BRANCH        ?= master
 
 # Memcached
+# MEMCACHED_IMG     ?= (tis is unused because this is part of infra operator)
 MEMCACHED           ?= config/samples/memcached_v1beta1_memcached.yaml
 MEMCACHED_CR        ?= ${OPERATOR_BASE_DIR}/infra-operator/${MEMCACHED}
-MEMCACHED_IMG       ?= ${SERVICE_REGISTRY}/${SERVICE_ORG}/openstack-memcached:current-tripleo
+MEMCACHED_DEPL_IMG  ?= unused
```
stuggi, line 38: wondering if it is confusing if we name the service image in some services `<>_DEPL_IMG` and in others `_IMG`, e.g. `KEYSTONEAPI_IMG` even we have there the `API` part. maybe we should have them all the same extension, either `_DEPL_IMG` or `_SVC_IMG`

kajinamit: Yeah I agree with it. Initially I was thinking of renaming parameters for operator images but that might cause more impact now as these are intensively used by CI. So I've updated the ones used for services.

Note that I selected _DEPL_IMG because that is already used and also the "service" term can be confusing because of the same term in k8s.

kajinamit:
recheck

stuggi (approved):
/lgtm

PR #149, opened 2023-03-15
We don't need `sudo` for virsh

cjeanner:
CRC isn't using `sudo` to run VMs, they are using the "system" connection. We can do the same, allowing us to NOT rely on sudo for virsh commands.


abays (approved):
/lgtm

dprince (approved):

PR #154, opened 2023-03-20
Allow to configure a bit more parameters for Compute VM

cjeanner:
This change allows to set CPU, RAM and disk size values for a compute.

In order to get a better integration with ci-framework, some more parameters are now available, in order to tweak the output directories for various artifacts loaded or created by the script.

It also makes a small change as to how the VM disk is created, using image layer (disk image chain). It allows to get rid of some qemu steps. In order to still be able to get the wanted disk size at the end, we now inject a couple of "firstboot-command" in order to resize the disk live.

The whole will allow to run multiple compute nodes, leveraging the EDPM_COMPUTE_SUFFIX parameter, and an integration with the ci-framework by allowing to output files in the common location defined therein.

Also, we ensure we're cleaning things from the correct location, by allowing to pass the same kind of parameters to the edpm-compute-cleanup.sh script.


rebtoor (approved):
+2 for me

cjeanner:
Guess I can merge it? Or shall I wait for anyone else to do it? @dprince and concerns? :)

dprince (approved):

PR #158, opened 2023-03-20
Support RHEL8 Hypervisors without --firstboot-command

fultonj:
A recent patch [1] added a call to `virt-customize --firstboot-command` but that option isn't supported in the `virt-customize` which ships with RHEL8.

This patch switches from --firstboot-command to --firstboot  which is supported on the RHEL8 version of virt-customize (1.44 as of now). This has the same effect but let's us continue to support RHEL8.

Also, the older version of virt-copy-out is not using the exported VIRSH_DEFAULT_CONNECT_URI so pass it with a `-c`.

[1] 0cb376b52e3f0e299fb40dd1f5c0eb8ff51cb492

### DIFF
--- a/devsetup/scripts/gen-edpm-compute-node.sh
+++ b/devsetup/scripts/gen-edpm-compute-node.sh
@@ -156,6 +156,11 @@ cat <<EOF >${OUTPUT_BASEDIR}/${EDPM_COMPUTE_NAME}.xml
 </domain>
 EOF
 
+cat <<EOF >${OUTPUT_BASEDIR}/${EDPM_COMPUTE_NAME}-firstboot.sh
+growpart /dev/vda 1
+xfs_growfs /
+EOF
+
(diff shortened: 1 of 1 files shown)


devsetup/scripts/gen-edpm-compute-node.sh
```diff
@@ -175,8 +180,7 @@ if [ ! -f ${DISK_FILEPATH} ]; then
     virt-customize -a ${DISK_FILEPATH} \
         --root-password password:12345678 \
         --hostname ${EDPM_COMPUTE_NAME} \
-        --firstboot-command "growpart /dev/sda 1" |
```
gibizer, line 178: as a sideeffect you are fixing this bug having a `|` instead of `\` here.

gibizer (approved):

gibizer:
I tested `make edpm_compute` locally and it works for me.

PR #159, opened 2023-03-20
Revert "Change tripleo-repos to repo-setup"

fultonj:
This reverts commit ab71b5654567fe8bf76a3b92daf852f703e19d23.

Causes the following failure:
```
tripleo-repos: error: argument REPO: invalid choice: 'current-podified-dev' (choose from 'current', 'deps', 'current-tripleo', 'current-tripleo-dev', 'ceph', 'opstools', 'tripleo-ci-testing', 'current-tripleo-rdo')
```
I don't see that option in tripleo-respos:

https://github.com/openstack/tripleo-repos/search?q=current-podified-dev

If tripleo-repos was supposed to be replaced with something called repo-setup which supports the current-podified-dev option, then that was missing from the commit.

### DIFF
--- a/devsetup/scripts/edpm-compute-repos.sh
+++ b/devsetup/scripts/edpm-compute-repos.sh
@@ -35,7 +35,7 @@ git clone https://git.openstack.org/openstack/tripleo-repos
 pushd tripleo-repos
 sudo python3 setup.py install
 popd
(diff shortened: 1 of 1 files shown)


gibizer:
#160 is merged and fixed the issue.

PR #163, opened 2023-03-22
Use LIBVIRT_DEFAULT_URI instead of VIRSH_DEFAULT_CONNECT_URI

stuggi:
VIRSH_DEFAULT_CONNECT_URI is deprecated and LIBVIRT_DEFAULT_URI should be used. [1]

[1] https://libvirt.org/uri.html#specifying-uris-to-virsh-virt-manager-and-virt-install

### DIFF
--- a/devsetup/scripts/gen-edpm-compute-node.sh
+++ b/devsetup/scripts/gen-edpm-compute-node.sh
@@ -14,7 +14,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 set -ex
-export VIRSH_DEFAULT_CONNECT_URI=qemu:///system
+export LIBVIRT_DEFAULT_URI=qemu:///system
 # expect that the common.sh is in the same dir as the calling script
 SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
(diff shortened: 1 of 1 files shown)


stuggi:
on my host the virt-copy-out fails not finding the vm as it was not using the VIRSH_DEFAULT_CONNECTION_URI. it works when setting LIBVIRT_DEFAULT_URI.

cjeanner:
/lgtm

Meh... of course, we'd need to use multiple env vars, why only one that will be known by everyone? :)

That said, `virt-copy-out` is weird by itself, and may be replaced by some other command iirc. Anyway.

fao89 (approved):

stuggi:
> I'm not sure this was needed as it was fixed already in #158 by using -c, instead of relying on implicit env vars. If you were testing PR156 and saw this issue, you probably hadn't rebased it on master.
> 
> In fact, does this fix still work? Because now it looks like it will try and use -c with an unset env var:
> 
> https://github.com/openstack-k8s-operators/install_yamls/blob/master/devsetup/scripts/gen-edpm-compute-node.sh#L198

you are right, I missed to rebase. follow up PR in https://github.com/openstack-k8s-operators/install_yamls/pull/164

PR #167, opened 2023-03-23
Remove weird ssh pub key addition on root

cjeanner:
The removed line was basically allowing a virtual machine to access its
hypervisor root account. It really doesn't sound right, and people should
consider cleaning up their /root/.ssh/authorized_keys asap.

If there's a real reason for such an access, it must be properly
documented, and if possible avoided.


fultonj:
+2

Slagle said "that was needed when we had sshuttle running for edpm->crc traffic, but we can remove it now"

fultonj (approved):

PR #172, opened 2023-03-23
Fix error with EDPM network_config

fao89:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/116#discussion_r1146623998

### DIFF
--- a/scripts/gen-edpm-kustomize.sh
+++ b/scripts/gen-edpm-kustomize.sh
@@ -114,7 +114,6 @@ patches:
         - InternalApi
         - Storage
         - Tenant
-        - External
         networks_lower:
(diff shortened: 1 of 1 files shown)


gibizer:
Yepp we need this as otherwise ansbile fails on missing external_ip variable

fultonj (approved):
We need this to fix CI. Maybe we can add it back later if the edpm_ansible role can be configured to not fail 

 https://logserver.rdoproject.org/47/47947/25/check/centos-9-crc-singlenode-podified-edpm-deployment/043c262/controller/controller/pod/dataplane-deployment-configure-networkggwzf-gdqns-logs.txt

PR #177, opened 2023-03-24
Generalize the cleanup strategy

fmount:
When `openstack_cleanup` is run, it might happen that we have more than one `CSV`/`Subscription`/`Catalogsource` in the same namespace. This is the case of `openstack-operator` and `manila-operator`: even though `Manila` is deployed by meta operator, the `OLM` dependency model installs separated resources. This patch introduces a generalized way to cleanup the `Namepace`, where for each operator that has a dedicated `Subscription`, the associated resources are deleted.

### DIFF
--- a/scripts/operator-cleanup.sh
+++ b/scripts/operator-cleanup.sh
@@ -15,6 +15,21 @@
 # under the License.
 set -x
 
+function rm_operator {
+
+    local NAMESPACE="$1"
+    local OPERATOR="$2"
+
+    local CSV=$(oc get csv --no-headers -o custom-columns=":metadata.name" --ignore-not-found=true | grep $OPERATOR)
+
+    if [ -n "$CSV" ]; then
+      oc delete -n ${NAMESPACE} csv ${CSV} --ignore-not-found=true
+    fi
+    oc delete -n ${NAMESPACE} subscription ${OPERATOR_NAME}-operator --ignore-not-found=true
+    oc delete -n ${NAMESPACE} catalogsource ${OPERATOR_NAME}-operator-index --ignore-not-found=true
+}
+
+
 if [ -z "$NAMESPACE" ]; then
     echo "Please set NAMESPACE"; exit 1
 fi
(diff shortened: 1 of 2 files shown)


scripts/operator-cleanup.sh
```diff
@@ -15,18 +15,34 @@
...
 
-if [ -n "$CSV" ]; then
-    oc delete -n ${NAMESPACE} csv ${CSV} --ignore-not-found=true
-fi
-oc delete -n ${NAMESPACE} subscription ${OPERATOR_NAME}-operator --ignore-not-found=true
-oc delete -n ${NAMESPACE} catalogsource ${OPERATOR_NAME}-operator-index --ignore-not-found=true
+# For each operator delete the associated resources (csv/subs/catalogsource)
+for operator in $OPERATORS; do
+    rm_operator $NAMESPACE $operator
+done
```
abays, line 48: This seems to remove the the CSV and `Subscriptions` of all operators I have deployed in the namespace, regardless of which `make <operator>_cleanup` I run.  i.e. I had both `MariaDB` and `Keystone` (via `make mariadb && make keystone`), but then running `make keystone_cleanup` seemed to also remove `MariaDB`.

fmount: ooh right, because we're using the same script and filter by <operator>.
I'll do another iteration and remove all the operators only if `make openstack_cleanup` is called.

fmount:
> @fmount Also, need to consider operators like nmstate, etc which might get installed in the local namespace. But perhaps we shouldn't clean them up. And also, soon the operators will all be in 'openstack-operators' not the service namespace we use today for both Openstack services and operators.

Yeah, I understand this, and perhaps is easier hardcoding what we want to remove (`openstack-{core,storage,extra}`) instead of trying to resolve the operators passed as parameter, delete most but exclude a subset.
In my understanding `openstack_cleanup` is supposed to delete all the resources of the namespace (because you might want to start from scratch), that's the reason of the `for loop`.

> 
> I think a patch like this is fine, but perhaps the biggest problem to consider is how to make the openstack-operator-bundle smaller itself. Is the intent here to propose that each operator have its own bundle or something and that we would benefit from a generic cleanup like this?
> 
Yeah, the purpose here is to be backward compatible (I still want to `make glance`, `make glance_deploy` for development purposes), but at the same time, given we're going to have `openstack-operator` and many dependencies (in theory just 2: `storage` and `extra`), I just want to delete the whole ctlplane by running `make openstack_cleanup`, which is the reason why if `OPERATOR_NAME` is `openstack`, I delete everything.

> If we go with functional groupings 'openstack-storage' for example. A hard coded cleanup like I demonstrated with manila is probably cleaner and less code to maintain. And also, less side effects. Just need to make this decision.

I agree on this, 4 lines is less error prone than what I'm doing here, hence feel free to close if you think that path is easiest for the group to mantain.

scripts/operator-cleanup.sh
fmount: Done

fmount:
@dprince @abays closing as my understanding is that there's no interest in a general cleanup approach.

PR #178, opened 2023-03-25
Use crc ip command to get CRC_DEFAULT_NETWORK_IP

raukadah:
During Network isolation configuration, default NNCP_INTERFACE interface enp6s0 needs to be created in the crc vm.

The crc ip is sometimes different in few systems. For example.
```
devsetup]$ crc ip
192.168.130.11
```
this will not create the proper network interface leading to failure of nncp operator installation.

This patch advises using crc ip to get the CRC_DEFAULT_NETWORK_IP to avoid such issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,9 @@ oc login -u kubeadmin -p 12345678 https://api.crc.testing:6443
 ```
 
 * attach libvirt default network to the crc (default IP 192.168.122.10). This network is used as a vlan trunk to isolate the networks using vlans.
+Try running `crc ip` command to get the default crc network ip.
 ```bash
-make crc_attach_default_interface
+CRC_DEFAULT_NETWORK_IP=$(crc ip) make crc_attach_default_interface
 ```
 
 * create edpm node


fao89 (approved):

README.md
```diff
@@ -60,8 +60,9 @@ oc login -u kubeadmin -p 12345678 https://api.crc.testing:6443
 ```
 
 * attach libvirt default network to the crc (default IP 192.168.122.10). This network is used as a vlan trunk to isolate the networks using vlans.
+Try running `crc ip` command to get the default crc network ip.
 ```bash
-make crc_attach_default_interface
+CRC_DEFAULT_NETWORK_IP=$(crc ip) make crc_attach_default_interface
```
slagle, line 65: I don't understand this patch. $(crc ip) is going to return the IP for the crc vm on the crc network (192.168.130.xx). The only place $CRC_DEFAULT_NETWORK_IP is used is to add add an dhcp reservation on the default network (192.168.122.xx).

slagle (changes requested):

README.md
abays: I agree.  It seems like the wrong IP address is being used here.

raukadah: Thank you @slagle and @abays for reviewing it. I misunderstood this step earlier  now it is clear.
crc vm get's default ip 192.168.130.11, and with crc_attach_default_interface we want to also add 192.168.122.10 ip to crc vm so other libvirt vms can access it. I will close this pr.

PR #181, opened 2023-03-27
use fqcn calling playbooks in edpm-play

rebtoor:
Since edpm-ansible is now a collection, I'll try to launch playbooks/roles in a compliant way, without using symlinks as workaround

playbooks have been renamed following Ansible [guidelines](https://docs.ansible.com/ansible/latest/collections_guide/collections_using_playbooks.html):

> Playbook names, like other collection resources, have a restricted set of valid characters. Names can contain only lowercase alphanumeric characters, plus _ and must start with an alpha character. The dash - character is not valid for playbook names in collections. Playbooks whose names contain invalid characters are not addressable: this is a limitation of the Python importer that is used to load collection resources.

### DIFF
--- a/devsetup/edpm/edpm-play.yaml
+++ b/devsetup/edpm/edpm-play.yaml
@@ -2,7 +2,7 @@
 apiVersion: ansibleee.openstack.org/v1alpha1
 kind: OpenStackAnsibleEE
 metadata:
-  name: osp.edpm.deploy-external-dataplane-compute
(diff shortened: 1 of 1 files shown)


slagle (commented):
lgtm

raukadah (approved):
Tested here: https://review.rdoproject.org/zuul/buildset/b9ad50cf6c9146559460cdbc86d840e8

PR #183, opened 2023-03-28
Add BMaaS to CRC devsetup

hjensas:
Add make targets in devsetup Makefile to configure libvirt networking, vm instances, CRC networking and a virtual RedFish emulator (sushy-emulator).

Access RedFish via:
  http://sushy-emulator.apps-crc.testing

Jira: OSP-23574

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -9,6 +9,21 @@ EDPM_NETWORK_CONFIG_TEMPLATE ?= templates/single_nic_vlans/single_nic_vlans.j2
 EDPM_SSHD_ALLOWED_RANGES ?= ['192.168.122.0/24']
 EDPM_CHRONY_NTP_SERVER ?= pool.ntp.org
 EDPM_PLAY_CRD ?= edpm/edpm-play.yaml
+BMAAS_INSTANCE_NAME_PREFIX ?= crc-bmaas
+BMAAS_INSTANCE_MEMORY ?= 4096
+BMAAS_INSTANCE_VCPUS ?= 2
+BMAAS_INSTANCE_DISK_SIZE ?= 10
+BMAAS_INSTANCE_OS_VARIANT ?= centos-stream9
+BMAAS_INSTANCE_VIRT_TYPE ?= kvm
+BMAAS_INSTANCE_NET_MODEL ?= virtio
+BMAAS_NETWORK_NAME ?= crc-bmaas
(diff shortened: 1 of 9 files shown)


hjensas:
recheck

devsetup/scripts/bmaas/network-setup.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+# set -x
+
```
stuggi, line 3: how about we make this a generic network-setup.sh script for devsetup. right now we use the default libvirt network for attaching the edpm node and do a static dhcp reservations for crc and the edpm vms. this sometimes result in issues where when you re-create the VM and the new VM with a new mac gets a different IP address even if there is a static reservation. seems that dnsmasq still has a lease for the old one. the proper way could be like in this script to destroy and recreate the network and recreate it when the env get re-created, which we probably don't want to do with the default libvirt network.

hjensas: Sure, I'll move this script out of the bmaas subdirectory and do some slight modifications to make it more generic.

hjensas: @stuggi

I removed the "crc-bmaas" mentions in the script, moved it up to the /scripts directory.
Changed the default network name to crc-net-iso, and used 172.16.0.1/24 as the IP address.
I think that's all that is needed for generic use?

stuggi: yes, thanks! lgtm

hjensas:
recheck

devsetup/scripts/bmaas/br-baremetal-bridge.sh
```diff
@@ -0,0 +1,119 @@
...
+    temp_file=$(mktemp -p "$MY_TMP_DIR")
+    cat << EOF > "$temp_file"
+apiVersion: nmstate.io/v1
+kind: NodeNetworkConfigurationPolicy
+metadata:
+  name: br-baremetal
+spec:
+  desiredState:
+    interfaces:
+    - name: br-baremetal
```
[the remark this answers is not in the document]
hjensas: I've updated the change to use $BMAAS_NETWORK_NAME for this.

Sandeepyadav93:
+1 I tested this patch last week to create bmaas lab on libvirt end and it worked fine, I managed to successfully provision bmaas nodes.

rabi (approved):
/lgtm

PR #185, opened 2023-03-29
Use make wait target to know the operator installation status

raukadah:
https://github.com/openstack-k8s-operators/install_yamls/pull/174 adds the make wait utility by which to wait for an operator's controller-manager pod to become fully available.

We can use the same to know the successful installation of operators.

It also adds irrelevant_files to zuul job to avoid running on docs or markdown changes.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/174


raukadah:
No longer needed.

PR #187, opened 2023-03-29
Adding more options to kustomize EDPM

fao89:


stuggi (commented):
just a note, the network details are kind of hard coded in couple of places at the moment. if the cidr is changed, it might also required a change to https://github.com/openstack-k8s-operators/install_yamls/blob/master/scripts/gen-netatt.sh . we can probably get rid of the hard ip stuff in here when ipam got created

PR #190, opened 2023-03-30
Don't configure IPs on worker nodes isolated networks IF

stuggi:
There is no need to configure IPs on the worker nodes interfaces connected to the isolated networkes as MetalLB uses the hostIP for the speakers. This configuration allows to have a single nncp for all worker nodes should be used for the osctlplane.

Also via the L2Advertisement resources the interface used to announce the IP can be limited.

Depends-On: https://github.com/openstack-k8s-operators/nova-operator/pull/314


stuggi:
moving to draft as https://github.com/openstack-k8s-operators/nova-operator/pull/314 is required

PR #191, opened 2023-03-30
[CI] Use crc-csi-hostpath-provisioner storage class

raukadah:

PR #192, opened 2023-03-31
Assign static IPs to edpm-computes based on suffix

fultonj:
Remove calls to 'virsh net-update' for DHCP reservations and instead use a 'virt-customize --firstboot' script to statically assign an IP based the EDPM_COMPUTE_SUFFIX. The DCHP reservations have been unreliable.

This patch has been tested against 'make edpm_deploy' to confirm os-net-config still functions correctly. It has also been tested to confirm IP is retained after reboot.

A side effect is that 'virsh -q domifaddr edpm-compute-0' does not return an address. However, we already depend on knowing the EDPM_COMPUTE_SUFFIX. References to 'virsh -q domifaddr' have been replaced accordingly. The $COMPUTE_IP was it with 'virsh -q domifaddr' and is needed for 'make edpm_compute' but is not needed for 'make edpm_deploy'.

### DIFF
--- a/devsetup/README.md
+++ b/devsetup/README.md
@@ -103,13 +103,6 @@ Enable repositories:
 make edpm_compute_repos
 ```
 
-Discover IP of the compute node VM:
-```
-virsh --connect=qemu:///system -q domifaddr edpm-compute-0 | awk 'NF>1{print $NF}' | cut -d/ -f1
-# wait until ip address appears, then assign to a variable
-COMPUTE_IP=$(virsh --connect=qemu:///system -q domifaddr edpm-compute-0 | awk 'NF>1{print $NF}' | cut -d/ -f1 )
-```
-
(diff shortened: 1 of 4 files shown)


fultonj:
This failure does not look related since it failed before an EDPM VM was even created.

https://review.rdoproject.org/zuul/build/cfef7a4150e44219b8e1878492e9b37b

slagle (approved):

raukadah:
> This failure does not look related since it failed before an EDPM VM was even created.
> 
> https://review.rdoproject.org/zuul/build/cfef7a4150e44219b8e1878492e9b37b
https://github.com/openstack-k8s-operators/install_yamls/pull/191 will fix it.

raukadah:
recheck

raukadah:
recheck

fao89 (approved):

PR #193, opened 2023-03-31
Change edpm_compute default console to PTY

fultonj:
The VM made by `make edpm_compute` will use a serial PTY console. Users can now SSH into their hypervisor and run `virsh console edpm-compute-0` to attach to a console.

### DIFF
--- a/devsetup/scripts/gen-edpm-compute-node.sh
+++ b/devsetup/scripts/gen-edpm-compute-node.sh
@@ -129,12 +129,10 @@ cat <<EOF >${OUTPUT_BASEDIR}/${EDPM_COMPUTE_NAME}.xml
       <model type='virtio'/>
       <address type='pci' domain='0x0000' bus='0x02' slot='0x00' function='0x0'/>
     </interface>
-    <serial type='stdio'>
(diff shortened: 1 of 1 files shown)


fao89 (approved):

slagle:
why wouldn't just make the vm use pty as the default? why does this even need to be configurable?

fultonj:
> why wouldn't just make the vm use pty as the default? why does this even need to be configurable?

Good question. I added it as an option since I assumed we needed the virtio behavior. I would be happy to make it the default however (simpler).

@raukadah do you see any reason not to change the default?

cjeanner:
I don't see what may prevent defaulting on an actual PTY - it also works fine in nested VM if needed.

fultonj:
> why wouldn't just make the vm use pty as the default? why does this even need to be configurable?

done.

raukadah (approved):

raukadah:
@slagle it looks good based on CI results. https://review.rdoproject.org/zuul/build/a68b800e215b424b983edb39a24268e1/console#5/0/1/controller . We can go ahead and merge it.

PR #195, opened 2023-04-03
The compute public key isn't needed anymore

cjeanner:
Following some more of the ssh cleanup started in #167, we can also
clean the copy-out of the ssh public key from the compute.


dprince:
Looks good to me.

fultonj:
I'm OK with this since it's removing the public keys of the compute nodes which were only used by `sshuttle` and not the public keys we use when `make edpm_compute_repos` is run. I.e. you can still run `-i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa` after this patch merges. It deletes the `edpm-compute-$I` keys below.

```
[fultonj@hamfast ~]$ cd ~/install_yamls/out/edpm/
[fultonj@hamfast edpm]$ ls *.pub | sort
ansibleee-ssh-key-id_rsa.pub
edpm-compute-0-id_rsa.pub
edpm-compute-1-id_rsa.pub
[fultonj@hamfast edpm]$ 
```

slagle (approved):

PR #196, opened 2023-04-03
Ignore 'not-found' storage bundle during cleanup

abays:


dprince:
@abays nice catch. This is what I meant to add. LGTM

PR #201, opened 2023-04-04
Allow to pass some env vars to edpm-compute-repos.sh

cjeanner:
Add support for two new environment variables:
- SSH_KEY
- CMDS_FILE

Since we can override the path to the generated SSH key from within the
gen-ansibleee-ssh-key.sh and gen-edpm-compute-node.sh scripts, we should
also allow to pass that option to the edpm-compute-repos.sh script.

Also, in order to keep some log/traces of the executed commands, this
patch also allows to pass a parameter for the generated commands file.


fao89 (approved):

raukadah (approved):

fultonj (approved):
/LGTM

PR #202, opened 2023-04-04
Bump nncp timeout to 240s

gibizer:
Even in my beefy desktop 120s sometimes not enough so this patch proposes a bumped timeout.


raukadah (approved):
I hit this issue yesterday. Thanks for increasing the timeout.

PR #203, opened 2023-04-05
Install repo-setup without installing git or pip

steveb:
This switches the repo-setup install to download a github tarball and using the pip bundled with python to install it in a venv.

This means no packages need to be installed before the repos are set up.

The script is run with the root user, so sudo calls have been removed also.

### DIFF
--- a/devsetup/scripts/edpm-compute-repos.sh
+++ b/devsetup/scripts/edpm-compute-repos.sh
@@ -32,14 +32,17 @@ if [[ ! -f $SSH_KEY ]]; then
 fi
 
 cat <<EOF > $CMDS_FILE
-rpm -q git || sudo yum -y install git
-sudo yum -y install python-setuptools python-requests python3-pip
-git clone https://github.com/openstack-k8s-operators/repo-setup
-pushd repo-setup
-sudo pip install -r requirements.txt
-sudo python3 setup.py install
+#!/bin/bash
+set -ex
+# Install latest version of repo-setup without installing pip or git
+pushd /tmp
+curl -sL https://github.com/openstack-k8s-operators/repo-setup/archive/refs/heads/main.tar.gz | tar -xz
+pushd repo-setup-main
+python3 -m venv ./venv
+PBR_VERSION=0.0.0 ./venv/bin/pip install ./
+./venv/bin/repo-setup ${REPO_SETUP_CMD}
 popd
-sudo /usr/local/bin/repo-setup ${REPO_SETUP_CMD}
+popd
 EOF
 
 scp $SSH_OPT $CMDS_FILE root@$IP:/tmp/repo-setup.sh


devsetup/scripts/edpm-compute-repos.sh
```diff
@@ -30,14 +30,16 @@ if [[ ! -f $SSH_KEY ]]; then
...
 cat <<EOF > $CMDS_FILE
-rpm -q git || sudo yum -y install git
-sudo yum -y install python-setuptools python-requests python3-pip
-git clone https://github.com/openstack-k8s-operators/repo-setup
-pushd repo-setup
-sudo pip install -r requirements.txt
-sudo python3 setup.py install
+# Install latest version of repo-setup without installing pip or git
+pushd /tmp
+curl -sL https://github.com/openstack-k8s-operators/repo-setup/archive/refs/heads/main.tar.gz | tar xvz
```
cjeanner, line 35: My security hat is tingling. Would there be a way to get some checksum or anything just to ensure the archive is valid before un-taring it?

devsetup/scripts/edpm-compute-repos.sh
```diff
@@ -30,14 +30,16 @@ if [[ ! -f $SSH_KEY ]]; then
...
+pushd repo-setup-main
+python3 -m venv ./venv
+source ./venv/bin/activate
```
cjeanner, line 38: iirc we can point directly to the pip located in venv/bin/pip, it wouldn't require to source the activate bit.

devsetup/scripts/edpm-compute-repos.sh
```diff
@@ -30,14 +30,16 @@ if [[ ! -f $SSH_KEY ]]; then
...
+PBR_VERSION=0.0.0 python3 -m pip install ./
+repo-setup current-podified-dev
```
cjeanner, line 40: so here we'd be able to to directly point to ./venv/bin/repo-setup

cjeanner (changes requested):
I love the idea, but I'd love it even better with the 2-3 comments addressed :)

fultonj (commented):
I'm fine with this. +2

devsetup/scripts/edpm-compute-repos.sh
fultonj: If we're going to go that route, is there a way we can easily get the checksum in order to compare? We shouldn't hard code it in the script as it will change very often.

cjeanner: I think there is a known URI, but I can't figure it out yet. But yeah, we won't hard-code the checksum, that's certain.

steveb: I can't see how to fetch a checksum for an archive from github. I'm not sure it would give any MITM protection anyway since it would also be from github.com. If repo-setup releases were tagged we could switch to tag downloads and hard coded checksums. Given this is for dev setups I'm not sure its worth too much effort.

What I'd *really* like is the latest repo-setup packaged as a simple rpm, and published to a short stable URL

cjeanner: I've seen many mention of such checksum, but no URI :/. weird.
It was more in the case of some network issue where the tar gets corrupted - indeed if some rogue MitM happens, over TLS, fetching the checksum from the same location would be meaningless.

I don't think there's any plan on providing RPM for that package, but we may reconsider this in the future - maybe something to bring to the team.

Resolving this conversation.

cjeanner (commented):
I'm OK with that - the two last comments are more cosmetics than anything.

cjeanner:
recheck

steveb:
@cjeanner I've refreshed the change to include your changes, thanks

cjeanner:
recheck

steveb:
recheck

bshephar (approved):

Sandeepyadav93:
/lgtm
/approve

tested locally it works as expected.

PR #205, opened 2023-04-07
Re-run gen-edpm-compute-node.sh without error

slagle:
In order to make gen-edpm-compute-node.sh re-runnable without erroring,
add checks around if the VM is already defined, and already running so
that those commands don't fail.


slagle:
recheck

raukadah (approved):
/lgtm

PR #208, opened 2023-04-11
Add target to install openstack storage crds

cescgina:
After the split of openstack storage CRDs, the kuttl tests for the
storage operators (cinder and glance) need to add the storage CRDs
explicitely.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,12 +23,13 @@ STORAGE_CLASS       ?= "local-storage"
 NETWORK_ISOLATION   ?= true
 
 # OpenStack Operator
-OPENSTACK_IMG        ?= quay.io/openstack-k8s-operators/openstack-operator-index:latest
-OPENSTACK_REPO       ?= https://github.com/openstack-k8s-operators/openstack-operator.git
-OPENSTACK_BRANCH     ?= master
-OPENSTACK_CTLPLANE   ?= $(if $(findstring true,$(NETWORK_ISOLATION)),config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml,config/samples/core_v1beta1_openstackcontrolplane.yaml)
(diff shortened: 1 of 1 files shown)


cescgina:
recheck

Makefile
```diff
@@ -394,6 +395,13 @@ openstack_crds: ## installs all openstack CRDs. Useful for infrastructure dev
 	for X in $$(file ${OUT}/openstack_crds/* | grep gzip | cut -f 1 -d ':'); do tar xvf $$X -C ${OUT}/openstack_crds/; done
 	for X in $$(grep -l CustomResourceDefinition ${OUT}/openstack_crds/manifests/*); do oc apply -f $$X; done
 
+.PHONY: openstack_storage_crds
```
dprince, line 398: Since this is basically the same logic we already have in the openstack_crds target could we just reuse that same target by setting a different ENV for OPENSTACK_BUNDLE_IMG first? Reason I say this is we plan on splitting out more bundles soon so we'll likely need to do the same thing again.

cescgina: Thanks for the suggestion, I've updated the PR, reusing the `openstack_crds` target.

Makefile
```diff
@@ -394,6 +395,11 @@ openstack_crds: ## installs all openstack CRDs. Useful for infrastructure dev
 	for X in $$(file ${OUT}/openstack_crds/* | grep gzip | cut -f 1 -d ':'); do tar xvf $$X -C ${OUT}/openstack_crds/; done
 	for X in $$(grep -l CustomResourceDefinition ${OUT}/openstack_crds/manifests/*); do oc apply -f $$X; done
```
dprince, line 396: @cescgina I like that we are sharing the function, but now wonder if the fact that we extract things into the same openstack_crds directory might eventually cause a side effect. We could perform a 'basename' on the URL provided. Or just default to 'openstack_crds' but have an separate ENV variable that could override it for the 'storage' version? What do you think? We could also do this in a follow up I think.

dprince (approved):

PR #210, opened 2023-04-12
[doc] Fix steps to deploy both compute nodes

karelyatin:
Only edpm-compute-0 was getting deployed,
this patch fixes it.


stuggi (approved):
/lgtm

PR #211, opened 2023-04-13
Set metadata_proxy_shared_secret for nova and neutron

mrKisaoLamb:
This change is required due to modifications made to the metadata secret in nova-metadata, as outlined in this pr: https://github.com/openstack-k8s-operators/nova-operator/pull/332

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,11 @@
 # general
 SHELL := /bin/bash
-NAMESPACE           ?= openstack
-PASSWORD            ?= 12345678
-SECRET              ?= osp-secret
-OUT                 ?= ${PWD}/out
-INSTALL_YAMLS       ?= ${PWD}  # used for kuttl tests
+NAMESPACE              ?= openstack
+PASSWORD               ?= 12345678
+SECRET                 ?= osp-secret
+OUT                    ?= ${PWD}/out
+INSTALL_YAMLS          ?= ${PWD}  # used for kuttl tests
+METADATA_SHARED_SECRET ?= 1234567842
 
 # are we deploying to microshift
 MICROSHIFT ?= 0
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -372,7 +371,7 @@ edpm_deploy_prep: export EDPM_NETWORK_CONFIG_TEMPLATE=${DATAPLANE_NETWORK_CONFIG
 edpm_deploy_prep: export EDPM_SSHD_ALLOWED_RANGES=${DATAPLANE_SSHD_ALLOWED_RANGES}
 edpm_deploy_prep: export EDPM_CHRONY_NTP_SERVER=${DATAPLANE_CHRONY_NTP_SERVER}
 edpm_deploy_prep: export EDPM_OVN_METADATA_AGENT_NOVA_METADATA_HOST=$(shell oc get svc nova-metadata-internal -o json |jq -r '.status.loadBalancer.ingress[0].ip')
-edpm_deploy_prep: export EDPM_OVN_METADATA_AGENT_PROXY_SHARED_SECRET=${DATAPLANE_OVN_METADATA_AGENT_PROXY_SHARED_SECRET}
+edpm_deploy_prep: export EDPM_OVN_METADATA_AGENT_PROXY_SHARED_SECRET=${SECRET}
```
gibizer, line 374: I think there is a mix-up about the meaning of the SECRET variable. I think that defines the *name* of the k8s Secret being created and passed to OpenstackControlPlane CR. See gen-input-kustomize.sh L47. It does not define an actual password string.

Nova-operator can take the a password string out from a specific field of the k8s Secret it gets so for nova-operator we only need "something" generated into "MetadataSecret" field in gen-input-kustomize.sh. The neutron part of the edpm node deployment needs the actual password string passed to it via EDPM_OVN_METADATA_AGENT_PROXY_SHARED_SECRET. So the value of EDPM_OVN_METADATA_AGENT_PROXY_SHARED_SECRET and the value of `Secret(osp-secret).MetadataSecret` needs to be the same. But it should not depend on the name of the k8s Secret being generated (which is the value of SECRET). 

What I would do.

1. rename DATAPLANE_OVN_METADATA_AGENT_PROXY_SHARED_SECRET to METADATA_SHARED_SECRET and give it a default value. It can be 42 :) or 1234568 (as the rest of our default passwords)
2. pass METADATA_SHARED_SECRET to EDPM_OVN_METADATA_AGENT_PROXY_SHARED_SECRET here as before
3. change gen-input-kustomize.sh to use METADATA_SHARED_SECRET as the value of the `MetadataSecret` field.

Later when neutron-operator implements taking the shared secret from the same k8s Secret that nova-operator then we can remove EDPM_OVN_METADATA_AGENT_PROXY_SHARED_SECRET from install yamls altogether and keep only the k8s Secret generation.

gibizer (changes requested):

Makefile
mrKisaoLamb: Done. I initially tried to reuse existing components because the changes were temporary, but I agree that temporary solutions tend to live longest in the code. So now, there is a separate value passed to the script and the same value passed to edpm

PR #214, opened 2023-04-14
Use public NTP server by default in compute nodes

pablintino:
Chrony NTP client running inside compute nodes is getting configured by default with clock.redhat.com, a NTP server that is not publicly available and, thus, not a safe default to be used for this public repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,7 +204,7 @@ DATAPLANE_COMPUTE_1_IP                           ?=192.168.122.101
 DATAPLANE_RUNNER_IMG                             ?=quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
 DATAPLANE_NETWORK_CONFIG_TEMPLATE                ?=templates/single_nic_vlans/single_nic_vlans.j2
 DATAPLANE_SSHD_ALLOWED_RANGES                    ?=['192.168.122.0/24']
-DATAPLANE_CHRONY_NTP_SERVER                      ?=clock.redhat.com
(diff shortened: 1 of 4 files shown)


stuggi (commented):
+1 from my side, lets get some others feedback before landing it

raukadah:
Thank you for the pull request. Can we also remove DATAPLANE_CHRONY_NTP_SERVER from https://github.com/openstack-k8s-operators/install_yamls/blob/master/ci/playbooks/deploy_edpm.yaml#L37 ?

stuggi:
since we change the default and deployments might fail, should we mention to optionally set the ntp server in the doc at https://github.com/openstack-k8s-operators/install_yamls/blame/master/README.md#L138 ?

pablintino:
Botch comments seems reasonable. I'll push a commit removing the env var in the playbook and adding a note in the doc at that point.

raukadah (approved):
Thank you :-)

bshephar:
Looks fine to me, but we might also need some other CI related files to account for this. Best to check with CI folks to ensure this isn't going to cause pain somewhere else down the line.

cjeanner:
IMHO it's better to get a default allowing standard upstream work. We can then override it when working downstream. Also, as a general thought, we shouldn't rely on anything "downstream specific" in a public repository.

I'd go with a "merge it" directly, but if we're seeking more feedback... We may want to wait until 2023-04-18 EOD UTC (that is tomorrow), and merge it if no one is complaining.

raukadah:
Approving it. If further suggestions comes, we can update it via followup pull request.

PR #217, opened 2023-04-18
Fix container image override for glance deployment

kajinamit:
The glances CR has three containerImage key in its spec. Currently only the top-level one is overridden and actual glance-api services are always deployed using the default image. This fixes that and ensures all container images are overridden.


kajinamit:
@fmount fyi. this is the change I needed when I deployed glance using a non-default container image.

Makefile
```diff
@@ -600,7 +600,7 @@ glance_deploy_prep: glance_deploy_cleanup ## prepares the CR to install the serv
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR} ${DEPLOY_DIR}
 	pushd ${OPERATOR_BASE_DIR} && git clone -b ${GLANCE_BRANCH} ${GLANCE_REPO} && popd
 	cp ${GLANCE_CR} ${DEPLOY_DIR}
-	bash scripts/gen-service-kustomize.sh
+	bash scripts/gen-glance-kustomize.sh
```
fmount, line 603: I'm wondering if we can find a way to generalize a single script instead of duplicating it for a 90% of its content.
An idea could be running something like:

```
bash scripts/gen-service-kustomize.sh "glance"
```

and within the script, to avoid messing up w/ a bunch of if statements, you can buld a function that renders what's needed for glance.
I'm telling this because I see a pattern here, and it would be great solving for all the services instead of creating ad-hoc scripts.
I see, on the other hand, that maintaining a script for each service is easier, so it would make sense having a "common" part that you can `source` within the script.

kajinamit: I decided to create a service-specific script because that is already done for telemetry. We can create more generic script probably which I'm trying in the latest change.

kajinamit:
recheck post failure

Makefile
```diff
@@ -594,7 +593,8 @@ glance_cleanup: ## deletes the operator, but does not cleanup the service resour
 
 .PHONY: glance_deploy_prep
 glance_deploy_prep: export KIND=Glance
-glance_deploy_prep: export IMAGE=${GLANCEAPI_DEPL_IMG}
+glance_deploy_prep: export IMAGE=${GLANCEAPI_DEPL_IMG},${GLANCEAPI_DEPL_IMG},${GLANCEAPI_DEPL_IMG}
```
fmount, line 596: So let's see if I understand this properly:
- do we need the image export here because the `ContainerImage` is not "inherited" from the top-level CR? Or the purpose here is to enforce different images for multiple glance instances? I see this pattern limiting if I grow (in the future) the number of glance-api instances (that can happen for many purposes)

kajinamit: > do we need the image export here because the ContainerImage is not "inherited" from the top-level CR? 

That is correct. We have to explicitly set containerImage spec in all layers because the one in base Glance CR is not inherited to the sub Glance(Internal|External)CR.

>  I see this pattern limiting if I grow (in the future) the number of glance-api instances (that can happen for many purposes)

If you add a new instance then you can add that to IMAGE and IMAGE_PATH. I don't fully understand what "limits" adding an instance. Do you mind elaborating your concern ?

Makefile
```diff
@@ -634,7 +634,6 @@ ovn_cleanup: ## deletes the operator, but does not cleanup the service resources
 
 .PHONY: ovn_deploy_prep
 ovn_deploy_prep: export KIND=.*
-ovn_deploy_prep: export IMAGE=unused
```
fmount, line 637: @kajinamit so if not exported here ( just picking ovn as an example), you don't "produce" any kustomization for the service, right?

kajinamit: That is correct. Previously we had to export IMAGES=usused in case we don't want to generate kustomization. This PR (actually its 2nd commit) makes this as default so you don't have to export IMAGES in any layer in case you don't need to kustomize containerImage spec

Makefile
fmount: no big concerns, sorry for the noise, I think I understand the purpose of your export here and actually there's no limit as this part is only limited to the `standalone` glance deploy.

Makefile
fmount: +1 on this

fmount (approved):

Makefile
fmount: sorry I missed where `IMAGES=unused` is defined as default.
Do we need something like `IMAGE ?=unused` somewhere?

fmount (changes requested):
waiting to confirm we have a top level `IMAGE ?= unused` definition that I missed in the code.

Makefile
kajinamit: The default is now defined here https://github.com/openstack-k8s-operators/install_yamls/pull/217/files#diff-ff72522ca932d763b97b0fadb34a28a71de16c34626f4c53810a2ffd397b1842R38

fmount: ack thanks, so obvious :D

fmount (approved):

PR #218, opened 2023-04-18
In `gen-edpm-compute-node.sh` script fix the check of release version for both RHEL and CentOS

rebtoor:
The check in the `devsetup/scripts/gen-edpm-compute-node.sh`script for the installation of the packages that provides `virt-customize` (guestfs-tools) is working only on RHEL instances since the `/etc/redhat-release` file format is different between RHEL and CentOS. The result is that the package isn't installed by the script so the `make edpm_compute` step fails.

Here's the format of both files.

RHEL:
```Red Hat Enterprise Linux release 9.1 (Plow)```

CentOS:
```CentOS Stream release 9```

Also, `/etc/centos-release` is used in the check to to highlight the difference between the two OSes (on CentOS `redhat-release` is a symlink of `centos-release`)

### DIFF
--- a/devsetup/scripts/gen-edpm-compute-node.sh
+++ b/devsetup/scripts/gen-edpm-compute-node.sh
@@ -186,10 +186,12 @@ if [ ! -f ${DISK_FILEPATH} ]; then
     fi
     qemu-img create -o backing_file=${CRC_POOL}/centos-9-stream-base.qcow2,backing_fmt=qcow2 -f qcow2 "${DISK_FILEPATH}" "${EDPM_COMPUTE_DISK_SIZE}G"
     if [[ ! -e /usr/bin/virt-customize ]]; then
-        if [[ $(awk '{print $6}' /etc/redhat-release) =~ ^8.* ]]; then
+        if [[ $(awk '{print $6}' /etc/redhat-release) =~ ^8.* ]] || \
+           [[ $(awk '{print $4}' /etc/centos-release) =~ ^8.* ]]; then
             sudo dnf -y install hexedit libguestfs-tools-c
         fi
-        if [[ $(awk '{print $6}' /etc/redhat-release) =~ ^9.* ]]; then
+        if [[ $(awk '{print $6}' /etc/redhat-release) =~ ^9.* ]] || \
+           [[ $(awk '{print $4}' /etc/centos-release) =~ ^9.* ]]; then
             sudo dnf -y install guestfs-tools
         fi
     fi
--- a/devsetup/scripts/gen-edpm-compute-node.sh
+++ b/devsetup/scripts/gen-edpm-compute-node.sh
@@ -186,12 +186,9 @@ if [ ! -f ${DISK_FILEPATH} ]; then
     fi
     qemu-img create -o backing_file=${CRC_POOL}/centos-9-stream-base.qcow2,backing_fmt=qcow2 -f qcow2 "${DISK_FILEPATH}" "${EDPM_COMPUTE_DISK_SIZE}G"
     if [[ ! -e /usr/bin/virt-customize ]]; then
-        if [[ $(awk '{print $6}' /etc/redhat-release) =~ ^8.* ]] || \
-           [[ $(awk '{print $4}' /etc/centos-release) =~ ^8.* ]]; then
+        if [ $(rpm -E %{rhel}) -eq 8 ]; then
             sudo dnf -y install hexedit libguestfs-tools-c
-        fi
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: devsetup/scripts/gen-edpm-compute-node.sh [hunk 1]
```diff
@@ -186,10 +186,12 @@ if [ ! -f ${DISK_FILEPATH} ]; then
     fi
     qemu-img create -o backing_file=${CRC_POOL}/centos-9-stream-base.qcow2,backing_fmt=qcow2 -f qcow2 "${DISK_FILEPATH}" "${EDPM_COMPUTE_DISK_SIZE}G"
     if [[ ! -e /usr/bin/virt-customize ]]; then
-        if [[ $(awk '{print $6}' /etc/redhat-release) =~ ^8.* ]]; then
+        if [[ $(awk '{print $6}' /etc/redhat-release) =~ ^8.* ]] || \
```
fultonj, line 189: Maybe we can use this to simplify `rpm -E %{rhel}`. 

[fultonj@hamfast zed]$ if [[ $(rpm -E %{rhel}) == 8 ]]; then echo 1; fi
1
[fultonj@hamfast zed]$ 

[fultonj@osp-storage-01 ~]$ if [[ $(rpm -E %{rhel}) == 9 ]]; then echo 1; fi
1
[fultonj@osp-storage-01 ~]$

fultonj: This command works on centos 9 too.
```
[root@edpm-compute-0 ~]# rpm -E %{rhel}
9
[root@edpm-compute-0 ~]# cat /etc/redhat-release 
CentOS Stream release 9
[root@edpm-compute-0 ~]# 
```

fultonj (changes requested):
Could you please change this to use `rpm -E %{rhel}` so it's simpler?

slagle (approved):

devsetup/scripts/gen-edpm-compute-node.sh
```diff
@@ -186,12 +186,9 @@ if [ ! -f ${DISK_FILEPATH} ]; then
     fi
     qemu-img create -o backing_file=${CRC_POOL}/centos-9-stream-base.qcow2,backing_fmt=qcow2 -f qcow2 "${DISK_FILEPATH}" "${EDPM_COMPUTE_DISK_SIZE}G"
     if [[ ! -e /usr/bin/virt-customize ]]; then
-        if [[ $(awk '{print $6}' /etc/redhat-release) =~ ^8.* ]] || \
-           [[ $(awk '{print $4}' /etc/centos-release) =~ ^8.* ]]; then
+        if [ $(rpm -E %{rhel}) -eq 8 ]; then
             sudo dnf -y install hexedit libguestfs-tools-c
```
karelyatin, line 190: i think would be better to remove rhel checks and and just do sudo dnf -y install /usr/bin/virt-customize ?
hexedit should be automatically installed with virt-customize as it's a dep or there is some issue not sure why it was originally added

rebtoor: It's something i was thinking of yesterday... I'd choose this approach too. @slagle, @fultonj ?

raukadah:
recheck

devsetup/scripts/gen-edpm-compute-node.sh
fultonj: `virt-customize` is provided by different packages on RHEL8 vs RHEL9. On my RHEL8 system I needed to add `hexedit`; the dep wasn't working to pull it in and I didn't dig into why.

https://github.com/openstack-k8s-operators/install_yamls/commit/fb31098ba71ec8244cf22e4e94c39b6fb2b170a3

If you can `dnf install /path/to/file` and have it identify the correct package relative to it's OS and confirm this works on all relevant platforms (cent-8 cent-9 rhel-8 rhel-9) then I'm OK with it. I did a lot of patches to devsetup/scripts/ because people only tested on rhel9 along the way but if you test it and it doesn't break, then that's fine with me.

fultonj: I see this example in DNF's docs so it sounds like a good approach.

`dnf install /usr/bin/rpmsign`: Install a package that provides the /usr/bin/rpmsign file.

https://dnf.readthedocs.io/en/latest/command_ref.html#install-examples

rebtoor: I'll propose a follow-up patch for that.

PR #222, opened 2023-04-19
Fix syntax error when workers > 1

olliewalsh:
Currently fails when there are multiple worker nodes:

    WORKERS=ostest-worker-0 ostest-worker-1 ostest-worker-2 \
    bash scripts/gen-nncp.sh
    /bin/bash: ostest-worker-1: command not found


abays (approved):
/lgtm

PR #223, opened 2023-04-20
Add secrets for heat-operator

kajinamit:
This adds some secret values used by heat-operator to global osp-secret.

This also introduces the secret entry for `[DEFAULT] auth_encryption_key` option. This option is used to encrypt auth data associated with stacks, and should not be rorated. So we prefer requiring a static value instead of auto-generating it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,12 @@
 # general
 SHELL := /bin/bash
-NAMESPACE              ?= openstack
-PASSWORD               ?= 12345678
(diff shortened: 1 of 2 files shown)


kajinamit:
recheck POST FAILURE

bshephar (approved):

bshephar (approved):

raukadah (approved):

PR #225, opened 2023-04-20
Add IMAGE=unused for heat_deploy_prep

jparoly:
This fixes the error where it expects an IMAGE value.

Also add devsetup/tls.crt to git ignore since local setup instructions end up creating this file which we do not want checked in

Change-Id: I798db4e00738b4bf618eac126b233bb974050288


Makefile
```diff
@@ -1161,6 +1161,7 @@ heat_cleanup: ## deletes the operator, but does not cleanup the service resource
 
 .PHONY: heat_deploy_prep
 heat_deploy_prep: export KIND=Heat
+heat_deploy_prep: export IMAGE=unused
```
fmount, line 1164: there's #217 that is supposed to address the problem for all the operators. 
Also, I think that patch set a global `IMAGE ?=unused` in the code, which 
is used by all the targets as a default, so we don't have to `export IMAGE=unused`
on each layer.
@kajinamit to comment further here.

kajinamit:
I'll rebase this on https://github.com/openstack-k8s-operators/install_yamls/pull/217 once that is merged. I'll mark this as WIP till then.

kajinamit:
Oops I left a comment in a wrong PR. Ignore that.

Makefile
kajinamit: Yeah I think this should be fixed by #217 so we can use that instead of this one.

I didn't use heat_deploy_prep directly so didn't notice this. Good catch.

fmount: Should we rebase this patch on top of #217 ?

kajinamit: No. Once #217 is merged you don't have to set IMAGE=unused  because that is the internal default.

bshephar (approved):

kajinamit (approved):

jparoly:
Can we please get this merged?

Makefile
```diff
@@ -1057,7 +1057,7 @@ heat_kuttl_run: ## runs kuttl tests for the heat operator, assumes that everythi
 	INSTALL_YAMLS=${INSTALL_YAMLS} kubectl-kuttl test --config ${HEAT_KUTTL_CONF} ${HEAT_KUTTL_DIR}
 
 .PHONY: heat_kuttl
-heat_kuttl: namespace input openstack_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy heat heat_deploy_prep heat_deploy  ## runs kuttl tests for the heat operator. Installs openstack crds and keystone operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
+heat_kuttl: namespace input openstack_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy rabbitmq rabbitmq_deploy infra heat heat_deploy_prep heat_deploy  ## runs kuttl tests for the heat operator. Installs openstack crds and keystone operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
 	make heat_kuttl_run
 	make deploy_cleanup
```
cescgina, line 1062: We should also add cleanup for infra and rabbitmq after the deploy_cleanup:
```
make rabbitmq_deploy_cleanup
make infra_cleanup
make rabbitmq_cleanup
```

abays: One this change request is completed, I think this is ready for merge

Makefile
```diff
@@ -1057,9 +1057,12 @@ heat_kuttl_run: ## runs kuttl tests for the heat operator, assumes that everythi
 	INSTALL_YAMLS=${INSTALL_YAMLS} kubectl-kuttl test --config ${HEAT_KUTTL_CONF} ${HEAT_KUTTL_DIR}
 
 .PHONY: heat_kuttl
-heat_kuttl: namespace input openstack_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy heat heat_deploy_prep heat_deploy  ## runs kuttl tests for the heat operator. Installs openstack crds and keystone operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
+heat_kuttl: namespace input openstack_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy rabbitmq rabbitmq_deploy infra heat heat_deploy_prep heat_deploy  ## runs kuttl tests for the heat operator. Installs openstack crds and keystone operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
```
cescgina, line 1060: Last thing I promise :) We should remove the `heat_deploy` as dependency, as that is done as part of the kuttl test run

jparoly: trying to push the change but there is a problem.

jparoly: finally working

cescgina (commented):
lgtm, thanks for the changes!

PR #227, opened 2023-04-24
fix tenant net ips for edpm nodes

stuggi:


fultonj (commented):
+2

As per the original network isol patch tenant network was 172.19.

https://github.com/openstack-k8s-operators/install_yamls/commit/06b8f5e3c3e381a82a3f567621d0d52f90a3bc89

PR #228, opened 2023-04-25
Add make targets infra, rabbitmq, rabbitmq_deploy to ironic-kuttle

hjensas:
When not using json-rpc we need the infra operator for the transporturls.rabbitmq.openstack.org CRD. We also need to deploy the rabbitmq cluster.


hjensas:
recheck

Failed on unrelated test "Wait for nova-edpm-compute-0-deploy-nova pod to finish"

raukadah:
recheck

stuggi (approved):
/lgtm

PR #229, opened 2023-04-26
Allow to NOT rm generated content

cjeanner:
It may happen we'd want to keep the generated content from previous run
or integrate external content, generated before hitting a specific step.

This is especially true for content related to kustomize: we may get an
external tool (like ci-framework) generating some content in the
DEPLOY_DIR; with the call to _cleanup tasks as preparation of _prep
tasks, that generated content would then be lost, leading to an unwanted
behavior.

This patch allows to override the "rm -Rf" usually called, by providing
a new parameter CLEANUP_DIR_CMD.
It's also exported as an environment variable, therefore it should allow
nested scripts to benefit of it.

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -9,6 +9,11 @@ EDPM_NETWORK_CONFIG_TEMPLATE ?= templates/single_nic_vlans/single_nic_vlans.j2
 EDPM_SSHD_ALLOWED_RANGES ?= ['192.168.122.0/24']
 EDPM_CHRONY_NTP_SERVER ?= pool.ntp.org
 EDPM_PLAY_CRD ?= edpm/edpm-play.yaml
+CLEANUP_DIR_CMD ?= rm -Rf
+
+define vars
+${1}: export CLEANUP_DIR_CMD=${CLEANUP_DIR_CMD}
+endef
 
 ##@ General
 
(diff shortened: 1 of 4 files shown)


devsetup/Makefile
```diff
@@ -9,6 +9,11 @@ EDPM_NETWORK_CONFIG_TEMPLATE ?= templates/single_nic_vlans/single_nic_vlans.j2
 EDPM_SSHD_ALLOWED_RANGES ?= ['192.168.122.0/24']
 EDPM_CHRONY_NTP_SERVER ?= pool.ntp.org
 EDPM_PLAY_CRD ?= edpm/edpm-play.yaml
+CLEANUP_DIR_CMD ?= rm -Rf
+
+define vars
+${1}: export CLEANUP_DIR_CMD=${CLEANUP_DIR_CMD}
+endef
```
raukadah, line 16: This function vars is not called anywhere. Can we drop this?

raukadah (commented):
Looks good, left one question.

devsetup/Makefile
cjeanner: oh, right. hmm. dumb me copy-pasting things. I'll amend it shortly.

cjeanner: It's now called in the needed places.

abays (approved):
/lgtm

PR #230, opened 2023-04-27
Allow git clone commands to take additional parameters

pablintino:
Till now git clones assumed that the source code to be used can be defined by the repo and branch tuple, but if the code comes from a PR we need to add the reference of the PR into the equation.
To avoid overcomplicating git clone management this change proposes to allow external customization of the git clone command, this way, install_yamls users can, for example, clone for a local repo already pointing to the desidered reference and speeding up the clone by providing the "-l" option.

Here goes an example:
``` bash
# An external tool, ie zuul, clones the repo in a non-default location
git clone git@github.com:openstack-k8s-operators/openstack-operator.git /home/zuul/

# The PR commit is checked out
git fetch origin pull/283/head

# Call openstack deploy with the PR code using the previously cloned repo as local remote
# Note: -l will bypass some internal git command logic about remote fetching, can be done as clone is from local FS
# Note: Passing and empty branch will make git clone to point the clone to the same reference/commit used in the source repository 
make openstack_deploy GIT_CLONE_OPTS="-l" OPENSTACK_BRANCH="" OPENSTACK_REPO=/home/zuul/
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,9 @@ STORAGE_CLASS       ?= "local-storage"
 # network isolation
 NETWORK_ISOLATION   ?= true
 
+# options to pass in all targets that use git clone
+GIT_CLONE_OPTS      ?=
+
 # OpenStack Operator
 OPENSTACK_IMG                ?= quay.io/openstack-k8s-operators/openstack-operator-index:latest
 OPENSTACK_REPO               ?= https://github.com/openstack-k8s-operators/openstack-operator.git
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -354,7 +357,7 @@ openstack_deploy_prep: export KIND=OpenStackControlPlane
 openstack_deploy_prep: openstack_deploy_cleanup $(if $(findstring true,$(NETWORK_ISOLATION)), nmstate nncp netattach metallb metallb_config)  ## prepares the CR to install the service based on the service sample file OPENSTACK
 	$(eval $(call vars,$@,openstack))
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR} ${DEPLOY_DIR}
-	pushd ${OPERATOR_BASE_DIR} && git clone -b ${OPENSTACK_BRANCH} ${OPENSTACK_REPO} && popd
+	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(OPENSTACK_BRANCH),-b ${OPENSTACK_BRANCH}) ${OPENSTACK_REPO} "${OPERATOR_NAME}-operator" && popd
```
raukadah, line 360: On doing git clone, the clonned directory is always going to be "${OPERATOR_NAME}-operator". Do we need to it?

raukadah (commented):
Thank you Pablo for the pr, Looks good, I have left comment. Please have a look. Thank you :-)

Makefile
pablintino: If we clone from an already cloned local repo, like the one cloned by zuul when using the ci-framework to validate a PR, the clone origin will be path, that may not end with "<operator-name>-operator". I added that at the end to ensure that the operator will be always cloned to the expected dir dispite the clone source.

raukadah: Thank you @pablintino for explaining it, make sense. I would suggest to add an example in the first conversation of this pr. It will be easier for other reviewers.

raukadah:
+2 , Thanks :-)

stuggi (approved):
/lgtm

karelyatin (approved):

karelyatin:
Example should have been:-
make openstack_deploy GIT_CLONE_OPTS="-l" OPENSTACK_BRANCH="" OPENSTACK_REPO=/home/zuul/<openstack-operator path>

PR #231, opened 2023-04-27
Remove INSTALL_YAMLS variable no longer used

cescgina:
Since the kuttl tests no longer use the install_yamls Makefile to run
the tests, this variable is no longer used and can be removed.


abays (approved):
/lgtm

PR #236, opened 2023-05-01
Add ovs_deploy_cleanup to deploy_cleanup

abays:
I noticed while running `ovs_kuttl` that the `deploy_cleanup` Make target does not include `ovs_deploy_cleanup`.  This patch adds it to the target.


gibizer (approved):
Looks good

PR #241, opened 2023-05-03
Style

jpodivin:
Style adjustments for our ansible files. This should minimize changes when at some point they are inevitably edited and trigger pre-commit. Changes were all derived from recommendations of `ansible-lint`. 

Although these are meant to only be style adjustments, with no effect on function of the affected files, it is possible that they influence actual install_yamls behavior in unforeseen ways. Please take a care to check changes in individual files.

Special attention should be given to the last patch, introducing `pipefail` behavior to an existing shell script.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,2 +1,3 @@
+---
 warn_list:
-  - syntax-check[specific] # couldn't resolve module/action "foo"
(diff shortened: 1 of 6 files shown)


stuggi:
looks good to me, @raukadah please have a look at the proposed ci playbooks changes

raukadah (approved):
Thanks @jpodivin for improving it.

PR #242, opened 2023-05-04
Switch to openstack namespace after running kuttl

cescgina:
Switch to openstack namespace after running the kuttl tests for
keystone. If the openstack namespace is not present, switch to the
default namespace.


abays:
recheck

abays:
recheck

abays:
recheck

abays (approved):
/lgtm

PR #247, opened 2023-05-05
Add oc to download_tools

afazekas:
Looks like (only) oc missing from the download_tools, adding it.


afazekas:
Forget the crc one.

PR #248, opened 2023-05-05
No fact gather download_tools

afazekas:
Fact gethering holding up the playbook
without a need.


abays:
Looks like linting wants us to `s/no/false`:

```
devsetup/download_tools.yaml:6:17: [warning] truthy value should be one of [false, true] (truthy)
```

abays (approved):
/lgtm

PR #255, opened 2023-05-10
[CI]Disable bmo_setup and decrease resources for rabbitmq

raukadah:
The Zuul Job runs crc vm with 16GB RAM and 6 CPUS.
    From last week, the crc vm is hitting with high io
    leading to all controller manager pods going to
    crashloopbackoff state.
    
    We also added Baremetal integration in install_yamls.
    
    In order to fix the CI and finish the EDPM deployment
    on same node.
    
    * We are reducing the rabbitmq resources to avoid high io.
    * Disabling bmo_setup[1] during meta operator installation.
    
    These are short term solution to unblock the CI.
    In future, we will be using bigger node.
    
    [1]. https://review.rdoproject.org/r/c/config/+/48523
    
### DIFF
--- a/ci/playbooks/deploy_podified_openstack.yaml
+++ b/ci/playbooks/deploy_podified_openstack.yaml
@@ -67,6 +67,21 @@
         oc rsh openstackclient openstack compute service list;
         oc rsh openstackclient openstack network agent list;
 
+    - name: Patch rabbitmq resources for lower resource consumption
+      changed_when: false
+      ansible.builtin.shell: |
+        crname=$(oc get openstackcontrolplane -o name)
+        oc patch ${crname} --type json \
+          -p='[{"op": "replace", "path": "/spec/rabbitmq/templates/rabbitmq/resources/requests/cpu", "value": 500m}]'
+        oc patch ${crname} --type json \
+          -p='[{"op": "replace", "path": "/spec/rabbitmq/templates/rabbitmq/resources/requests/memory", "value": 500Mi}]'
+        oc patch ${crname} --type json \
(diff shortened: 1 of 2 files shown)


raukadah:
recheck

karelyatin (approved):

fao89 (approved):

raukadah:
recheck

rabi (changes requested):
Without BMO  openstack-baremetal-controller would crash. We should not drop that.

rabi (commented):
Looks like it's only changed in the rdo edpm job where crc is constrained with resources. That should be fine I guess.

ci/playbooks/deploy_podified_openstack.yaml
```diff
@@ -67,6 +67,21 @@
         oc rsh openstackclient openstack compute service list;
         oc rsh openstackclient openstack network agent list;
 
+    - name: Patch rabbitmq resources for lower resource consumption
+      changed_when: false
+      ansible.builtin.shell: |
+        crname=$(oc get openstackcontrolplane -o name)
+        oc patch ${crname} --type json \
+          -p='[{"op": "replace", "path": "/spec/rabbitmq/templates/rabbitmq/resources/requests/cpu", "value": 500Mi}]'
```
karelyatin, line 75: cpu should be 500m, not sure if 500Mi works

karelyatin: looking job logs i don't see rabbitmq pods so seems related.

karelyatin (changes requested):

karelyatin (approved):
/lgtm

PR #256, opened 2023-05-10
Add preProvisioned: true for edpm_deploy

rabi:
This is required after we changed default to preProvisioned: false. With preProvisioned: true there won't be any baremetal deployment attempted.


rabi:
This would be only required when dataplane-operator is bumped in openstack-operator bundle.

rabi:
recheck

rabi:
recheck

rabi:
recheck

fao89 (approved):

PR #257, opened 2023-05-10
Add horizon kuttl commands

jparoly:
Change-Id: I262fecaf8d0db8d348bffa6cfa4ab7e3f9cf8eec

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,6 +177,8 @@ HORIZON_BRANCH      ?= main
 HORIZON             ?= config/samples/horizon_v1beta1_horizon.yaml
 HORIZON_CR          ?= ${OPERATOR_BASE_DIR}/horizon-operator/${HORIZON}
 HORIZON_DEPL_IMG    ?= unused
+HORIZON_KUTTL_CONF  ?= ${OPERATOR_BASE_DIR}/horizon-operator/kuttl-test.yaml
+HORIZON_KUTTL_DIR   ?= ${OPERATOR_BASE_DIR}/horizon-operator/tests/kuttl/tests
 
 # Heat
 HEAT_IMG            ?= quay.io/openstack-k8s-operators/heat-operator-index:latest
@@ -1154,6 +1156,17 @@ glance_kuttl: namespace input openstack_crds openstack_storage_crds deploy_clean
 	make glance_kuttl_run
 	make deploy_cleanup
 	make cleanup
+
+.PHONY: horizon_kuttl_run
+horizon_kuttl_run: ## runs kuttl tests for the horizon operator, assumes that everything needed for running the test was deployed beforehand.
+	kubectl-kuttl test --config ${HORIZON_KUTTL_CONF} ${HORIZON_KUTTL_DIR}
+
+.PHONY: horizon_kuttl
+horizon_kuttl: namespace input openstack_crds openstack_storage_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy infra horizon_deploy_prep horizon ## runs kuttl tests for the horizon operator. Installs openstack and openstack-storage crds, mariadb, keystone and horizon operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
+	make horizon_kuttl_run
+	make deploy_cleanup
+	make cleanup
+	make infra_cleanup
 
 ##@ HORIZON
 .PHONY: horizon_prep


abays:
recheck

Makefile
```diff
@@ -1155,6 +1157,16 @@ glance_kuttl: namespace input openstack_crds openstack_storage_crds deploy_clean
 	make deploy_cleanup
 	make cleanup
 
+.PHONY: horizon_kuttl_run
+horizon_kuttl_run: ## runs kuttl tests for the horizon operator, assumes that everything needed for running the test was deployed beforehand.
+	kubectl-kuttl test --config ${HORIZON_KUTTL_CONF} ${HORIZON_KUTTL_DIR}
+
+.PHONY: horizon_kuttl
+horizon_kuttl: namespace input openstack_crds openstack_storage_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy horizon_deploy_prep horizon ## runs kuttl tests for the horizon operator. Installs openstack and openstack-storage crds, mariadb, keystone and horizon operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
```
kajinamit, line 1165: horizon requires infra operator for memcached. Probably we should add it to the dependency ?

abays: Does it also need `memcached_deploy` then?

abays: Also, `cleanup` and `deploy_cleanup` are missing includes for Horizon, Infra and Memcached.  I think they would need them for these new targets to clean properly.

Makefile
```diff
@@ -1155,6 +1157,16 @@ glance_kuttl: namespace input openstack_crds openstack_storage_crds deploy_clean
...
 
+.PHONY: horizon_kuttl_run
+horizon_kuttl_run: ## runs kuttl tests for the horizon operator, assumes that everything needed for running the test was deployed beforehand.
+	kubectl-kuttl test --config ${HORIZON_KUTTL_CONF} ${HORIZON_KUTTL_DIR}
+
+.PHONY: horizon_kuttl
+horizon_kuttl: namespace input openstack_crds openstack_storage_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy infra horizon_deploy_prep horizon ## runs kuttl tests for the horizon operator. Installs openstack and openstack-storage crds, mariadb, keystone and horizon operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
+	make horizon_kuttl_run
+	make deploy_cleanup
+	make cleanup
```
cescgina, line 1168: Missing `infra_cleanup` here

Makefile
```diff
@@ -1155,6 +1157,17 @@ glance_kuttl: namespace input openstack_crds openstack_storage_crds deploy_clean
 	make deploy_cleanup
 	make cleanup
 
+.PHONY: horizon_kuttl_run
+horizon_kuttl_run: ## runs kuttl tests for the horizon operator, assumes that everything needed for running the test was deployed beforehand.
+	kubectl-kuttl test --config ${HORIZON_KUTTL_CONF} ${HORIZON_KUTTL_DIR}
+
+.PHONY: horizon_kuttl
+horizon_kuttl: namespace input openstack_crds openstack_storage_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy infra horizon_deploy_prep horizon ## runs kuttl tests for the horizon operator. Installs openstack and openstack-storage crds, mariadb, keystone and horizon operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
+	make horizon_kuttl_run
```
cescgina, line 1166: I tried to run `horizon_kuttl` and it's always failing on the first try because it tries to deploy the horizon resources before the controller-manager pod is up. Since it uses the webhook to set the containerImage, we need to make sure the operator is sueccesfully installed before proceeding. I believe something like `oc wait deployment horizon-operator-controller-manager -n ${NAMESPACE} --for condition=Available --timeout=120s` should work, but @abays  will now better

cescgina: Thanks, I tried that but it doesn't seem to find the deployment. I checked the labels from the horizon-controller-manager deployment and it doesn't seem to have the `app.kubernetes.io/name=horizon-operator` but rather a `app.kubernetes.io/name=deployment` label

abays: Ah, fair point.  There is still some labeling standardization that needs to happen for certain operators, with this being one of them.  Ultimately they should all have something like `openstack.org/operator-name: horizon` and we would use such a label as a selector here.  I tried to change some of these in the course of the webhooks work but did not cover all them.

abays: Actually, Horizon already has the right format for the label mentioned above.  But other operators are missing it and then we'd need to use the standard format in the `make wait` script.

jparoly:
@abays anything else we need prior to merging this PR?

abays:
> @abays anything else we need prior to merging this PR?

We can deal with the "wait for operator controller-manager pod to be ready" issue elsewhere I suppose.  This PR doesn't cause it anyhow, it just makes it possible to notice.  I'm in the process of standardizing the operators' controller-manager pods' labeling such that we can use `make wait` in the near future in these KUTTL contexts to ensure operator availability before continuing with the tests.

/lgtm

abays (approved):

PR #260, opened 2023-05-11
Customize the amount of storage associated to Ceph

fmount:
We recently hit issues in CI because this pod is evicted with the following message:

```
Usage of EmptyDir volume "data" exceeds the limit "500Mi" 
```

This patch exposes the `DATA_SIZE` parameter that can be used to customize the amount
of space reserved to the toy `Ceph` cluster (especially for `CI` purposes).


fmount:
Tested locally (`make ceph DATA_SIZE=2Gi TIMEOUT=90`) and see the value being updated:


```
Volumes:
  data:
    Type:       EmptyDir (a temporary directory that shares a pod's lifetime)
    Medium:
    SizeLimit:  2Gi
  run:
    Type:       EmptyDir (a temporary directory that shares a pod's lifetime)
    Medium:
    SizeLimit:  2Gi
  log:
    Type:       EmptyDir (a temporary directory that shares a pod's lifetime)
    Medium:
    SizeLimit:  2Gi
```

PR #269, opened 2023-05-15
Fix CINDER_BRANCH after rename

ASBishop:


dprince:
@ASBishop thanks. I'm going to just all the remaining renames here at one time to save time: https://github.com/openstack-k8s-operators/install_yamls/pull/270

PR #273, opened 2023-05-17
Apply OVNController CRD if it's present

booxter:
The new OVNController CRD is about to be introduced in ovn-operator (it's the old OVS CRD from ovs-operator - we are merging ovn- and ovs- into one operator).

This change will make sure that the new manifest from ovn-operator repo is applied, if present. This is needed for CI of ovn-operator repo to pass.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,8 @@ OVNDBS              ?= config/samples/ovn_v1beta1_ovndbcluster.yaml
 OVNDBS_CR           ?= ${OPERATOR_BASE_DIR}/ovn-operator/${OVNDBS}
 OVNNORTHD           ?= config/samples/ovn_v1beta1_ovnnorthd.yaml
 OVNNORTHD_CR        ?= ${OPERATOR_BASE_DIR}/ovn-operator/${OVNNORTHD}
+OVNCONTROLLER       ?= config/samples/ovn_v1beta1_ovncontroller.yaml
+OVNCONTROLLER_CR    ?= ${OPERATOR_BASE_DIR}/ovn-operator/${OVNCONTROLLER}
 # TODO: Image customizations for all OVN services
 OVN_KUTTL_CONF      ?= ${OPERATOR_BASE_DIR}/ovn-operator/kuttl-test.yaml
 OVN_KUTTL_DIR       ?= ${OPERATOR_BASE_DIR}/ovn-operator/tests/kuttl/tests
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -714,6 +716,7 @@ ovn_deploy_prep: ovn_deploy_cleanup ## prepares the CR to install the service ba
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR} ${DEPLOY_DIR}
 	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(OVN_BRANCH),-b ${OVN_BRANCH}) ${OVN_REPO} "${OPERATOR_NAME}-operator" && popd
 	cp ${OVNDBS_CR} ${OVNNORTHD_CR} ${DEPLOY_DIR}
+	test -f ${OVNCONTROLLER_CR} && cp ${OVNCONTROLLER_CR} ${DEPLOY_DIR} || true
```
karelyatin, line 719: I don't think this is related or will help to the issue we seeing in ovn-operator merge PR as these copyied CRs are not used in kuttl job, we would need this but as Depends-On on the ovn operator PR, so can add ${OVNCONTROLLER_CR} to L718 itself.

booxter: This is related. You can see in kuttl log here: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_ovn-operator/52/pull-ci-openstack-k8s-operators-ovn-operator-main-ovn-operator-build-deploy-kuttl/1658601138502504448#1:build-log.txt%3A435

This is because `ovn_kuttl` target depends on `ovn_deploy_prep` that executes the `cp` command. Without it, the kuttl job in ovn-operator will not copy the new CRD yaml file and then will fail applying `OVNController` resources because of unknown `Kind`.

booxter: OK I think I understand what you mean: this `cp` command is copying samples to apply, not CRD definitions. Instead, we do `skopeo copy` from a container in Quay. But if a new CRD is introduced, with `kuttl` tests that cover it, then we miss the new CRDs in `manifests` directory. I think the proper solution is for the CI script (is it in `ci-operator` somewhere?) to copy the manifests from the repo under test in addition to what’s included in the bundle image. Otherwise it's a Catch 22 situation where we can't pass CI for a patch because CI doesn't apply manifests from the PR. Thoughts?

booxter:
This is not needed. It should be solved by CI pulling bundle image from the just built artifacts.

https://github.com/openshift/release/pull/39457

karelyatin:
> This is not needed. It should be solved by CI pulling bundle image from the just built artifacts.
> 
> [openshift/release#39457](https://github.com/openshift/release/pull/39457)

Yes not need to fix CI, but would atleast need a change here to add OVNController CR as per previous comment and also need to cleanup ovs bits from Makefile.

PR #275, opened 2023-05-17
Gather latest available CentOS 9 stream

rebtoor:
Since CentOS Stream images are often removed from server, we can retrieve the lastest available image using `CentOS-Stream-GenericCloud-9-latest.x86_64.qcow2` as file name.

### DIFF
--- a/devsetup/scripts/gen-edpm-compute-node.sh
+++ b/devsetup/scripts/gen-edpm-compute-node.sh
@@ -29,7 +29,7 @@ EDPM_COMPUTE_NETWORK=${EDPM_COMPUTE_NETWORK:-default}
 EDPM_COMPUTE_NETWORK_TYPE=${EDPM_COMPUTE_NETWORK_TYPE:-network}
 DATAPLANE_DNS_SERVER=${DATAPLANE_DNS_SERVER:-192.168.122.1}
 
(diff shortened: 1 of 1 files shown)


rebtoor:
recheck

devsetup/scripts/gen-edpm-compute-node.sh
```diff
@@ -29,7 +29,8 @@ EDPM_COMPUTE_NETWORK=${EDPM_COMPUTE_NETWORK:-default}
 EDPM_COMPUTE_NETWORK_TYPE=${EDPM_COMPUTE_NETWORK_TYPE:-network}
 DATAPLANE_DNS_SERVER=${DATAPLANE_DNS_SERVER:-192.168.122.1}
 
-CENTOS_9_STREAM_URL=${CENTOS_9_STREAM_URL:-"https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-GenericCloud-9-20230516.0.x86_64.qcow2"}
```
slagle, line 32: We can just use https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-GenericCloud-9-latest.x86_64.qcow2

slagle (changes requested):

devsetup/scripts/gen-edpm-compute-node.sh
rebtoor: yeah, we should! My bad, I didn't saw the file. :facepalm:

slagle (approved):

PR #276, opened 2023-05-17
Update crc nodeset and parameters

raukadah:
Recently we have added centos-9-crc-3xl nodeset for running EDPM job having more memory. It will help to avoid overloading the CRC vm with requests and high cpu usage.

This patch uses the nodeset from ci-framework[1].
centos-9-crc-xxl will be used from podified job
and centos-9-crc-3xl will be used for edpm job.

The same crc parameters will be used in both jobs.

[1]. https://github.com/openstack-k8s-operators/ci-framework/blob/main/zuul.d/nodeset.yaml

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -4,10 +4,6 @@
     parent: base-crc
     abstract: true
     timeout: 10800
-    nodeset:
-      nodes:
-        name: controller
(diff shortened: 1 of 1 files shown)


rebtoor (approved):

raukadah:
recheck

raukadah:
recheck

PR #277, opened 2023-05-17
Deploy "N" EDPM nodes

fao89:
ref: https://github.com/openstack-k8s-operators/dataplane-operator/pull/189

`make edpm_deploy DATAPLANE_TOTAL_NODES=4 DATAPLANE_SINGLE_NODE=false`

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -3,6 +3,7 @@ KUBEADMIN_PWD ?= 12345678
 PULL_SECRET  ?= ${PWD}/pull-secret.txt
 CRC_DEFAULT_NETWORK_IP ?= 192.168.122.10
 EDPM_COMPUTE_SUFFIX ?= 0
+EDPM_TOTAL_NODES ?= 1
 EDPM_COMPUTE_IP ?= 192.168.122.100
 OPENSTACK_RUNNER_IMG ?= quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
 EDPM_NETWORK_CONFIG_TEMPLATE ?= templates/single_nic_vlans/single_nic_vlans.j2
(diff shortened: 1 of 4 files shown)


fultonj (commented):
I normally just loop on the make command so I'm not sure this patch is necessary.
```
 for I in $(seq 0 $NODES); do
  make edpm_compute EDPM_COMPUTE_SUFFIX=$I
done
```
Did someone request it or does the CI need it?

If we're going to have this loop do we also want to add it to `edpm_compute_repos`? If we don't we'd just loop it anway.

nit: should the variable name of the loop `for number` be capitalized `for NUMBER`?

Overall, I think we need to determine if we need the patch to add the loop when the user can just use their own loop.

fao89:
@fultonj I agree, I'll drop the loop.
About `edpm_compute_repos` instead of looping, I believe we can use a service:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/189
the idea is making easier to do scale tests (OSP-23847), so instead of calling `edpm_compute_repos` 200 times, it would be better t have it set in parallel through AEE

devsetup/Makefile
```diff
@@ -97,11 +98,27 @@ edpm_compute_repos: ## Configure EDPM compute VM with repositories
 	$(eval $(call vars))
 	scripts/edpm-compute-repos.sh ${EDPM_COMPUTE_SUFFIX}
 
+.PHONY: edpm_nodes
```
slagle, line 101: I think this target could probably just take the place of edpm_compute/edpm_compute_cleanup. No reason to have multiple ways of doing it. If you just want one node, you'd set EDPM_TOTAL_NODES=1.

devsetup/Makefile
```diff
@@ -97,11 +98,27 @@ edpm_compute_repos: ## Configure EDPM compute VM with repositories
 	$(eval $(call vars))
 	scripts/edpm-compute-repos.sh ${EDPM_COMPUTE_SUFFIX}
```
slagle, line 99: I'd then add the loop here as well so that you don't need to call this target multiple times.

slagle (changes requested):

scripts/gen-edpm-kustomize.sh
```diff
@@ -71,6 +71,11 @@ patches:
     - op: replace
       path: /spec/nodes/edpm-compute-0/node/ansibleSSHPrivateKeySecret
       value: ${EDPM_ANSIBLE_SECRET}
+    - op: replace
+      path: /spec/roles/edpm-compute/services
+      value:
+      - repo-setup
```
fao89, line 77: requires: https://github.com/openstack-k8s-operators/dataplane-operator/pull/189

fultonj (commented):
Looks good to me but I didn't test it in my environment yet.

fao89:
recheck

fao89:
recheck

fao89:
recheck

PR #278, opened 2023-05-18
Add deps for dataplane-operator kuttl tests

slagle:
The SSH key secret needs to exist for the dataplane-operator kuttl tests
to pass. Call the script that creates the secret before running tests.

The initial dataplane-operator services also need to be created from the
config/services directory in the operator.

The tests also require the openstack-ansibleee-operator. The ansibleee
Makefile target is added as a dependency for the tests.


rebtoor (approved):

rebtoor:
recheck

PR #279, opened 2023-05-18
Update certain ansibleVars

rabi:
- non RH ntp servers
- Increase /var for container storage


Sandeepyadav93:
/lgtm
/approve

Ntp correction is needed to run on the public cloud and without the growvols_args it was failing because of the smaller var size.

PR #280, opened 2023-05-18
Add cleanup targets to dataplane_kuttl

slagle:
The cleanup targets need to be executed after kuttl tests since the ci
runs a deploy test after a kuttl test, and the tests expect a cleaned
environment.

Also adds the dataplane_cleanup and ansibleee_cleanup targets to the
cleanup target so that they are actually cleaned up.


slagle:
recheck

rebtoor (approved):

PR #281, opened 2023-05-18
Workaround deployment label for horizon

jparoly:
There is still some labeling standardization that needs to happen for certain operators, with this being one of them. Ultimately they should all have something like openstack.org/operator-name: horizon and we would use such a label but right now the format in the wait script is still app.kubernetes.io/name=${OPERATOR_NAME}-operator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1219,9 +1219,9 @@ horizon_kuttl_run: ## runs kuttl tests for the horizon operator, assumes that ev
 	kubectl-kuttl test --config ${HORIZON_KUTTL_CONF} ${HORIZON_KUTTL_DIR}
 
 .PHONY: horizon_kuttl
-horizon_kuttl: namespace input openstack_crds openstack_storage_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy infra horizon_deploy_prep horizon ## runs kuttl tests for the horizon operator. Installs openstack and openstack-storage crds, mariadb, keystone and horizon operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
-	$(eval $(call vars,$@,horizon))
-	make wait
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1201,6 +1201,8 @@ horizon_kuttl_run: ## runs kuttl tests for the horizon operator, assumes that ev
 
 .PHONY: horizon_kuttl
 horizon_kuttl: namespace input openstack_crds openstack_storage_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy infra horizon_deploy_prep horizon ## runs kuttl tests for the horizon operator. Installs openstack and openstack-storage crds, mariadb, keystone and horizon operators and cleans up previous deployments before running the tests and, add cleanup after running the tests.
+	timeout 120s bash -c 'until [ "$(oc get deployment -l openstack.org/operator-name: horizon -n ${NAMESPACE} -o name)" != "" ]; do sleep 1; done'
+	oc wait deployment -l openstack.org/operator-name: horizon -n ${NAMESPACE} --for condition=Available --timeout=120s
```
abays, line 1205: This pattern is almost available across all operators.  Just waiting for these two reviews and then we can update the label used by `make wait`:

https://github.com/openstack-k8s-operators/cinder-operator/pull/204
https://github.com/openstack-k8s-operators/keystone-operator/pull/245

cescgina:
This recent change should take care of the problem for all operators: https://github.com/openstack-k8s-operators/install_yamls/pull/282, so there will be no need for this PR

jparoly:
When I run the tests locally with this change I get: `make: *** [Makefile:1204: horizon_kuttl] Error 124`

jparoly:
> This recent change should take care of the problem for all operators: #282, so there will be no need for this PR

ack, closing

jparoly:
> > This recent change should take care of the problem for all operators: #282, so there will be no need for this PR
> 
> ack, closing

ummm looks like we do need `make horizon_deploy` in the setup for the test though, correct?

jparoly:
> 

like so: `namespace input openstack_crds openstack_storage_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy infra horizon_deploy_prep horizon horizon_deploy `

jparoly:
> > 
> 
> like so: `namespace input openstack_crds openstack_storage_crds deploy_cleanup mariadb mariadb_deploy keystone keystone_deploy infra horizon_deploy_prep horizon horizon_deploy `

well, maybe not, the tests are passing as is on main

PR #282, opened 2023-05-19
Wait for operator before running kuttl tests

karelyatin:
Before kuttl tests starts we need to ensure
operator is ready so defaults from webhook can
be used.

The issue was uncovered in [1] where new
controller is being added, kuttl tests fails
as webhook was not ready when tests were triggered.

Also update label in operator wait script to the one that is being replicated across operators. Still some
operators don't have these set, for those make
targets can be updated later.

[1] https://github.com/openstack-k8s-operators/ovn-operator/pull/52


cescgina (commented):
lgtm, thanks, it was coming up in other kuttl tests as well. Could you also  add the `wait` to the horizon kuttl tests (line 1223)?

abays (commented):
Thanks @karelyatin.  I'll give it a quick test and then approve.

abays (approved):
/lgtm

PR #285, opened 2023-05-22
Add controlplane_deploy to baremetal deploy target

rabi:
Also removes NETWORK_ISOLATION=false


devsetup/Makefile
```diff
@@ -90,23 +90,24 @@ crc_attach_default_interface_cleanup: ## Detach default libvirt network from CRC
 edpm_compute_baremetal: export PROVISIONING_INTERFACE=${BMAAS_NETWORK_NAME}
 edpm_compute_baremetal: export BMAAS_NETWORK_IPADDRESS=172.22.0.3
 edpm_compute_baremetal: export BMAAS_BRIDGE_IPADDRESS=172.22.0.2
-edpm_compute_baremetal: ## Deploy dataplane with BMAAS
+edpm_compute_baremetal: ## Deploy controlplane and dataplane with BMAAS
```
abays, line 93: Should the target name change here if this is now deploying the control plane in addition to the data plane?  I'm just wondering if the `_compute_` in the name is misleading in that context.

abays: Alright, works for me then

PR #286, opened 2023-05-22
Use qcow2 format for cirros image

raukadah:
Since qemu-img command is not found in openstackclient pod and the cirros image conversion from img to raw fails. We can directly use the qcow2 format to avoid this issue.

It also removes workaround for lp/1999553. It is no longer needed.


raukadah:
Here is the output from the changes: https://review.rdoproject.org/zuul/build/94648a596cf54326a00ee7b2a1b0c4ce/console#5/0/9/controller

PR #289, opened 2023-05-23
openstack_operators namespace

dprince:


Makefile
```diff
@@ -337,6 +339,21 @@ crc_storage_cleanup: ## cleanup local storage PVs in CRC vm
 	if oc get sc ${STORAGE_CLASS}; then oc delete sc ${STORAGE_CLASS}; fi
 	bash scripts/delete-pv.sh
 
+##@ OPERATOR_NAMESPACE
+.PHONY: operator_namespace
+operator_namespace: export NAMESPACE=${OPERATOR_NAMESPACE}
+operator_namespace: ## creates the namespace specified via OPERATOR_NAMESPACE env var (defaults to openstack)
```
cescgina, line 345: The comment here should read  `defaults to openstack-operators`, shouldn't it?

raukadah:
@dprince We have merged https://review.rdoproject.org/r/c/config/+/48702. We need to update this patch by adding
```
operators_namespace: openstack-operators
```
at https://github.com/openstack-k8s-operators/install_yamls/blob/main/zuul.d/jobs.yaml#L15 . The zuul job will pass.

Makefile
```diff
@@ -734,12 +751,12 @@ ovn_prep: ## creates the files to install the operator using olm
 	bash scripts/gen-olm.sh
 
 .PHONY: ovn
-ovn: namespace ovn_prep ## installs the operator, also runs the prep step. Set OVN_IMG for custom image.
+ovn: openstack_namespace ovn_prep ## installs the operator, also runs the prep step. Set OVN_IMG for custom image.
 	$(eval $(call vars,$@,ovn))
 	oc apply -f ${OPERATOR_DIR}
 
 .PHONY: ovn_cleanup
-ovn_cleanup: ## deletes the operator, but does not cleanup the service resources
```
cescgina, line 742: same as before

cescgina:
We want to update `scripts/operator-wait` to use `$OPERATOR_NAMESPACE` instead of `$NAMESPACE`

Makefile
dprince: yes, good catch

Makefile
dprince: removed. thanks.

dprince:
I think this is now blocked on landing this: https://github.com/openstack-k8s-operators/telemetry-operator/pull/101

dprince:
Latest tests fail due to 'public endpoint for network service in regionOne region not found'. Not sure that is related to this PR, need to try to reproduce locally

abays:
> Latest tests fail due to 'public endpoint for network service in regionOne region not found'. Not sure that is related to this PR, need to try to reproduce locally

I don't think it's a problem specific to this PR.  I see a newer PR with the same issue: https://review.rdoproject.org/zuul/build/37dcc462c9d440ccbc4a81ccdc0ff920

PR #290, opened 2023-05-24
Add repo-setup service instead of replacing services list

slagle:
Instead of replacing the entire list of composable services for the
dataplane-operator, use the JSON patch add op to insert the repo-setup
service at the beginning of the list[1]. This eliminates the need to also
maintain the list of needed services here, since we can just insert at
the beginning of the list instead of replacing the whole list.

[1] https://datatracker.ietf.org/doc/html/rfc6902#appendix-A.2

### DIFF
--- a/scripts/gen-edpm-kustomize.sh
+++ b/scripts/gen-edpm-kustomize.sh
@@ -71,11 +71,9 @@ patches:
     - op: replace
       path: /spec/nodes/edpm-compute-0/node/ansibleSSHPrivateKeySecret
       value: ${EDPM_ANSIBLE_SECRET}
-    - op: replace
-      path: /spec/roles/edpm-compute/services
-      value:
-      - repo-setup
-      - configure-network
+    - op: add
+      path: /spec/roles/edpm-compute/services/0
+      value: repo-setup
     - op: replace
       path: /spec/roles/edpm-compute/openStackAnsibleEERunnerImage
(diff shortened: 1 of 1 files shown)


scripts/gen-edpm-kustomize.sh
```diff
@@ -71,11 +71,10 @@ patches:
     - op: replace
       path: /spec/nodes/edpm-compute-0/node/ansibleSSHPrivateKeySecret
       value: ${EDPM_ANSIBLE_SECRET}
-    - op: replace
-      path: /spec/roles/edpm-compute/services
+    - op: add
+      path: /spec/roles/edpm-compute/services/0
```
raukadah, line 75: https://logserver.rdoproject.org/90/290/ae1893d747a51f8060943cd7a50878fad66bf934/github-check/centos-9-crc-singlenode-podified-edpm-deployment/43e4dd6/job-output.txt
```
 oc apply -f -\n+ popd\n+ popd\nThe OpenStackDataPlane \"openstack-edpm\" is invalid: spec.roles.edpm-compute.services[0]: Invalid value: \"array\": spec.roles.edpm-compute.services[0] in body must be of type string: \"array\"
```

raukadah (commented):
The EDPM Zuul job failure is linked with the patch itself

raukadah (approved):
/lgtm

PR #291, opened 2023-05-25
Increase the timeout for wait target

rabi:
openstack-operator takes long time on ci crc nodes with less resources used for edpm jobs to be deployed. It's better we increase the timeout.


abays (approved):
/lgtm

PR #292, opened 2023-05-25
Remove ssh-keyscan from gen-edpm-compute-node script

arxcruz:
The ssh-keyscan will fail if there is no info on known_hosts for the ip address 192.168.122.1 and so the virt-customize will command fail to run, specially when you're trying to deploy locally in your machine instead of using nested vms.

### DIFF
--- a/devsetup/scripts/gen-edpm-compute-node.sh
+++ b/devsetup/scripts/gen-edpm-compute-node.sh
@@ -181,6 +181,8 @@ sed -i s/dhcp/none/g $NETSCRIPT
 sed -i /PERSISTENT_DHCLIENT/d $NETSCRIPT
 EOF
 
+chmod +x ${OUTPUT_BASEDIR}/${EDPM_COMPUTE_NAME}-firstboot.sh
+
 if [ ! -f ${DISK_FILEPATH} ]; then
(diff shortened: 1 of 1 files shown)


cjeanner:
/approve

Note: this ssh-keyscan was a forgotten bit when I cleaned the ssshuttle related tasks/content. We don't need to inject the hypervisor host key in the VM.

devsetup/scripts/gen-edpm-compute-node.sh
```diff
@@ -181,6 +181,8 @@ sed -i s/dhcp/none/g $NETSCRIPT
 sed -i /PERSISTENT_DHCLIENT/d $NETSCRIPT
 EOF
 
+chmod +x ${OUTPUT_BASEDIR}/${EDPM_COMPUTE_NAME}-firstboot.sh
```
jpodivin, line 184: I'm not 100% sure how is this related to the ssh-keyscan removal. Maybe it would be a good idea to make a note of it in PR or even in the commit msg.

arxcruz: Updated the commit message.

arxcruz:
recheck

arxcruz:
recheck

PR #297, opened 2023-05-25
Fix rabbitmq_deploy_cleanup target when CRD doesn't exist

slagle:
The rabbitmq CRD's may not always have been loaded when
rabbitmq_deploy_cleanup is called. Since rabbitmq_deploy_cleanup is now
part of the deploy_cleanup target which is called by many of the kuttl
tests cleanup, it needs to not fail if the CRD's don't exist. Note that
--ignore-not-found does not supress the error when the resource kind
does not exist.


abays (approved):
/lgtm

abays (approved):
/lgtm

PR #302, opened 2023-05-31
Remove explicit OpenStackProvisionServer CR

rabi:
When OpenStackProvisionerServer CR is not created and provided it would be auto-created. Let's use that as the default.


abays (approved):
I don't think the CI failures are related to this PR

/lgtm

abays:
recheck

PR #303, opened 2023-05-31
Also delete operators group for openstack cleanup

bogdando:
When calling make openstack_cleanup, delete the deployed operators group for openstack alongside its subscription


raukadah:
recheck

bogdando:
looks no longer relevant

PR #304, opened 2023-06-01
Make devsetup/cifwm_prepare not fail when rerun

queria:
Currently if one tries to rerun cifwm_prepare it fails on git cloning attempt since the directory is already in place.

Example of failure:
> + make cifmw_prepare
> git clone https://github.com/openstack-k8s-operators/ci-framework ci-framework
> fatal: destination path 'ci-framework' already exists and is not an empty directory.
> make: *** [Makefile:180: cifmw_prepare] Error 128

As that is not expected from make target, make it so that this target depends on the directory existing target, which itself calls the git clone for its creation.


cjeanner (commented):
If really, I'd rather get something checking for the directory availability, if already there, `git pull` in order to ensure we're up-to-date... ?

devsetup/Makefile
```diff
@@ -175,10 +175,12 @@ edpm_deploy_instance: ## Spin a instance on edpm node
...
 	-oc delete openstackansibleee deploy-external-dataplane-compute
 
-.PHONY: cifmw_prepare
-cifmw_prepare: ## Clone the ci-framework repository in the ci-framework directory. That location is ignored from git.
+./ci-framework:
 	git clone https://github.com/openstack-k8s-operators/ci-framework ci-framework
 
+.PHONY: cifmw_prepare
+cifmw_prepare: ./ci-framework ## Clone the ci-framework repository in the ci-framework directory. That location is ignored from git.
+
```
queria, line 183: yes here we can always just do 'cd ci-framewrok; git pull' (as this target depends on existance of ./ci-framework itself now)

cjeanner: I wouldn't expect a dev to change things there in, and re-run `make cifmw_prepare` to be honest. Safeguards are always good though.

queria: It does conditional pull now - so should be safe even for modified env.

devsetup/Makefile
```diff
@@ -175,10 +175,13 @@ edpm_deploy_instance: ## Spin a instance on edpm node
 edpm_play_cleanup: ## Cleanup EDPM openstackansibleee resource
 	-oc delete openstackansibleee deploy-external-dataplane-compute
 
-.PHONY: cifmw_prepare
-cifmw_prepare: ## Clone the ci-framework repository in the ci-framework directory. That location is ignored from git.
+./ci-framework:
```
marios, line 178: names ;) 
i'd call this something more descriptive though, like ci-framework-clone or somesuch.

marios (commented):
makes sense

devsetup/Makefile
queria: Oh yep, i would guess the aim was to also invoke the setup part, not sure - with that (renaming...) I would prefer to wait for Cedric.

cjeanner: maybe we can add some more "checks" in there in order to ensure we can consume the framework - for instance "is ansible present" and related? or, even, run an actual task that will create the venv and make things ready to consume?

I really added that one in the very-very early stages of the framework, just to show "hey, we're here now". We can of course evolve (actually, we must evolve ;)).

queria: So what would be the recommendation for this PR please?
Rename (but that would involve changing all the docs / local flows too)? Or add more setup steps (any specific ones to add, or I can pick any)? Or leave it be as is?

marios: thanks for discussion today lets get this merged and you can update for more functionality as needed (imo)

marios (approved):
lgtm

PR #307, opened 2023-06-05
Improve EDPM scripts

fao89:


fultonj (approved):
+2

devsetup/scripts/gen-ansibleee-ssh-key.sh
```diff
@@ -32,6 +32,13 @@ fi
 
 pushd ${OUTPUT_DIR}
 
+if oc get secret dataplane-ansible-ssh-private-key-secret -n ${NAMESPACE} 2>&1 1>/dev/null; then
```
rabi, line 35: Why do we need to check for the key and let things to fail? Was that added back intentionally? We do a dry-run[1] so it won't fail if the secret does exist, would just update it.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/gen-ansibleee-ssh-key.sh#L59-L60

PR #312, opened 2023-06-08
Adding target for crc path setup

jpodivin:
The path and env variables will now be set automatically without user having to manualy ask crc for them. The README has been adjusted to reflect the newly simplified procedure.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -343,7 +343,7 @@ wait: ## wait for an operator's controller-manager pod to be ready (requires OPE
 
 ##@ CRC
 .PHONY: crc_storage
-crc_storage: ## initialize local storage PVs in CRC vm
+crc_storage: crc_env_vars ## initialize local storage PVs in CRC vm
 	$(eval $(call vars,$@))
 	bash scripts/create-pv.sh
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -359,6 +359,10 @@ endif
 	if oc get sc ${STORAGE_CLASS}; then oc delete sc ${STORAGE_CLASS}; fi
 	bash scripts/delete-pv.sh
 
+.PHONY: crc_env_vars
+crc_env_vars: ## set your CRC ENV variables and PATH for 'oc'
+	eval $(crc oc-env)
```
stuggi, line 364: I think this is ok, since `crc_storage` is also kind of expects to be a crc, while its not a must. in case of not using CRC crc_env_vars will fail. @dprince thoughts?

steveb: As an aside, I think the crc_storage target is misnamed, I use it on single node openshift environments. How about a new `local_storage` target, then a `crc_storage` target which depends on `crc_env_vars` and `local_storage`?

jpodivin:
Irrelevant

PR #127, opened 2023-03-03
Deploy CRC with network isolation

stuggi:
* clone install_yamls
```
git clone https://github.com/openstack-k8s-operators/install_yamls.git
```
* install CRC
```
cd install_yamls/devsetup
CPUS=12 MEMORY=25600 DISK=100 make crc
```
* login to OCP
```
eval $(crc oc-env)
oc login -u kubeadmin -p 12345678 https://api.crc.testing:6443
```
* attach libvirt default network to the crc (default IP 192.168.122.10)
```
make crc_attach_default_interface
```
* install nmstate operator
```
cd ..
make nmstate
```
* configure interfaces on the CRC
```
make nncp
```
* create network-attachment-definitions for internalapi, storage and tenant
```
make netattach
```
the following NADs with ip ranges get configured:
internalapi: 172.17.0.30-172.17.0.70
storage:     172.18.0.30-172.18.0.70
tenant:      172.19.0.30-172.19.0.70

* install metallb operator
```
make metallb
```
* create ipaddresspools (internalapi, storage, tenant) and l2advertisement resources (metallb-sytem namespace)
```
make metallb_config
```
the following IPAddressPools with ip ranges get configured:
internalapi: 172.17.0.80-172.17.0.90
storage:     172.18.0.80-172.18.0.90
tenant:      172.19.0.80-172.19.0.90

* create dependencies
```
make crc_storage
make input
```
* install opentack-operator
```
make openstack
```
* deploy ctlplane using core_v1beta1_openstackcontrolplane_network_isolation.yaml sample, check Depends-On PR 
```
export OPENSTACK_CTLPLANE=config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml make openstack_deploy
```

At this point the ctlplane is deployed with the services using isolated networks as specified in the CR sample.

* create edpm VM (default IP 192.168.122.100)
```
cd devsetup
make edpm_compute
```
* deploy edpm compute
```
make edpm_play
```

* wait until finished, then can check the env
```
oc rsh openstackclient
sh-5.1$ openstack compute service list
+--------------------------------------+----------------+------------------------+----------+---------+-------+----------------------------+
| ID                                   | Binary         | Host                   | Zone     | Status  | State | Updated At                 |
+--------------------------------------+----------------+------------------------+----------+---------+-------+----------------------------+
[...]

PR #316, opened 2023-06-11
Fix openstack_crds target

rabi:
This fixes a few things.

- Removes prerequisites run with `openstack_deploy_prep` when NETWORK_ISOLATION=tue. We don't need to do them as they're already done with `openstack_prep`.
- Also sets REGISTRY_AUTH_FILE as all users don't have permission to write to /run/containers. Else we get the error "reading JSON file "/run/containers/1001/auth.json": open /run/containers/1001/auth.json: permission denied".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -459,7 +459,7 @@ openstack_cleanup: operator_namespace## deletes the operator, but does not clean
 
 .PHONY: openstack_deploy_prep
 openstack_deploy_prep: export KIND=OpenStackControlPlane
-openstack_deploy_prep: openstack_deploy_cleanup $(if $(findstring true,$(NETWORK_ISOLATION)), nmstate nncp netattach metallb metallb_config)  ## prepares the CR to install the service based on the service sample file OPENSTACK
+openstack_deploy_prep: openstack_deploy_cleanup ## prepares the CR to install the service based on the service sample file OPENSTACK
 	$(eval $(call vars,$@,openstack))
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR} ${DEPLOY_DIR}
 	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(OPENSTACK_BRANCH),-b ${OPENSTACK_BRANCH}) ${OPENSTACK_REPO} "${OPERATOR_NAME}-operator" && popd


stuggi:
@fao89 you added running network isolation pre-requ in  c05d142813f9b822baf2823b34aeb14ffe4eec10 . you see an issue removing it from `openstack_deploy_prep`?

fao89:
@stuggi I was following the steps you provided in your PR to enable network isolation, so I'm the one who should be asking you if you see an issue removing it

rabi:
>> so I'm the one who should be asking you if you see an issue removing it

Hmm... @dprince Is there an alternative? No one seems the intent of why it was added to `openstack_predeploy` and we're not ready to merge this fix for the current broken kuttl CI jobs across repos?

rabi:
> I used openstack_predeploy in the openstack_crds task as we need to checkout the openstack-operator repo and I didn't want to duplicate it.

That's ok. The issue is `openstack_predeploy` used to run NETWORK_ISOLATION=true targets as prereqs and now they all run with `openstack_crds` and fails for tests where the additional interfaces does not exist.

stuggi:
> @stuggi I was following the steps you provided in your PR to enable network isolation, so I'm the one who should be asking you if you see an issue removing it

then it should be ok. if there is no other reason those steps installed the network components before we deployed the openstack operator https://github.com/openstack-k8s-operators/install_yamls/pull/127

scripts/openstack-crds.sh
```diff
@@ -8,6 +8,8 @@ function extract_crds {
     done
 }
 
+export REGISTRY_AUTH_FILE=$HOME/auth.json
```
dprince, line 11: I think we should remove this from the PR. I'm not seeing this issue in my environment. Also, this seems like something that would be better set/managed in the users $HOME directory environment not hard coded like this.

dprince:
Good to merge other than I think we should remove the ENV variable getting set in scripts/openstack-crds.sh.

rabi:
> Good to merge other than I think we should remove the ENV variable getting set in scripts/openstack-crds.sh.

OK, it does not work for me. But updated the PR to remove that ENV var as @dprince thinks it's some issue on my setup and it works for him without it.

dprince (approved):

slagle (approved):

PR #317, opened 2023-06-12
Customize the updated image path for custom ansible image

slagle:
dataplane-operator commit 17e87a135fc6671129c921560e26fd7ecf259113 moved
the OpenStackAnsibleEEImage field from the node/role to the service so
that the image can be customized per service.

This patch updates install_yamls to kustomize the correct path on the
service when a custom ${DATAPLANE_RUNNER_IMG} is set.

### DIFF
--- a/scripts/gen-edpm-services-kustomize.sh
+++ b/scripts/gen-edpm-services-kustomize.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+#
+# Copyright 2023 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


rebtoor (approved):

rebtoor (commented):
+2 for me. Waiting for another +2 to merge it.

scripts/gen-edpm-services-kustomize.sh
```diff
@@ -0,0 +1,58 @@
...
+kind: Kustomization
+resources:
+namespace: ${NAMESPACE}
+patches:
+- target:
+    kind: ${KIND}
+  patch: |-
+    - op: replace
+      path: /spec/openStackAnsibleEERunnerImage
+      value: ${OPENSTACK_RUNNER_IMG}
```
fao89, line 53: sweet

fao89 (approved):

PR #318, opened 2023-06-12
Ensure repo-setup strategy is linear

fao89:
After this change, I ran 3 times in a row, and it got all the nodes


fultonj (approved):
Glad it was an easy fix. +2

/lgtm

PR #319, opened 2023-06-12
Create dataplane-operator CRD's for dataplane kuttl tests

slagle:
The CRDs need to be created for the kuttl tests to run.


dprince (approved):

dprince:
Now seeing inconsistent vendoring messages related to go.mod in the CI jobs. Not sure this is the right solution

fao89 (approved):

fao89:
recheck

slagle:
> Now seeing inconsistent vendoring messages related to go.mod in the CI jobs. Not sure this is the right solution

We can go with this one instead: https://github.com/openstack-k8s-operators/install_yamls/pull/322

gen-olm.sh is what is creating the CRDs for the other kuttl tests, and is called by the <service> targets (make neutron). We need to call that earlier in the case of the dataplane.

slagle:
holding in favor of https://github.com/openstack-k8s-operators/install_yamls/pull/322

slagle:
going with https://github.com/openstack-k8s-operators/install_yamls/pull/322 instead

PR #320, opened 2023-06-13
Use `OPENSTACK_RUNNER_IMG` correctly

rabi:
Sets `OPENSTACK_RUNNER_IMG` correctly, so that we can use CI
built runner images.

Also, we don't support NETWORK_ISOLATION=true  for
edpm_compute_baremetal target as it would add an additional
interface to CRC and require reboot. 

Depends-On: https://github.com/openstack-k8s-operators/dataplane-operator/pull/232

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -5,7 +5,7 @@ CRC_DEFAULT_NETWORK_IP ?= 192.168.122.10
 EDPM_COMPUTE_SUFFIX ?= 0
 EDPM_TOTAL_NODES ?= 1
 EDPM_COMPUTE_IP ?= 192.168.122.100
-OPENSTACK_RUNNER_IMG ?= quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
+DATAPLANE_RUNNER_IMG ?= quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
 EDPM_NETWORK_CONFIG_TEMPLATE ?= templates/single_nic_vlans/single_nic_vlans.j2
 EDPM_SSHD_ALLOWED_RANGES ?= ['192.168.122.0/24']
 EDPM_CHRONY_NTP_SERVER ?= pool.ntp.org
@@ -92,10 +92,11 @@ edpm_compute_baremetal: export PROVISIONING_INTERFACE=${BMAAS_NETWORK_NAME}
 edpm_compute_baremetal: export BMAAS_NETWORK_IPADDRESS=172.22.0.3
 edpm_compute_baremetal: export BMAAS_BRIDGE_IPADDRESS=172.22.0.2
 edpm_compute_baremetal: export OPERATOR_NAME=openstack
+edpm_compute_baremetal: export OPENSTACK_RUNNER_IMG=${DATAPLANE_RUNNER_IMG}
 edpm_compute_baremetal: ## Deploy controlplane and dataplane with BMAAS
 	$(eval $(call vars))
 	make bmaas
-	pushd .. && make openstack wait openstack_deploy && popd
+	pushd .. && NETWORK_ISOLATION=false make openstack wait openstack_deploy && popd
 	scripts/gen-ansibleee-ssh-key.sh
 	scripts/edpm-compute-bmaas.sh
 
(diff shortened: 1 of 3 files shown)


rabi:
The [install-yamls-crc-podified-edpm-baremetal](https://review.rdoproject.org/zuul/build/9a6de5d8bbb940ee999b2eedf9d16801) failure is due to some recent changes in edpm-ansible. There is a separate [PR](https://github.com/openstack-k8s-operators/edpm-ansible/pull/190) proposed for that.

raukadah:
recheck

rabi:
recheck

devsetup/Makefile
```diff
@@ -92,10 +92,11 @@ edpm_compute_baremetal: export PROVISIONING_INTERFACE=${BMAAS_NETWORK_NAME}
 edpm_compute_baremetal: export BMAAS_NETWORK_IPADDRESS=172.22.0.3
 edpm_compute_baremetal: export BMAAS_BRIDGE_IPADDRESS=172.22.0.2
 edpm_compute_baremetal: export OPERATOR_NAME=openstack
+edpm_compute_baremetal: export export OPENSTACK_RUNNER_IMG=${DATAPLANE_RUNNER_IMG}
```
rebtoor, line 95: `export` was written twice

devsetup/Makefile
```diff
@@ -5,7 +5,7 @@ CRC_DEFAULT_NETWORK_IP ?= 192.168.122.10
 EDPM_COMPUTE_SUFFIX ?= 0
 EDPM_TOTAL_NODES ?= 1
 EDPM_COMPUTE_IP ?= 192.168.122.100
-OPENSTACK_RUNNER_IMG ?= quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
+DATAPLANE_RUNNER_IMG ?= quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
```
rebtoor, line 8: `OPENSTACK_RUNNER_IMG` is currently being used by `edpm_play` target but I don't know if it should still exist TBH.

scripts/gen-edpm-kustomize.sh
```diff
@@ -199,10 +199,10 @@ cat <<EOF >>kustomization.yaml
       path: /spec/nodes/edpm-compute-${INDEX}/ansibleHost
       value: 192.168.122.$((100+${INDEX}))
     - op: replace
-      path: /spec/nodes/edpm-compute-${INDEX}/hostName
+      path: /spec/roless/edpm-compute-${INDEX}/hostName
```
rebtoor, line 202: there's an extra trailing `s` in `roles`

scripts/gen-edpm-kustomize.sh
```diff
@@ -199,10 +199,10 @@ cat <<EOF >>kustomization.yaml
...
       value: edpm-compute-${INDEX}
     - op: replace
-      path: /spec/nodes/edpm-compute-${INDEX}/openStackAnsibleEERunnerImage
+      path: /spec/roles/edpm-compute/openStackAnsibleEERunnerImage
```
rebtoor, line 205: I spoke with @slagle about this yesterday because i couldn't override runner image using environment variables. It seems that the `openStackAnsibleEERunnerImage` property should be customized under service. @slagle opened https://github.com/openstack-k8s-operators/install_yamls/pull/317 to address this issue.

scripts/gen-edpm-kustomize.sh
```diff
@@ -223,7 +223,7 @@ cat <<EOF >>kustomization.yaml
       path: /spec/nodes/edpm-compute-1/ansibleHost
       value: ${EDPM_COMPUTE_1_IP}
     - op: replace
-      path: /spec/nodes/edpm-compute-1/openStackAnsibleEERunnerImage
+      path: /spec/roles/edpm-compute/openStackAnsibleEERunnerImage
```
rebtoor, line 226: as above

rebtoor (changes requested):

scripts/gen-edpm-kustomize.sh
rabi: We have default services. Also we can't expect users to override all services when we create an image with edpm-ansible image. IMO we should still override it per role along with per service.

scripts/gen-edpm-kustomize.sh
rabi: oops

devsetup/Makefile
rabi: I'll change it.

devsetup/Makefile
rabi: phew... thanks!

scripts/gen-edpm-kustomize.sh
slagle: makes sense that we might need to set it on a role. the behavior (if i'm interpreting https://github.com/openstack-k8s-operators/dataplane-operator/pull/232 correctly) is that the image value on a service would override what is set on the role.

PR #321, opened 2023-06-13
Fix edpm_ovn related variable names

rebtoor:
Just like https://github.com/openstack-k8s-operators/dataplane-operator/pull/233

The variables names for edpm_ovn role were incorrect (`default` vs `DEFAULT`) and this led to neutron.conf file on compute nodes being rendered incorrectly (e.g. empty `transport_url`) so ovn_metadata_agent container kept restarting.

See [here](https://github.com/openstack-k8s-operators/edpm-ansible/blob/6c9c626bc357c325a6201dd8ff0d7db4b8eec46f/roles/edpm_ovn/defaults/main.yml#L94-L113) for correct variable naming.


rebtoor:
recheck

fao89 (approved):

PR #322, opened 2023-06-13
Run dataplane target before dataplane_kuttl_prep

slagle:
dataplane_kuttl_prep needs the dataplane CRDs since it creates the
services under config/services. Moving the dataplane target earlier will
cause gen-olm.sh to run first and create the CRDs.


rabi (approved):
/lgtm

rebtoor (approved):

dprince (approved):

PR #324, opened 2023-06-13
DNM - Updates to edpm scripts

viroel:
This patch update some edpm scripts:
1) in gen-edpm-compute-node.sh, the firstboot script now checks and waits for device wth0 to be in a managed state 
2) in edpm-compute-repos.sh, skip generating a setup repos script if the file was provided by the user


viroel:
@raukadah @slagle this is open for review, can you take a look? Changes are needed for rhel-9 job.

viroel:
/hold

trying a more simple version

viroel:
recheck

viroel:
recheck

devsetup/scripts/edpm-compute-repos.sh
```diff
@@ -31,6 +31,7 @@ if [[ ! -f $SSH_KEY ]]; then
     exit 1
 fi
 
+if [[ ! -f $CMDS_FILE ]]; then
```
raukadah, line 34: The changes looks good. Can we set a default empty value for `CMDS_FILE` at Line 27? Something like that:
```
CMDS_FILE=${CMDS_FILE:-""}
```

slagle: +1, we need to default the var

viroel: @raukadah there is a default value already in line 25: `CMDS_FILE=${CMDS_FILE:-"/tmp/edpm_compute_repos"}`
if we change valut to be empty, it will break the current deployment, since nobody sets this file path. This is a temporary file that is copied inside the compute instance and gets deleted afterwards.

raukadah: Oh sorry I missed that var. thank you for pointing that out.

raukadah (approved):

PR #326, opened 2023-06-14
Avoid silent errors when targets aren't ready yet

fao89:
Some targets depends on the readiness of previous targets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ NAMESPACE                ?= openstack
 PASSWORD                 ?= 12345678
 SECRET                   ?= osp-secret
 OUT                      ?= ${PWD}/out
+TIMEOUT                  ?= 300s
 DBSERVICE           ?= mariadb
 ifeq ($(DBSERVICE), galera)
 DBSERVICE_CONTAINER = openstack-mariadb-0
@@ -448,6 +449,11 @@ openstack: operator_namespace openstack_prep ## installs the operator, also runs
 	$(eval $(call vars,$@,openstack))
 	oc apply -f ${OPERATOR_DIR}
 
+.PHONY: openstack_wait
+openstack_wait: openstack ## waits openstack CSV to succeed.
+	$(eval $(call vars,$@,openstack))
+	timeout $(TIMEOUT) bash -c 'until $$(oc get csv -l operators.coreos.com/openstack-operator.openstack-operators -n ${OPERATOR_NAMESPACE} | grep -q Succeeded); do sleep 1; done'
+
 .PHONY: openstack_cleanup
 openstack_cleanup: operator_namespace## deletes the operator, but does not cleanup the service resources
 	$(eval $(call vars,$@,openstack))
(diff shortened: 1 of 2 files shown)


scripts/gen-edpm-kustomize.sh
```diff
@@ -31,6 +31,30 @@ if [ -z "$DEPLOY_DIR" ]; then
...
+    echo "Please set EDPM_OVN_METADATA_AGENT_SB_CONNECTION"; exit 1
+fi
+
+if [ -z "$EDPM_OVN_DBS" ]; then
+    echo "Please set EDPM_OVN_DBS"; exit 1
+fi
+
+if [ -z "$EDPM_NADS" ]; then
+    echo "Please set EDPM_NADS"; exit 1
+fi
```
fao89, line 56: these variables get empty when `shell oc` errors
```
Error from server (NotFound): services "nova-metadata-internal" not found                                                                                                                     
Error from server (NotFound): secrets "rabbitmq-transport-url-neutron-neutron-transport" not found    
```

scripts/gen-edpm-kustomize.sh
```diff
@@ -31,6 +31,30 @@ if [ -z "$DEPLOY_DIR" ]; then
     echo "Please set DEPLOY_DIR"; exit 1
```
raukadah, line 31: The changes looks good.  Based on the commit message, this file change is related to above commit message.
Can we move Makefile changes to a separate pr?

slagle (commented):
+2.

scripts/gen-edpm-kustomize.sh
fao89: actually, the Makefile changes are related, if you use `make openstack_wait_deploy` it will ensure the variables from `make edpm_deploy` won't be empty

scripts/gen-edpm-kustomize.sh
raukadah: Thank you for clarifying it :-)

scripts/gen-edpm-kustomize.sh
raukadah: Thank you for clarifying it. Let me use these targets in ci-framework side.

Makefile
```diff
@@ -448,6 +448,11 @@ openstack: operator_namespace openstack_prep ## installs the operator, also runs
 	$(eval $(call vars,$@,openstack))
 	oc apply -f ${OPERATOR_DIR}
 
+.PHONY: openstack_wait
+openstack_wait: openstack ## waits openstack CSV to succeed.
+	$(eval $(call vars,$@,openstack))
+	timeout 300s bash -c 'until $$(oc get csv -l operators.coreos.com/openstack-operator.openstack-operators -n ${OPERATOR_NAMESPACE} | grep -q Succeeded); do sleep 1; done'
```
raukadah, line 454: Can we parameterized these timeout? Based on environment it might take more than 300 secs.

cjeanner:
/lgtm

Just a small thing: this means we'll have to update calls in ci-framework: instead of calling the make_openstack, we'll have to call make_openstack_wait (and so on for the other targets).

Wondering why this "wait" isn't in the existing target, and used by default?

fao89:
@cjeanner as I don't know how people were using these targets, I was afraid to break things, so I added these new targets, but I can make it the default

fao89:
recheck

Makefile
```diff
@@ -448,6 +449,11 @@ openstack: operator_namespace openstack_prep ## installs the operator, also runs
 	$(eval $(call vars,$@,openstack))
 	oc apply -f ${OPERATOR_DIR}
 
+.PHONY: openstack_wait
```
raukadah, line 452: We also have `OPERATOR_NAME=openstack [make wait](https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L342)` target to check the operator installation status. Do we want to re-use that here?

Issue #298, opened 2023-05-26
Getting kubectl: command not found error


Jaganathancse:
Getting below issue when running make openstack command using crc,
~/install_yamls/out/operator/baremetal-operator
~/install_yamls/out/operator/baremetal-operator ~/install_yamls/out/operator/baremetal-operator
tools/deploy.sh: line 213: kubectl: command not found
make: *** [Makefile:384: crc_bmo_setup] Error 127

PR #327, opened 2023-06-15
Adding note about necessity of kubectl for install_yamls use

jpodivin:
While it may seem reasonable, there has been some confusion about prerequisites for use of the project #298 .

While we can't be completely sure what the final requirements will be, it can be safely assumed that they are going to involve everything you need for Openshift. Adding a note about that can't hurt and may save us some time in the long run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,12 @@ Aside from generating Yaml and running *oc* commands to apply them to your clust
 
 Helper scripts to automate installing CRC and required tools with versions used in openstack-k8s-operators can be found in [CRC/tools deployment](devsetup/README.md). These scripts/playbook required sudo permissions.
 
+**Note**
+The `install_yamls` project expects several dependencies on the host machine.
+Without them the deployment will fail and you will have install them first.
+In general terms, all tools required by Openshift are also required by `install_yamls`.
+Most importanly, the `kubectl` must be present on the system.
+
 ## Goals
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,6 +6,12 @@ Aside from generating Yaml and running *oc* commands to apply them to your clust
 
 Helper scripts to automate installing CRC and required tools with versions used in openstack-k8s-operators can be found in [CRC/tools deployment](devsetup/README.md). These scripts/playbook required sudo permissions.
 
+**Note**
+The `install_yamls` project expects several dependencies on the host machine.
+Without them the deployment will fail and you will have install them first.
+In general terms, all tools required by Openshift are also required by `install_yamls`.
+Most importanly, the `kubectl` must be present on the system.
```
abays, line 13: You can also symlink `kubectl` to `oc`.  Or at least I've done that in past environments without explicitly installing `kubectl`.

jpodivin: I suppose that could be alternative. But I think we should keep it simple for sake of the reader.

steveb:
I have also proposed a change to install kubectl as one of the developer tools, as currently required to build the baremetal-operator during "make openstack" https://github.com/openstack-k8s-operators/install_yamls/pull/372

dprince (approved):

PR #332, opened 2023-06-16
Only add second node when NODE_COUNT=2

Sandeepyadav93:
By default we create dataplane.yaml for 2 nodes,
With this commit we only add second node when
NODE_COUNT=2.


rabi (approved):
/lgtm

Sandeepyadav93:
recheck

Checking if issue is consistent

raukadah:
recheck

Sandeepyadav93:
recheck

PR #334, opened 2023-06-20
Allow dataplane secret to be updated

fao89:


devsetup/scripts/gen-ansibleee-ssh-key.sh
```diff
@@ -32,14 +32,13 @@ fi
...
-fi
-
 if [ ! -f ${SSH_KEY_FILE} ]; then
+    if oc get secret dataplane-ansible-ssh-private-key-secret -n ${NAMESPACE} 2>&1 1>/dev/null; then
+        echo "Secret dataplane-ansible-ssh-private-key-secret already exists."
+        echo "Delete it first to recreate:"
+        echo "oc delete secret dataplane-ansible-ssh-private-key-secret -n ${NAMESPACE}"
+        exit 0
+    fi
     ssh-keygen -f ${SSH_KEY_FILE} -N "" -t ${SSH_ALGORITHM} -b ${SSH_KEY_SIZE}
```
fultonj, line 42: Perhaps these conditions shouldn't be nested?

cond1: If no key file, then create it. 
cond2: If no secret give warning and fail.

Why make cond2 contingent on cond1? Don't we want cond2 to run regardless of outcome of cond1?

fao89: if you already have the key file, the secret will be created or "updated" using the same key file, which isn't a problem.
But if you already have a secret and somehow you don't have a key file, it means the secret will get updated with a key that not match an existing node

fao89:
retest

devsetup/scripts/gen-ansibleee-ssh-key.sh
```diff
@@ -32,14 +32,13 @@ fi
...
 
-if oc get secret dataplane-ansible-ssh-private-key-secret -n ${NAMESPACE} 2>&1 1>/dev/null; then
-    echo "Secret dataplane-ansible-ssh-private-key-secret already exists."
-    echo "Delete it first to recreate:"
-    echo "oc delete secret dataplane-ansible-ssh-private-key-secret"
-    exit 0
-fi
-
 if [ ! -f ${SSH_KEY_FILE} ]; then
+    if oc get secret dataplane-ansible-ssh-private-key-secret -n ${NAMESPACE} 2>&1 1>/dev/null; then
```
rabi, line 36: We don't need this at all here or where it was added before. [1] would allow you to update the secret if exists. We should just remove all this check and instructions.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/gen-ansibleee-ssh-key.sh#L59-L60

rabi (changes requested):

devsetup/scripts/gen-ansibleee-ssh-key.sh
fao89: I'll close this PR

PR #335, opened 2023-06-20
Adding secret cleanup target

jpodivin:
Patch also exposes the NAMESPACE parameter in the devsetup makefile.

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -10,6 +10,7 @@ EDPM_NETWORK_CONFIG_TEMPLATE ?= templates/single_nic_vlans/single_nic_vlans.j2
 EDPM_SSHD_ALLOWED_RANGES ?= ['192.168.122.0/24']
 EDPM_CHRONY_NTP_SERVER ?= pool.ntp.org
 EDPM_PLAY_CRD ?= edpm/edpm-play.yaml
+NAMESPACE ?= openstack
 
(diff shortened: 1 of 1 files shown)


fultonj (commented):
+2

devsetup/Makefile
```diff
@@ -129,7 +130,7 @@ edpm_compute_repos: ## Configure EDPM compute VM with repositories
 	scripts/edpm-compute-repos.sh ${EDPM_COMPUTE_SUFFIX}
 
 .PHONY: edpm_compute_cleanup
-edpm_compute_cleanup: ## Delete EDPM compute VM
+edpm_compute_cleanup: ansibleee_key_cleanup ## Delete EDPM compute VM
```
fao89, line 133: in a scenario you need to delete just one node it would break, as it would remove the secret for the other nodes

jpodivin: The makefile doesn't really support selective node cleanup at this moment. The only way to do that would be to manually set `EDPM_TOTAL_NODES` != 1 and `EDPM_COMPUTE_SUFFIX` to your chosen nodes index. But that is very unwieldy, and if you were then to remove rest of the nodes, you would have to apply the same procedure again and again.

fao89: `make edpm_compute_cleanup EDPM_COMPUTE_SUFFIX=3` would delete just the edpm-compute-3 node

jpodivin:
irrelevant

PR #340, opened 2023-06-21
WIP: teardown developer deployment

jpodivin:


jpodivin:
recheck

jpodivin:
recheck

devsetup/Makefile
```diff
@@ -316,3 +310,14 @@ bmaas_cleanup:
 	make bmaas_crc_baremetal_bridge_cleanup
 	make bmaas_crc_attach_network_cleanup
 	make bmaas_network_cleanup
+
+.PHONY: ansibleee_cleanup
+ansibleee_cleanup: edpm_play_cleanup
```
slagle, line 315: This has the same name as the ansibleee_cleanup from Makefile in the root of the repo, which does something completely different. I think this needs to be renamed to indicate we are talking about openstackansibleee resources, and not the operator.

devsetup/Makefile
```diff
@@ -316,3 +310,14 @@ bmaas_cleanup:
...
+	oc get openstackansibleees.ansibleee.openstack.org -o name | xargs -r oc delete
+
+.PHONY: operator_cleanup
+operator_cleanup:
+	scripts/openstack-operator-cleanup.sh
```
slagle, line 320: does this really need to be its own script? I'd rather see it inline.

slagle (changes requested):

PR #344, opened 2023-06-25
Use KIND=Swift to kustomize the service spec

rabi:
Along with PR[1] it would set the correct storage class.

[1] https://github.com/openstack-k8s-operators/swift-operator/pull/5


cschwede:
recheck

rabi:
recheck

dprince (approved):

PR #347, opened 2023-06-27
Add target to just deploy dataplane

rabi:
`make edpm_compute_baremetal` deploys both controlplane and dataplane. However, we've a requirement in CI to deploy the operators and then patch some environment variables before deploying rest. `make edpm_baremetal` can be used there. We can probably trash the all-in-one target (edpm_compute_baremetal) later if we want. However, the new target won't work when control plane has been deployed with `NETWORK_ISOLATION=true`.

Also, fixes one BMAAS cleanup issue.


rabi:
recheck

dprince (approved):

PR #350, opened 2023-06-27
Reduce growvols_args /var to 80%

raukadah:
Currently EDPM baremetal jobs are failing with following error during EDPM deployment.
```
dd: error writing '/swap': No space left on device
```
Reducing the growvols_args /var to 80% might fix the issue.

### DIFF
--- a/devsetup/scripts/gen-edpm-bmaas-kustomize.sh
+++ b/devsetup/scripts/gen-edpm-bmaas-kustomize.sh
@@ -136,7 +136,7 @@ fi)
           edpm_chrony_ntp_servers:
             - 0.pool.ntp.org
             - 1.pool.ntp.org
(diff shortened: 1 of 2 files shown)


rabi:
Please add the file to run the job https://github.com/openstack-k8s-operators/install_yamls/blob/main/zuul.d/jobs.yaml#L22

bogdando (commented):
please also adjust dataplane-operator/config/samples/dataplane_v1beta1_openstackdataplane_baremetal.yaml
and edpm-ansible/roles/edpm_growvols/defaults/main.yml

(swap is a hard requirement, so the user facing defaults and examples should also allocate some free space for it)

@SeanMooney fyi

rabi (approved):
/lgtm

PR #354, opened 2023-06-27
Fix pod name used for cleanup targets when using galera

dciabrin:
The name of the galera pods follow a template <dbname>-galera-<n>, and not <dbname>-mariadb-<n>. Fix the typo in the Makefile.


hjensas (approved):

abays (approved):
/lgtm

PR #357, opened 2023-06-28
Fix MON_IP in the Ceph pod for multinode OCP CI

fmount:
The `Ceph` pod is supposed to be single node, single pod, single container. If `MON_IP` is not passed and computed by the `get_pod_ip` function, in a `multinode` context it might return a list of `InternalIP` addresses. This patch represents an attempt to only support a single `IP` address from the list and assign it to the `mon` spawned as first process during the cluster bootstrap.
By doing this we pin the ceph pod on a specific (worker) node using `nodeSelector`.

### DIFF
--- a/scripts/gen-ceph-kustomize.sh
+++ b/scripts/gen-ceph-kustomize.sh
@@ -40,6 +40,7 @@ HOSTNETWORK=${HOSTNETWORK:-true}
 POOLS=("volumes" "images" "backups" "cephfs.cephfs.meta" "cephfs.cephfs.data")
 DAEMONS="osd,mds"
 DATA_SIZE=${DATA_SIZE:-500Mi}
+WORKER=${WORKER:-""}
 
 function add_ceph_pod {
 cat <<EOF >ceph-pod.yaml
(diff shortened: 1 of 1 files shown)


fultonj (commented):
+2

Sandeepyadav93:
recheck

Sandeepyadav93:
/lgtm

thank you, Tested locally on multinode OCP environment, it works.

`make ceph` target is only for dev/ci usecase, so I think its okay to pin ceph pod to a particular worker.

scripts/gen-ceph-kustomize.sh
```diff
@@ -119,7 +122,12 @@ function get_pod_ip {
         echo "HOSTNETWORK must be set to true!"
         exit 1
     fi
-    MON_IP=$(oc get nodes -o 'jsonpath={.items[*].status.addresses[?(.type=="InternalIP")].address}')
+    NODES=$(oc get nodes -o 'jsonpath={.items[*].status.addresses[*].address}')
```
Sandeepyadav93, line 125: @fmount I think we should only filter worker nodes

oc get nodes --selector='node-role.kubernetes.io/worker='

fmount: +1, doing that in my next update.

fmount: Done

Sandeepyadav93:
/lgtm

thanks for the update

PR #358, opened 2023-06-29
Paramterize EDPM standalone VM creation

bogdando:
Add variables for vcpu, ram and disk.
Add control flag to deploy ceph.


devsetup/Makefile
```diff
@@ -4,6 +4,10 @@ PULL_SECRET  ?= ${PWD}/pull-secret.txt
 CRC_DEFAULT_NETWORK_IP ?= 192.168.122.10
 EDPM_COMPUTE_SUFFIX ?= 0
 EDPM_TOTAL_NODES ?= 1
+EDPM_COMPUTE_VCPUS ?= 8
+EDPM_COMPUTE_RAM ?= 20
+EDPM_COMPUTE_DISK_SIZE ?= 70
```
fao89, line 9: it would change the default for edpm nodes: https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/gen-edpm-compute-node.sh#L23-L30
I believe this might break CI as it is requiring more resources

devsetup/scripts/standalone.sh
```diff
@@ -51,12 +55,13 @@ sudo hostnamectl set-hostname standalone.localdomain --transient
 export NTP_SERVER=${NTP_SERVER:-"clock.corp.redhat.com"}
 
 /tmp/network.sh
-/tmp/ceph.sh
+[[ "$EDPM_COMPUTE_CEPH_ENABLED" == "true" ]] && /tmp/ceph.sh
```
fao89, line 58: 👍🏽

devsetup/standalone/openstack.sh
```diff
@@ -56,15 +57,19 @@ parameter_defaults:
   ValidateControllersIcmp: false
 EOF
 
+if [ "$EDPM_COMPUTE_CEPH_ENABLED" = "true" ] ; then
+  CEPH_ARGS= "-e ~/deployed_ceph.yaml -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm-rbd-only.yaml"
+else
+	CEPH_ARGS=""
+fi
```
fao89, line 64: sweet!

fao89:
recheck

devsetup/Makefile
bogdando: I'm not sure if it still would change those, looks the same now?

bogdando:
PTAL @fao89

bogdando:
PTAL

devsetup/Makefile
```diff
@@ -193,6 +197,10 @@ edpm_deploy_instance: ## Spin a instance on edpm node
 	-oc rsh openstackclient bash /tmp/edpm-deploy-instance.sh
 
 .PHONY: standalone
+standalone: export EDPM_COMPUTE_VCPUS=${EDPM_COMPUTE_VCPUS}
+standalone: export EDPM_COMPUTE_RAM=${EDPM_COMPUTE_RAM}
+standalone: export EDPM_COMPUTE_DISK_SIZE=${EDPM_COMPUTE_DISK_SIZE}
```
bogdando, line 202: akefile:202: *** Recursive variable 'EDPM_COMPUTE_DISK_SIZE' references itself (eventually).  Stop.

devsetup/Makefile
bogdando: Does your concern still apply?

fao89: yes, it does!
We have different defaults for EDPM and adoption workflows:
https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/gen-edpm-compute-node.sh#L28-L36

fao89: for example, you are making EDPM_COMPUTE_DISK_SIZE=70 for both adoption and EDPM dev env,
but the EDPM dev env default is currently 20.
Basically dropping these lines would fix it, as the defaults are set here already: https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/gen-edpm-compute-node.sh#L28-L36

bogdando: thanks!
I guess, that's done now?

fao89 (approved):

PR #359, opened 2023-06-29
Fallback to cold snapshot of standalone VM

bogdando:
When the snapshot cannot be created live (error: Requested operation is not valid: migration with virtiofs device is not supported), fallback to the cold snapshotting method.

Also warn about the need to sync system time after revert is done.

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -192,11 +192,18 @@ standalone_snapshot: ## Create standalone snapshot
 	virsh --connect=qemu:///system detach-device-alias edpm-compute-${EDPM_COMPUTE_SUFFIX} fs0 --live || true
 	sleep 3
 	virsh --connect=qemu:///system snapshot-create-as --atomic --domain edpm-compute-${EDPM_COMPUTE_SUFFIX} --name clean
+	if [ $? -ne 0 ] ; then
+		virsh --connect=qemu:///system suspend edpm-compute-${EDPM_COMPUTE_SUFFIX}
+		virsh --connect=qemu:///system destroy edpm-compute-${EDPM_COMPUTE_SUFFIX}
+		virsh --connect=qemu:///system snapshot-create-as --atomic --domain edpm-compute-${EDPM_COMPUTE_SUFFIX} --name clean
+	fi
 
 .PHONY: standalone_revert
 standalone_revert: ## Revert standalone snapshot
 	$(eval $(call vars))
 	virsh --connect=qemu:///system snapshot-revert --domain edpm-compute-${EDPM_COMPUTE_SUFFIX} --snapshotname clean
+	virsh --connect=qemu:///system resume --domain edpm-compute-${EDPM_COMPUTE_SUFFIX} ||:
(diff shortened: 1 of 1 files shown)


devsetup/Makefile
```diff
@@ -176,11 +176,18 @@ standalone_snapshot: ## Create standalone snapshot
...
+		virsh --connect=qemu:///system destroy edpm-compute-${EDPM_COMPUTE_SUFFIX}
+		virsh --connect=qemu:///system snapshot-create-as --atomic --domain edpm-compute-${EDPM_COMPUTE_SUFFIX} --name clean
+	fi
 
 .PHONY: standalone_revert
 standalone_revert: ## Revert standalone snapshot
 	$(eval $(call vars))
 	virsh --connect=qemu:///system snapshot-revert --domain edpm-compute-${EDPM_COMPUTE_SUFFIX} --snapshotname clean
+	virsh --connect=qemu:///system resume --domain edpm-compute-${EDPM_COMPUTE_SUFFIX} ||:
+	$(warning Make sure the system time is synchronized for the reverted VM!)
```
rebtoor, line 190: I don't get it this, especially after the double pipe, care to explain?

rebtoor (commented):
Asked info about lines 189/190

devsetup/Makefile
bogdando: that meant to ingore the resume command failures, as if we have creted a live snapshot, this command will fail to resume, but if we created the cold one, it needs to be resumed. Just a shortcut to suppress failures, but we can wrap it with the similar 'if else' section as above

devsetup/Makefile
```diff
@@ -176,11 +176,18 @@ standalone_snapshot: ## Create standalone snapshot
 	$(eval $(call vars))
 	virsh --connect=qemu:///system detach-device-alias edpm-compute-${EDPM_COMPUTE_SUFFIX} fs0 --live
```
jistr, line 177: This line should take care of removing the virtiofs device and allow the live snapshot. At least it worked for me... :)

booxter: Is it because we need a delay between the two to give libvirt space to detach? Is it async?

jistr: Yes this just came up at the hackathon. I would rather not stop the standalone VM as it might not come back up 100% well, but a simple e.g. 3-5 second wait between the virtiofs detach and the snapshot should do it. The snapshot is being done so infrequently that it shouldn't matter it takes a little longer.

jistr: (Sean mentioned the detach operation is indeed async.)

bogdando: I think this change still makes sense, as a fallback option to have at least the cold snapshot created, if taking a 3 seconds snap didn't trick the virtiofs monster

bogdando:
PTAL

devsetup/Makefile
bogdando: makes sense, this indeed looks no longer needed

PR #362, opened 2023-06-30
Increase beremetal disk size to 20G

rabi:
This will align it with the pre-provisioned disk size. We should not have issues as we've reduced the default node count from 2 to 1.


rebtoor (approved):

rebtoor:
recheck

abays (approved):
/lgtm

PR #338, opened 2023-06-20
Deploy Memcached instance in kuttl tests

kajinamit:
We are adding support for caching using memcached globally in all OpenStack services. This is the prep work for it and ensures the CR instance is created in kuttl tests.

This also refactors the entrypoints for all kuttl tests to make all implementations more consistent and simple.

PR #363, opened 2023-07-01
Run openstack_crds for kuttl targets

rabi:
It's possible that when we run the targets to install  operators, it would take time to create the CRDs. We don't wait for the operators to be ready before proceeding with creating the CRs as part of kuttl tests and they fail[2] all the time. `make openstack_crds` ensures that all crds are there before any kuttl job run and we should keep that.

This partially reverts PR[1].

[1] https://github.com/openstack-k8s-operators/install_yamls/pull/338
[2] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_openstack-ansibleee-operator/178/pull-ci-openstack-k8s-operators-openstack-ansibleee-operator-main-ansibleee-operator-build-deploy-kuttl/1675012736523702272/artifacts/ansibleee-operator-build-deploy-kuttl/openstack-k8s-operators-kuttl/build-log.txt

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1132,7 +1132,7 @@ mariadb_kuttl_run: ## runs kuttl tests for the mariadb operator, assumes that ev
 	kubectl-kuttl test --config ${MARIADB_KUTTL_CONF} ${MARIADB_KUTTL_DIR}
 
 .PHONY: mariadb_kuttl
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

PR #365, opened 2023-07-03
Fix validation of MariaDB/Galera

kajinamit:
Currently we always assert MariaDB in validate_mariadb target but this does not work in case Galera is used. This ensures the correct file is used in assertion according to the database type.


abays (approved):
/lgtm

PR #366, opened 2023-07-03
Ignore delete failure for openstack_deploy_cleanup

rabi:
It's possible that the CRDs don't exist and it should just ignore that.


kajinamit:
/lgtm

This resolves the failure in the clean env without OpenStackControlPlane CRD installed.

abays (approved):
/lgtm

PR #367, opened 2023-07-03
Wait for cert-manager webhook pod to show up

rabi:
`crc_bmo_setup` fails many times as the cert-manager pods don't show up in time when we want check for them and it fails with "error: no matching resources found". Noticed frequently, [1] is an example.

[1] https://logserver.rdoproject.org/62/262/2472d4ec91b0b9123f3ec5bde33a97c5b80d7ca8/github-check/dataplane-operator-crc-podified-edpm-baremetal/551a9e6/ci-framework-data/logs/ci_make_001_run_openstack.log


abays:
recheck

rabi:
recheck

abays (approved):
/lgtm

PR #368, opened 2023-07-04
Suppress failure in database drop

kajinamit:
Currently we see a lot of error messages caused by non-existing database, but these are all ignored.

```
ERROR 1008 (HY000) at line 1: Can't drop database 'neutron'; database doesn't exist
command terminated with exit code 1
```

We can silence these messages by 'if exists' statement supported by mariadb.

Note that we have to handle the case where mariadb is not yet deployed so the fallback using the true command is still kept.


abays (approved):
/lgtm

PR #369, opened 2023-07-04
Make PV cleanup more robust

kajinamit:
This adds a new label to all PVS created by the crc_storage target. This allows us to query all PVs later during cleanup. This also ensures that all PVCs associated with the PVs created by crc_storage target so that we can remove all PVs.

### DIFF
--- a/scripts/cleanup-crc-pv.sh
+++ b/scripts/cleanup-crc-pv.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -ex
+
+# First, remove all PVCs still bound. Some operators (eg mariadb-operator) do
+# not remove pvc after removing deployments
+for pvc in `oc get pv --selector provisioned-by=crc-devsetup --no-headers | grep Bound | awk '{print $6}'`; do
+    NS=`echo $pvc | cut -d '/' -f 1`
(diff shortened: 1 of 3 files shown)


scripts/cleanup-crc-pv.sh
```diff
@@ -0,0 +1,13 @@
+#/bin/bash
+#
+for pvc in `oc get pv --selector provisioned-by=crc-devsetup --no-headers | grep Bound | awk '{print $6}'`
```
dciabrin, line 3: On my env the awk needs $5 to return the name, not sure what changes.

dciabrin (changes requested):

dciabrin (approved):

scripts/cleanup-crc-pv.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
Sandeepyadav93, line 1: i think it will be worth adding `set -ex` here

kajinamit: Yes. That's better. Added it in the latest version.

stuggi (approved):
/lgtm

PR #370, opened 2023-07-04
add devsetup dependency for gcc in download_tools

davidjpeacock:
GCC was missing as a dependency, needed at least for base Fedora deployments.

Additionally, README referenced stale title for devsetup subdirectory.


davidjpeacock:
recheck

fultonj (approved):

PR #373, opened 2023-07-05
Switch Pre-provsioned edpm to use IPAM

rabi:
This switches pre-provisioned edpm deployment to use IPAM.

Depends-On: https://github.com/openstack-k8s-operators/dataplane-operator/pull/270

### DIFF
--- a/scripts/gen-edpm-kustomize.sh
+++ b/scripts/gen-edpm-kustomize.sh
@@ -85,23 +85,24 @@ patches:
       path: /spec/nodes/edpm-compute-0/node/ansibleVars
       value: |
         ctlplane_ip: ${EDPM_COMPUTE_IP}
-        internal_api_ip: 172.17.0.100
-        storage_ip: 172.18.0.100
-        tenant_ip: 172.19.0.100
-        fqdn_internal_api: '{{ ansible_fqdn }}'
-    - op: replace
-      path: /spec/nodes/edpm-compute-0/node/ansibleSSHPrivateKeySecret
-      value: ${EDPM_ANSIBLE_SECRET}
     - op: add
       path: /spec/roles/edpm-compute/services/0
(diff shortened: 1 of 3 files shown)


scripts/gen-edpm-kustomize.sh
```diff
@@ -85,23 +85,25 @@ patches:
       path: /spec/nodes/edpm-compute-0/node/ansibleVars
       value: |
         ctlplane_ip: ${EDPM_COMPUTE_IP}
-        internal_api_ip: 172.17.0.100
-        storage_ip: 172.18.0.100
-        tenant_ip: 172.19.0.100
-        fqdn_internal_api: '{{ ansible_fqdn }}'
-    - op: replace
-      path: /spec/nodes/edpm-compute-0/node/ansibleSSHPrivateKeySecret
-      value: ${EDPM_ANSIBLE_SECRET}
```
fao89, line 94: why removing the SSH secret?

fao89 (commented):
should we update https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane.yaml ?

scripts/gen-edpm-kustomize.sh
rabi: We don't need to duplicate for the nodes when provided for the https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-edpm-kustomize.sh#L193

fao89 (approved):

rabi:
recheck

rabi:
recheck

stuggi (commented):
+1 from me for changing the IPs for the nncp here. nice to see we can get rid of all this "hard" coded edpm config.

PR #374, opened 2023-07-05
Workaround mariadb 10.5.16 bug with 'drop tables'

dciabrin:
'drop table' statements can dead lock under specific conditions linked to query cache. The fix is apparently available in 10.5.19 [1], so until it gets consumed, apply a workaround in cleanup targets

[1] https://jira.mariadb.org/browse/MDEV-29760


stuggi (approved):
/lgtm

PR #376, opened 2023-07-05
fix baremetal-operator repo existence causing abrupt failure

davidjpeacock:
If baremetal-operator repo is already cloned, for example from a previous deployment attempt, due to its presence and non-empty status the git clone will fail, and the rest of the make is aborted abruptly.

This patch does a pre-emptive rm of the existing clone if there is one and will silently do nothing if it does not exist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -441,7 +441,7 @@ crc_bmo_setup:
 	oc apply -f ${CERTMANAGER_URL}
 	timeout ${TIMEOUT} bash -c 'until [ "$$(oc get pod -l app=webhook -n cert-manager -o name)" != "" ]; do sleep 1; done'
 	oc wait pod -n cert-manager --for condition=Ready -l app=webhook --timeout=$(TIMEOUT)
-	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(BMO_BRANCH),-b ${BMO_BRANCH}) ${BMO_REPO} "baremetal-operator" && popd
+	pushd ${OPERATOR_BASE_DIR} && rm -rf ${BMO_REPO} && git clone ${GIT_CLONE_OPTS} $(if $(BMO_BRANCH),-b ${BMO_BRANCH}) ${BMO_REPO} "baremetal-operator" && popd
 	pushd ${OPERATOR_BASE_DIR}/baremetal-operator && sed -i 's/eth2/${PROVISIONING_INTERFACE}/g' ironic-deployment/default/ironic_bmo_configmap.env && popd
 	pushd ${OPERATOR_BASE_DIR}/baremetal-operator && sed -i 's/eth2/${PROVISIONING_INTERFACE}/g' config/default/ironic.env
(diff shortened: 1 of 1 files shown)


abays:
This directory is normally removed via `make crc_bmo_cleanup` directly or transitively via `make openstack_cleanup`.  I've added @rabi for review to get his opinion.

davidjpeacock:
> This directory is normally removed via `make crc_bmo_cleanup` directly or transitively via `make openstack_cleanup`. I've added @rabi for review to get his opinion.

Thank you for the insight.  I found it to be the only non-self healing element of the deployment, so figured it can't hurt to have an implicit clean up.

rabi:
> This directory is normally removed via `make crc_bmo_cleanup` directly or transitively via `make openstack_cleanup`. I've added @rabi for review to get his opinion.

The issue mentioned can happen in only 2 scenarios.

1.  User runs `make openstack` multiple times without running `make openstack cleanup`
2.  You cleanup crc and re-deploy  after running `make openstack` and re-run `make openstack`

Whenever faced that issue, it's better to run 'make openstack_cleanup` .

If we really want to add something for user error, then we shoud add pre target `crc_bmo_cleanup` @ https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L438

I was initially worried that it would cleanup bmo from non-crc environments, but I think it's fine as we're using a different namspace `baremetal-operator-system` with crc.

davidjpeacock:
Thanks Rabi for the insight, I appreciate it.

> > This directory is normally removed via `make crc_bmo_cleanup` directly or transitively via `make openstack_cleanup`. I've added @rabi for review to get his opinion.
> 
> The issue mentioned can happen in only 2 scenarios.
> 
>     1. User runs `make openstack` multiple times without running `make openstack cleanup`

I'm kinda torn on this being a user error.  You're right of course at face value that a clean up should happen after a botched make build, but being able to continue a make with a minor adjustment is idiomatic, that is to say, normal and expected behaviour.  For example currently there is an issue with a lack of kubectl which will force an abort.  It's normal and expected after fixing a package missing like that to be able to re-run your make without having to do a clean up first.

I doubt I'll be the only one to encounter scenarios such as this, and when the fix I suggest is harmless, I don't see a problem with the way I propose avoiding this situation.

> Whenever faced that issue, it's better to run `make openstack_cleanup` .

You are not wrong at all.  I agree that clean up is healthy.

> If we really want to add something for user error, then we shoud add pre target `crc_bmo_cleanup` @ https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L438

Okie dokie, so instead of my suggested fix are you happy for me to work with working on your suggestion?  Ultimately I just want to avoid the user having to run a manual clean up step in minor "trip up" situations as described above.

> I was initially worried that it would cleanup bmo from non-crc environments, but I think it's fine as we're using a different namspace `baremetal-operator-system` with crc.

Yeah that was my thinking too.

steveb (approved):

rabi:
Proposed https://github.com/openstack-k8s-operators/install_yamls/pull/454 instead.

abays:
This should be fixed now by https://github.com/openstack-k8s-operators/install_yamls/pull/454

PR #372, opened 2023-07-05
Install kubectl

steveb:
This is required by the build script of metal3 baremetal-operator, which is run during "make openstack".

PR #377, opened 2023-07-05
Install kubectl from download_tools

slagle:
The baremetal-operator[1] requires kubectl to install, but kubectl was
never installed by install_yamls. This adds installing kubectl to the
download_tools ansible role. The kubectl version defaults to the version
of kubernetes used by crc.

[1] https://github.com/metal3-io/baremetal-operator/blob/main/tools/deploy.sh


abays:
Looks like this may be a duplicate of https://github.com/openstack-k8s-operators/install_yamls/pull/372

steveb:
Looks pretty similar yeah :)

slagle:
> Looks like this may be a duplicate of #372

Heh, I should have checked first. I'll close this one.

PR #379, opened 2023-07-06
DB cleanup workaround for galera

dciabrin:
This is a second attempt at working around a galera issue with 'drop database' statement with mariadb 10.5.16:

  . the semantics of write set replication requires us to use
    'drop table if exists' otherwise the DML statement will be
    interpreted as a failed replication when the database does
    not exists, which creates spurious files in the mariadb
    directory.

  . in order to avoid the drop thread from deadlocking, resetting
    the query cache is not enough for mariadb 10.5.16. Flushing
    tables seems to be enough to workaround all the deadlocks
    issues for now.


abays (approved):
/lgtm

abays:
recheck

PR #380, opened 2023-07-07
Check operator status to ensure operator is installed

kajinamit:
... instead of checking deployment status. This is more robust way to ensure all resources including CRDs are installed.

With this change we'll remove openstack_crds target from kuttl targets. The problem with using openstack_crds in kuttl targets is that the CRDs in the openstack operator bundle image can contain old CRD which does not match the actual CRD version expected by the operator image we are deploying.

### DIFF
--- a/scripts/get-operator-status.sh
+++ b/scripts/get-operator-status.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+set -x
+
+if [ -z "${OPERATOR_NAMESPACE}" ]; then
+    echo "Please set OPERATOR_NAMESPACE"; exit 1
+fi
+
+if [ -z "$OPERATOR_NAME" ]; then
+    echo "Please set OPERATOR_NAME"; exit 1
+fi
+
+CSVNAME=$(oc get csv -n ${OPERATOR_NAMESPACE} -o jsonpath='{range .items[*]}{@.metadata.name}{"\n"}{end}' | egrep -e "^${OPERATOR_NAME}-operator\.v")
+if [ -z "$CSVNAME" ]; then
+    echo "NOTFOUND"
+    exit 1
+fi
+
+PHASE=$(oc get -n ${OPERATOR_NAMESPACE} csv/${CSVNAME} -o jsonpath='{.status.phase}')
+echo $PHASE
+if [ "$PHASE" != "Succeeded" ]; then
+    exit 1
+fi
+exit 0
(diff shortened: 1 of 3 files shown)


scripts/get-operator-status.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if [ -z "${OPERATOR_NAMESPACE}" ]; then
+    echo "Please set OPERATOR_NAMESPACE"; exit 1
+fi
+
+if [ -z "$OPERATOR_NAME" ]; then
+    echo "Please set OPERATOR_NAME"; exit 1
+fi
+
+CSVNAME=$(oc get csv -n ${OPERATOR_NAMESPACE} -o jsonpath='{range .items[*]}{@.metadata.name}{"\n"}{end}' | grep $OPERATOR_NAME)
```
rabi, line 12: Just checking for a single operator would not be enough I guess. dataplane-operator also installs ansibleee-operator, so just waiting for dataplane would not be sufficient.  So we need to keep all dependencies and check for them.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L1364

kajinamit: I've updated the change to call the `make wait` target after operator installation. With this we ensure all operators installed via dependencies.

kajinamit:
recheck

looks like webhook was not up when the resources were being patched.

kajinamit:
recheck

scripts/get-operator-status.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if [ -z "$OPERATOR_NAME" ]; then
+    echo "Please set OPERATOR_NAME"; exit 1
+fi
+
+CSVNAME=$(oc get csv -n ${OPERATOR_NAMESPACE} -o jsonpath='{range .items[*]}{@.metadata.name}{"\n"}{end}' | egrep -e "^${OPERATOR_NAME}-operator\.v")
+if [ -z "$CSVNAME" ]; then
+    echo "NOTFOUND"
+    exit 1
+fi
```
abays, line 16: This `egrep -e "^${OPERATOR_NAME}-operator\.v"` doesn't seem to be working in my local env for Rabbit, because the operator CSV name is actually `rabbitmq-cluster-operator`

abays:
recheck

abays (approved):
/lgtm

PR #384, opened 2023-07-09
Check operator status less frequently

kajinamit:
Installation of operators rarely completes within a few seconds and it usually takes ~ minute order. Let's reduce the frequency to avoid too much frequent query with little gain.


bshephar (approved):

abays (approved):
/lgtm

PR #385, opened 2023-07-10
Fix name of RabbitMQ CR instance

kajinamit:
Commit 77be8e1f133cb06bd020c297ab8cb03bc4a95254 removed the existing override and now the CR is deployed with a different name defined in the repository.

Let's re-introduce the name override to avoid updating the cluster name we have defined in each sample file.


bshephar (approved):
Seems fine to me for resolving the issue described.

kajinamit:
recheck

scripts/gen-service-kustomize.sh
```diff
@@ -83,6 +83,15 @@ EOF
     fi
 done
 
+if [ -n "$CR_NAME" ]; then
+    cat <<EOF >>kustomization.yaml
+  patch: |-
```
karelyatin, line 88: patch: line extra

karelyatin (changes requested):

scripts/gen-service-kustomize.sh
kajinamit: Oh, yes. Good catch. I've fixed it

scripts/gen-service-kustomize.sh
```diff
@@ -38,8 +38,6 @@ fi
 IMAGE=${IMAGE:-unused}
 IMAGE_PATH=${IMAGE_PATH:-containerImage}
 
-NAME=${KIND,,}
```
kajinamit, line 41: I noticed this is not used, then we can probably use NAME instead of CR_NAME.

karelyatin (commented):
/lgtm

karelyatin (approved):

PR #387, opened 2023-07-10
Allow customizing container image for heat-api-cfn

kajinamit:
... because we already support customizing container image for heat-api and heat-engine.


abays (approved):
/lgtm

PR #393, opened 2023-07-11
Allow to set MTU for networks

cjeanner:
It may happen we can't configure the underlying network with an MTU
higher to 1500. In such case, network isolation will not be possible.

Allowing to configure isolated network MTU should allow to deploy on any
kind of environment, and actually be closer to actual production use
where customer can't or don't want to enable jumbo frames.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ STORAGE_CLASS       ?= "local-storage"
 
 # network isolation
 NETWORK_ISOLATION   ?= true
+NETWORK_MTU         ?= 1500
 
 # options to pass in all targets that use git clone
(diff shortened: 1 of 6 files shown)


cjeanner:
recheck

fultonj (commented):
+2 
/lgtm
Thank you. We should set the MTU for the HCI usecase so this should be handy.

https://github.com/openstack-k8s-operators/docs/blob/main/hci.md#mtu-settings-for-ceph

cjeanner:
recheck

cjeanner:
(rebased against `main` - I think I saw some changes related to galera lately)

PR #398, opened 2023-07-11
Add variables and functions for swift kuttl

jparoly:


abays:
Maybe we should hold this until there are actual KUTTL tests in the Swift operator?  I don't see any at the moment.

jparoly:
> Maybe we should hold this until there are actual KUTTL tests in the Swift operator? I don't see any at the moment.

jparoly:
Here is the PR where I am adding the swift kuttl tests: https://github.com/openstack-k8s-operators/swift-operator/pull/41

Makefile
```diff
@@ -1487,6 +1491,20 @@ manila_kuttl: kuttl_common_prep ceph manila manila_deploy_prep ## runs kuttl tes
 	make kuttl_common_cleanup
 	make cleanup
 
+.PHONY: swift_kuttl_run
+swift_kuttl_run: ## runs kuttl tests for the swift operator, assumes that everything needed for running the test was deployed beforehand.
+	kubectl-kuttl test --config ${SWIFT_KUTTL_CONF} ${SWIFT_KUTTL_DIR}
```
cescgina, line 1496: I think it would be a good idea to add here `--namespace ${NAMESPACE}` so that the tests can be run in a different namespace easily

abays: I agree.  Since we already have `export NAMESPACE = ${SWIFT_KUTTL_NAMESPACE}` on L1499, it seems like we're just missing `--namespace ${NAMESPACE}` here.

jparoly: done, thank you both

cescgina (commented):
lgtm, thanks for the changes

Makefile
```diff
@@ -343,6 +343,10 @@ SWIFT_REPO       ?= https://github.com/openstack-k8s-operators/swift-operator.gi
 SWIFT_BRANCH     ?= main
 SWIFT            ?= config/samples/swift_v1beta1_swift.yaml
 SWIFT_CR         ?= ${OPERATOR_BASE_DIR}/swift-operator/${SWIFT}
+SWIFTAPI_DEPL_IMG  ?= unused
```
abays, line 346: Is this used anywhere?

jparoly: no its not, do we need it? hmmm

jparoly: I guess we don't need images for swift, but other do require images?  Is that the difference?  If so, I will remove this line.

abays: We can always add it later if we need it for something

jparoly:
recheck

jparoly:
recheck

jparoly:
recheck

abays:
Each Zuul job has passed independently across a few runs.

abays (approved):
/lgtm

PR #399, opened 2023-07-12
Update infra-operator kuttl tests set up to extend test coverage

kajinamit:
With this PR, now we deploy KeystoneAPI before we run kuttl tests for infra-operator. This is required so that we can test OpenStackClient CR which depends on KeystoneAPI.

Also, this adds the step to ensure the Memcached CR deployed via `make memcached_deploy` has the expected name, so that we can change the name in the sample config file. This is needed so that we can deploy two instances in infra-operator kuttl tests. One is used by KeystoneAPI CR while the other is created for tests.


kajinamit:
recheck

kajinamit:
recheck

stuggi (approved):
/lgtm

PR #401, opened 2023-07-13
Pin to crc-2.19.0

steveb:
Containers such as cinder-api and ironic-db-sync fail with Permission denied when opening /dev/stdout, see the Jira for details.

This is still occuring with crc 2.23.0, this change pins to the last working crc release.

Jira: [OSP-25627](https://issues.redhat.com//browse/OSP-25627)


dprince:
Seen at least 2 requests now  due to potential issues with CRC. I'd be okay landing this and perhaps filing Jira's to follow up on the issues we've seen thus far.

dprince (approved):

booxter:
/lgtm

Fixed the problem in my local data-plane-adoption env. Thanks!

To document all issues I had with the newer image (in case someone needs a checklist before it's unpinned):

1. cinder-api failed with permission denied error.
2. pods now don't have networks-status annotation (only network-status is present) due to https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/pull/45
3. (not sure if definitely because of the new image) mariadb-client restoring database from backups was timing out quite often.

I hope it helps.

karelyatin (approved):
/lgtm

PR #403, opened 2023-07-13
Add edpm_baremetal_deploy

rabi:
This target can be used when you've BMHs already available to be used and has no pre-req of using crc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -272,13 +272,16 @@ ANSIBLEEE_KUTTL_DIR  ?= ${OPERATOR_BASE_DIR}/openstack-ansibleee-operator/tests/
 BAREMETAL_IMG       ?= quay.io/openstack-k8s-operators/openstack-baremetal-operator-index:latest
 BAREMETAL_REPO      ?= https://github.com/openstack-k8s-operators/openstack-baremetal-operator.git
 BAREMETAL_BRANCH    ?= main
+BMH_NAMESPACE       ?= ${NAMESPACE}
 
 # Dataplane Operator
 DATAPLANE_IMG                                    ?= quay.io/openstack-k8s-operators/dataplane-operator-index:latest
 DATAPLANE_REPO                                   ?= https://github.com/openstack-k8s-operators/dataplane-operator.git
 DATAPLANE_BRANCH                                 ?= main
 OPENSTACK_DATAPLANE                              ?= config/samples/dataplane_v1beta1_openstackdataplane.yaml
+OPENSTACK_DATAPLANE_BAREMETAL                    ?= config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml
 DATAPLANE_CR                                     ?= ${OPERATOR_BASE_DIR}/dataplane-operator/${OPENSTACK_DATAPLANE}
+DATAPLANE_BAREMETAL_CR                           ?= ${OPERATOR_BASE_DIR}/dataplane-operator/${OPENSTACK_DATAPLANE_BAREMETAL}
 DATAPLANE_ANSIBLE_SECRET                         ?=dataplane-ansible-ssh-private-key-secret
 DATAPLANE_ANSIBLE_USER                           ?=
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -575,6 +577,37 @@ edpm_deploy: input edpm_deploy_prep ## installs the dataplane instance using kus
 	$(eval $(call vars,$@,dataplane))
 	oc kustomize ${DEPLOY_DIR} | oc apply -f -
 
+.PHONY: edpm_deploy_baremetal_prep
+edpm_deploy_baremetal_prep: export KIND=OpenStackDataPlane
+edpm_deploy_baremetal_prep: export EDPM_ANSIBLE_SECRET=${DATAPLANE_ANSIBLE_SECRET}
+edpm_deploy_baremetal_prep: export EDPM_ANSIBLE_USER=cloud-amin
+edpm_deploy_baremetal_prep: export BMH_NAMESPACE=openstack
```
abays, line 584: The `openstack` here needs to be replaced with something dynamic, I think

Makefile
```diff
@@ -575,6 +577,37 @@ edpm_deploy: input edpm_deploy_prep ## installs the dataplane instance using kus
 	$(eval $(call vars,$@,dataplane))
 	oc kustomize ${DEPLOY_DIR} | oc apply -f -
 
+.PHONY: edpm_deploy_baremetal_prep
+edpm_deploy_baremetal_prep: export KIND=OpenStackDataPlane
+edpm_deploy_baremetal_prep: export EDPM_ANSIBLE_SECRET=${DATAPLANE_ANSIBLE_SECRET}
+edpm_deploy_baremetal_prep: export EDPM_ANSIBLE_USER=cloud-amin
```
abays, line 583: `s/cloud-amin/cloud-admin`

rabi: done

Makefile
rabi: yeah we can use ${NAMESPACE}, but it's something that has to be provided in most cases.

abays: Would that work if you wanted to deploy the `OSDP` in `openstack` namespace, but have BMHs in `openshift-machine-api`?  That's the dilemma we face against an AI/IPI OCP cluster.  But maybe I could have created my BMHs in `openstack` instead of `openshift-machine-api`.  It just used to be that you had to put them in the same namespace as the Baremetal operator, which happens to be `openshift-machine-api` in AI/IPI installations.

rabi: the default is `openshift-machine-api`, If your BMH is in `openshift-machine-api`  then you've to set `$BMH_NAMESPACE=openshift-machine-api` with the way we've written the kustomization for this target.

scripts/gen-edpm-baremetal-kustomize.sh
```diff
@@ -0,0 +1,201 @@
...
+    - op: replace
+      path: /spec/roles/edpm-compute/nodeTemplate/ansibleSSHPrivateKeySecret
+      value: ${EDPM_ANSIBLE_SECRET}
+EOF
+if [ "$EDPM_SINGLE_NODE" == "true" ]; then
+cat <<EOF >>kustomization.yaml
+    - op: remove
+      path: /spec/nodes/edpm-compute-1
+EOF
+elif [ "$EDPM_TOTAL_NODES" -gt 2 ]; then
```
abays, line 175: Should this be `-gt 1`?  Otherwise it seems you need at least 3 for `EDPM_TOTAL_NODES` to have multiple nodes created.

rabi: I think there is a `DATAPLANE_SINGLE_NODE` var which has to be changed to `false` from the default `true`. I just copied it from what they've for pre-provisoned, though it could have been managed with one var. We delete one node L172-173 from the sample if  `DATAPLANE_SINGLE_NODE` is true

rabi:
recheck

rabi:
recheck

abays (approved):
/lgtm

PR #405, opened 2023-07-13
Improve variables consistency in devsetup dir

pablintino:
Till this PR multiple variables for the same purpose exist. OUTPUT_BASEDIR and OUTPUT_DIR pointed to the same place, so only OUTPUT_DIR will remain.
Something similar happened with the SSH_KEY and the SSH_KEY_FILE so both now converge to SSH_KEY_FILE name.

### DIFF
--- a/devsetup/scripts/edpm-compute-bmaas.sh
+++ b/devsetup/scripts/edpm-compute-bmaas.sh
@@ -14,16 +14,18 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 set -ex
+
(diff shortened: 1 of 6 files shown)


cjeanner:
/retest
/lgtm

Some consistency is always a good thing.

slagle:
Yes, this looks good. However this same commit is included in https://github.com/openstack-k8s-operators/install_yamls/pull/407 which I just approved. So I think you can drop this one.

pablintino:
> Yes, this looks good. However this same commit is included in #407 which I just approved. So I think you can drop this one.

True, it was not desired but I left the initial commit on my HEAD without notice and pushed it in the new PR... Nice that both are in anyway. Thanks!

PR #406, opened 2023-07-13
Wait to ensure the kustomize file is written

luis5tb:
It seems the ansbile runner image is not getting kustomized due to a race between generating the file and using it


luis5tb:
The problem was in a different place

PR #410, opened 2023-07-14
Run cinder kuttl tests in arbitrary namespace

cescgina:
Run cinder kuttl tests in arbitrary namepsace instead of the default
openstack namespace.


cescgina:
recheck

abays (approved):
/lgtm

PR #411, opened 2023-07-15
Remove some redundant lines from kustomize yaml

rabi:
Cleanups the `else` clause which overrides `provisioningInterface` when using two nodes.


rabi:
recheck

fao89 (approved):

PR #412, opened 2023-07-17
Addional Ceph mon related config for multinode OCP

fmount:
When `HOSTNETWORK` is false and we run a multinode cluster, we want to use the `podIP` to run the mon, so that it will be accessible in the k8s overlay from all the ctlplane pods using the openshift-sdn. This patch adds the additional `MON_CONF` to the main `Ceph` `CR` with the required bits to get the `POD IP` address before running the actual mon process.


rebtoor:
recheck

fmount:
@viroel @Sandeepyadav93 before merging this change I'd like to see a full run in the multinode OCP (prow?) with:

`make ceph HOSTNETWORK=false TIMEOUT=90`

viroel:
> @viroel @Sandeepyadav93 before merging this change I'd like to see a full run in the multinode OCP (prow?) with:
> 
> `make ceph HOSTNETWORK=false TIMEOUT=90`

Look good, since it is passing on install-yamls-deploy job. We need to change the deploy-step to have the HOSTNETWORK=false and also claim a 3-node cluster which is not enabled by default.
So I would say that is OK to merge this patch first and test it on openshift/release repo with HOSTNETWORK=false

fultonj (approved):

PR #413, opened 2023-07-18
Move netconfig_deploy as pretarget of openstack_deploy

rabi:
Ideally netconfig has to be created with the controlplane and used by dataplane. The other issue is infra-operator contains other services like memcached which are cleaned up when cleaning edpm.


abays (approved):
/lgtm

PR #414, opened 2023-07-18
Only cleanup netconfig CR in netconfig_deploy_cleanup

rabi:
Moving to `openstack_deploy_cleanup` does not seem to help with the cleanup as memcached has to be cleaned up after the services. This is workaround for that till we decide on moving memcached out of infra-operator.


abays (approved):
/lgtm

PR #415, opened 2023-07-19
Revert "Only cleanup netconfig CR in netconfig_deploy_cleanup"

rabi:
Reason for Revert: The reason why the change was proposed is probably not valid. With `netconfig_deploy` we only copy the netconfig CR to the `$DEPLOY_DIR`. Unless memcached  is deployed separately (normally it's part of the openstackcontrolplane CR), it should not exist in `$DEPLOY_DIR`.

Reverts openstack-k8s-operators/install_yamls#414


rabi:
recheck

rabi:
recheck

rabi:
Included in https://github.com/openstack-k8s-operators/install_yamls/pull/417 instead.

PR #417, opened 2023-07-19
Revert "Only cleanup netconfig CR in netconfig_deploy_cleanup"

rabi:
This also moves `netconfig_deploy` from `openstack_depploy_prep` to `openstack_deploy` as that is used in `openstack_crds` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -511,7 +511,7 @@ openstack_cleanup: operator_namespace## deletes the operator, but does not clean
 
 .PHONY: openstack_deploy_prep
 openstack_deploy_prep: export KIND=OpenStackControlPlane
-openstack_deploy_prep: openstack_deploy_cleanup netconfig_deploy ## prepares the CR to install the service based on the service sample file OPENSTACK
+openstack_deploy_prep: openstack_deploy_cleanup ## prepares the CR to install the service based on the service sample file OPENSTACK
 	$(eval $(call vars,$@,openstack))
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR} ${DEPLOY_DIR}
 	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(OPENSTACK_BRANCH),-b ${OPENSTACK_BRANCH}) ${OPENSTACK_REPO} "${OPERATOR_NAME}-operator" && popd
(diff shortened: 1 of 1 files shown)


karelyatin (commented):
/lgtm

Makefile
```diff
@@ -685,7 +685,8 @@ netconfig_deploy: input netconfig_deploy_prep ## installs the service instance u
 .PHONY: netconfig_deploy_cleanup
 netconfig_deploy_cleanup: ## cleans up the service instance, Does not affect the operator.
 	$(eval $(call vars,$@,infra))
-	oc delete --all netconfig --ignore-not-found=true || true
+	oc kustomize ${DEPLOY_DIR} | oc delete --ignore-not-found=true -f -
+	rm -Rf ${OPERATOR_BASE_DIR}/infra-operator ${DEPLOY_DIR}
```
abays, line 689: We have an issue with the overloading of the `infra-operator` dir.  Both `NetConfig` and `Memcached` use the same directory path for their CRs and thus would delete each other's content in that dir during `make (memcached|netconfig)_deploy_cleanup` [1].  They wouldn't be deleting the corresponding resource from the cluster, but when one of the targets removes this dir during `make <whatever>_deploy_cleanup` it would then prevent the other's `make <whatever>_deploy_cleanup` from properly removing the resource from the cluster if it was executed (because there would be no dir nor CR which to feed to `oc delete`).  I'm not sure at the moment what the best way to fix this is.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/b9f2345ffbdffc37ce0951b7f16d7ad036016630/Makefile#L708-L712

abays: Actually, the overloaded directory _would_ remove both the `Memcached` and `NetConfig` resources from the cluster whenever you execute `*_deploy_cleanup`, because the command is:

```
oc kustomize ${DEPLOY_DIR} | oc delete --ignore-not-found=true -f -
```

Any CRs in the `DEPLOY_DIR` would be deleted, and both `NetConfig` and `Memcached` could be in there.

abays: Maybe it might help to append `/memcached` and `/netconfig` to `DEPLOY_DIR` in the respective targets?  That might prevent them from deleting each other's resources.

rabi: Are you deploying memcached separately with `memcached_deploy`? Isn't it deployed with the openstack CR[1]? We only copy one CR to the $DEPLOY_DIR with `netconfig_deploy`[2], so that should only be cleaned up?

[1]https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/samples/core_v1beta1_openstackcontrolplane_galera.yaml#L43
[2] https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L676

abays: We can deal with this later on, as the concern here is not the normal path most people/CI use

rabi:
recheck

abays (approved):
/lgtm

rabi:
So edpm jobs are broken[1] by https://github.com/openstack-k8s-operators/openstack-operator/pull/397. We probably need to force merge this @abays ?

[1] https://logserver.rdoproject.org/17/417/c852980f08806855963e57129966ba071aada270/github-check/install-yamls-crc-podified-galera-deployment/0050728/job-output.txt

PR #421, opened 2023-07-19
Parameterize DATAPLANE_NETWORK_INTERFACE_NAME

raukadah:
In CI, we need to use eth1 instead of eth0 in the interface name. This patch parameterized it.


rebtoor (approved):

danpawlik (approved):

slagle:
recheck

slagle (approved):

fao89:
recheck

scripts/gen-edpm-baremetal-kustomize.sh
```diff
@@ -108,7 +108,7 @@ patches:
         # These vars are for the network config templates themselves and are
         # considered EDPM network defaults.
         neutron_physical_bridge_name: br-ex
-        neutron_public_interface_name: eth0
+        neutron_public_interface_name: ${EDPM_NETWORK_INTERFACE_NAME}
```
rabi, line 111: What nic config/template we're going to use with this var? The one currently used does not use this ansible var[1].  Also, as I mentioned in another PR nic mappings can be set using `net_config_data_lookup`.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_network_config/templates/single_nic_vlans/single_nic_vlans.j2

cjeanner: I think we can, from ci-framework, in a hook we already have, nudge a small snippet of "kustomization" in the right location in order to ensure we're updating the "nic1" mapping and that neutron_public_interface_name thingy - and, basically, *all* of the possible things.

Sooo... 2 patches pushed by cifmw, one to "add" the os-net-config mapping, and a second to "replace" that specific ansible parameter. If we ensure the filename is loaded as the last bit, we should be good.

Thanks for the pointers, @rabi !

cjeanner: After some discussions:
we can't use kustomize for now, because the `ansibleVars` is a plain string - no mapping. So we can't add/edit its content, as anything would override the whole content:
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/common.go#L54

There's apparently a plan to change this parameter type in the `dataplane-operator` but for now, it's not here, and we can't just wait for it to be done.

Therefore, we need this PR asap, as a workaround.

fao89:
recheck

stuggi:
merging to umblock the team

PR #424, opened 2023-07-20
Remove *_deploy_validate targets

kajinamit:
Current implementation of *_deploy_validate targets do not work fully, mainly because it does not consider how sample files are overridden when deploying an instance.

One good example is the secret spec. In sample files in each operator we sometimes have specific secrets instead of the global osp-secret, but the *_deploy targets now override secret reference to `osp-secret`. This inconsistency causes problem when we try to use the same assertion file for both deployment validation and kuttl tests, because we have to update the file according to the override made in deploy targets.

This removes the `*_deploy_validate` targets now to avoid that complexity. These targets are not really used except for mariadb one. Even for mariadb, in general we don't have to wait until MariaDB instance is fully ready, and operators should be able to wait until the instance becomes ready to proceed with the deployment steps such as DB creation.


cescgina (commented):
lgtm, removes unneeded complexity

kajinamit:
recheck

abays (approved):
/lgtm

(as long as we're sure no other repo relies on these targets -- I did a quick search and couldn't find anything)

abays:
recheck

abays:
recheck

abays:
Tests have passed independently across the various batched runs. Will override.

PR #425, opened 2023-07-20
Run mariadb kuttl tests in a specific namespace

kajinamit:


Makefile
```diff
@@ -1228,9 +1229,12 @@ nova_deploy_cleanup: namespace ## cleans up the service instance, Does not affec
 
 .PHONY: mariadb_kuttl_run
 mariadb_kuttl_run: ## runs kuttl tests for the mariadb operator, assumes that everything needed for running the test was deployed beforehand.
-	kubectl-kuttl test --config ${MARIADB_KUTTL_CONF} ${MARIADB_KUTTL_DIR}
+	MARIADB_KUTTL_DIR=${MARIADB_KUTTL_DIR} kubectl-kuttl test --config ${MARIADB_KUTTL_CONF} ${MARIADB_KUTTL_DIR} --namespace ${NAMESPACE}
```
cescgina, line 1232: A quick grep does not show any use for `MARIADB_KUTTL_DIR` in the tests, I don't think we need it here

kajinamit: You are correct and this is not needed.

I was wondering adding this consistently would be helpful for future test implementation which would require path to some test files, but I can drop this now.

kajinamit:
We might want to hold this during the dev preview release process, because we are close to tag mariadb-operator.

Makefile
```diff
@@ -101,18 +101,24 @@ KEYSTONE_KUTTL_DIR       ?= ${OPERATOR_BASE_DIR}/keystone-operator/tests/kuttl/t
...
+endif
+MARIADB_CR              ?= ${OPERATOR_BASE_DIR}/mariadb-operator/${MARIADB}
+MARIADB_DEPL_IMG        ?= unused
+MARIADB_KUTTL_CONF      ?= ${OPERATOR_BASE_DIR}/mariadb-operator/kuttl-test.yaml
+MARIADB_KUTTL_DIR       ?= ${OPERATOR_BASE_DIR}/mariadb-operator/tests/kuttl/tests
+MARIADB_KUTTL_NAMESPACE ?= mariadb-kuttl-tests
+ifeq ($(DBSERVICE), galera)
+MARIADB_ASSERT          ?= ${MARIADB_KUTTL_DIR}/galera_deploy/01-assert.yaml
+else
+MARIADB_ASSERT          ?= ${MARIADB_KUTTL_DIR}/mariadb_deploy/01-assert.yaml
```
abays, line 120: Is `MARIADB_ASSERT` actually used anywhere?

abays:
Failed on PV creation.  Retrying `install-yamls-deploy`.

/test install-yamls-deploy

abays:
recheck

Makefile
kajinamit: No it is no longer needed. I've removed it.

abays:
recheck

abays (approved):
/lgtm

PR #426, opened 2023-07-20
Allow to set default gateway

cjeanner:
In CI, the default gateway isn't the one associated with the private
network. Allowing to pass it will ensure we don't lose public access.


cjeanner:
recheck

cjeanner:
Issue is unrelated to the change, re-running "in case I'm lucky" with nested crc gods:
https://review.rdoproject.org/zuul/build/5d33507a78534750957d8c02451cfe82/log/ci-framework-data/logs/ci_make_006_run_edpm_deploy.log#14

@stuggi @fao89 FYI, this is needed for the extracted crc layout, as explained in the commit message. Of course, once we can edit the ansibleVars directly, this kind of change won't be necessary anymore :).
Care to give it a push even if CI fails? It's working on the test patch here:
https://github.com/openstack-k8s-operators/ci-framework/pull/361

We're currently facing the "next error", which is DNS related. We have to kustomize dnsmasq, but that's on us only :).

Thanks!

fao89 (approved):

abays:
The jobs all passed independently across two attempts, so we are merging this.

PR #427, opened 2023-07-21
Allow adding external kustomization for the openstack_deploy target

pablintino:
Since now in the extracted crc ci-framework jobs we are using a multinode setup with two interfaces per node, we need to kustomize parameters of the OpenStackControlPlane CR like the upstream DNS server used by the deployed dnsmasq.

Without this change, only the kustomization added by the gent-service-kustomize.sh can be applied as it replaces the kustomize.yaml file. With this change, the content of the generated kustomization by gen-service-kustomize.sh will be appended to the already existing kustomize.yaml, if already exists.

### DIFF
--- a/scripts/gen-service-kustomize.sh
+++ b/scripts/gen-service-kustomize.sh
@@ -44,7 +44,8 @@ fi
 
 pushd ${DEPLOY_DIR}
 
-cat <<EOF >kustomization.yaml
+cat <<EOF >>kustomization.yaml
+---
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:


abays (approved):
This might be a concern if `*_deploy_cleanup` wasn't a pretarget for `*_deploy_prep` (because then you might have multiple identical `Kustomization` resources in the same `kustomization.yaml` file if you ran `make *_deploy` multiple times and I'm not sure how that would behave), but since we always clean the `out/openstack/<operator>/cr` thanks to the `*_deploy_cleanup` pretarget, this should be fine.

/lgtm

pablintino:
> This might be a concern if `*_deploy_cleanup` wasn't a pretarget for `*_deploy_prep` (because then you might have multiple identical `Kustomization` resources in the same `kustomization.yaml` file if you ran `make *_deploy` multiple times and I'm not sure how that would behave), but since we always clean the `out/openstack/<operator>/cr` thanks to the `*_deploy_cleanup` pretarget, this should be fine.
> 
> /lgtm

Good catch, true, in our case we start always from a clean env, so there shouldn't be anything there from previous runs.

abays:
recheck

pablintino:
> recheck

I have the feeling that this change makes kustomize apply only the first kustomization in the manifest yaml... In the extracted job that we are preparing the STORAGE_CLASS doesn't get replaced by the value we were passing. I'll try to troubleshoot this and I'll reach you with some news.

pablintino:
Closing as it seems that kustomize can only apply one single Kustomization resource in a single file.

PR #429, opened 2023-07-24
Remove hardcoded default route

rabi:
Use the details  from netconfig.


rabi:
recheck

abays (approved):
This looks similar to what I did to get this working in a non-CRC cluster

/lgtm

PR #432, opened 2023-07-25
Make netconfig_deploy_cleanup optionally rm the CRs

pablintino:
For CI we need to inject our own Kustomizations. We placed back in the day a parameter that allows skipping the deletion of the defined CRs in the operators cloned repo, but the netconfig_deploy targets are missing that parameter.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,13 @@ DBSERVICE_CONTAINER = mariadb-openstack
 endif
 METADATA_SHARED_SECRET   ?= 1234567842
 HEAT_AUTH_ENCRYPTION_KEY ?= 767c3ed056cbaa3b9dfedb8c6f825bf0
+
+# Allows overriding the cleanup command used in *_cleanup targets.
+# Useful in CI, to allow injectin kustomization in each operator CR directory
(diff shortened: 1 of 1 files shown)


rabi:
There are a few other targets using `rm -Rf`[1]. We should add a comment on why it was added[2] for others to know.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L667
[2] https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L17

pablintino:
> There are a few other targets using `rm -Rf`[1]. We should add a comment on why it was added[2] for others to know.
> 
> [1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L667 [2] https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L17

Let me add a comment and address the other targets. That is used in CI only cause we start from a fresh/clean install_yamls clone, that's why it's safe to use in CI, but I'll add a comment anyway.

pablintino:
recheck

cjeanner:
recheck

abays:
recheck

abays (approved):
/lgtm

PR #439, opened 2023-07-27
Openstack Kuttl: crc_storage retries &  error detection

Sandeepyadav93:
With this PR, we do below:-

* Replace crc_storage with new target with retries to improve kuttl passing rate as `crc_storage` is flaky.
* Add error detection with `set -e` as currently if commands are failing we are still moving ahead.


abays:
recheck

slagle (commented):
+2

abays (approved):
/lgtm

jparoly:
Tested this locally with crc and with 6 tests merged with https://github.com/openstack-k8s-operators/openstack-operator/pull/384 and all tests passed!  Also when creating a failure condition in one test it fails then exits the loop and does not run further tests, which is what we want at this point.

jparoly (approved):

PR #441, opened 2023-07-28
Ceph: set namespace when using "oc rsh"

gouthampacha:
When using multiple namespaces, the script may not find the "ceph" pod and fail.


fmount:
Thanks, that makes a lot of sense given we have multiple namespaces in place.

fmount (approved):

PR #443, opened 2023-07-31
Use dedicated timeout env var for certmanager target

abays:
Some environments seem to have a default `TIMEOUT` env var that gets picked-up and used, which causes problems for the `certmanager` targets:

```
while ! (oc get pod --no-headers=true -l name=cert-manager-operator -n cert-manager-operator| grep "cert-manager-operator"); do sleep 10; done
No resources found in cert-manager-operator namespace.
No resources found in cert-manager-operator namespace.
No resources found in cert-manager-operator namespace.
cert-manager-operator-controller-manager-6d49987bb7-wgzqc   0/2   ContainerCreating   0     9s
oc wait pod -n cert-manager-operator --for condition=Ready -l name=cert-manager-operator --timeout=90
error: invalid argument "90" for "--timeout" flag: time: missing unit in duration "90"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -338,6 +338,9 @@ SWIFT_BRANCH     ?= main
 SWIFT            ?= config/samples/swift_v1beta1_swift.yaml
 SWIFT_CR         ?= ${OPERATOR_BASE_DIR}/swift-operator/${SWIFT}
 
+# CertManager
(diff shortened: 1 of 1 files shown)


Sandeepyadav93:
recheck

Sandeepyadav93:
recheck

PR #445, opened 2023-08-01
Refactoring inline script to use modules

jpodivin:


devsetup/edpm/services/dataplane_v1beta1_openstackdataplaneservice_reposetup.yaml
```diff
@@ -15,13 +15,18 @@ spec:
...
+          ansible.builtin.package:
+            name:
+              - git
+              - python-setuptools
+              - python-requests
+              - python3-pip
+            state: present
+        - name: Install repo-setup directly from github
+          ansible.builtin.pip:
+            name: git+https://github.com/openstack-k8s-operators/repo-setup
```
slagle, line 28: become: true? Or is it not needed?

jpodivin: I was a wondering about that too. But it appears that the user executing these is already given all permissions necessary. And the standard deployment runs as expected. Still, just in case the situation changes I can have it future proofed.

slagle (approved):

steveb (changes requested):
Installing git and pip from rpm pulls in a large number of packages just to pip install something which bootstraps the repo setup, this is why I proposed this script based approach which requires neither git nor pip.

The commit doesn't describe why this change is proposed, is it just to have some clean ansible to do this task? I'm not certain the extra time and network transfer is worth it in this case

jpodivin:
Since a different patch was merged in the meantime eliminating need for this change 

> Installing git and pip from rpm pulls in a large number of packages just to pip install something which bootstraps the repo setup, this is why I proposed this script based approach which requires neither git nor pip.
> 
> The commit is against the previous version of this script, we should close it now I think?

Yes it is against previous version of the script, the PR changing it was merged after I've submitted this one. Therefore there is no need for this change and I'm closing the PR.

PR #447, opened 2023-08-01
DNM -  Add more EDPM deploy options for HCI

viroel:
This patch it might not be needed anymore with "edpm_kustomize" role from ci-framework. but still using it for tests.

This patch adds new options to used for HCI deployment. With these options it is possible to configure dataplane network only on EDPM nodes, before deploying Ceph. Adds options to deploy nova with ceph as backend and to configure storage management network.


viroel:
recheck

viroel:
recheck

viroel:
recheck

PR #448, opened 2023-08-01
Rename Ceph vars to avoid conflicts

viroel:
This patch renames most of Ceph variables from gen-ceph-kustomize script which can conflict with vars from other scripts (e.g TIMEOUT from operator-wait.sh)

Depends-On: https://github.com/openshift/release/pull/41876


abays (approved):
/lgtm

PR #449, opened 2023-08-01
Ensure dnsmasq/dns and netconfig/netconfig don't exist during kuttl tests

slagle:
These resources are not expected to exist during kuttl test execution.
When they do, the baremetal conditions are added to the
OpenStackDataPlaneRole resources which causes the kuttl tests to fail
various assertions.

These resources are created when an OpenStackControlPlane is created
with install_yamls. Testing for them before executing kuttl tests from
dataplane-operator makes it clear that they should not exist.


fao89:
recheck

fao89 (approved):

PR #451, opened 2023-08-02
Remove EDPM_NETWORK_CONFIG_TEMPLATE

bshephar:
As part of a wider effort to simplify the interface for users while configuring nic-configs. We are removing the option to provide a filesystem path, since this method is only able to reference default jinja2 templates from the edpm-ansible project. Accordingly, we need to update this in install_yamls.


fao89 (approved):

fao89:
recheck

bshephar:
recheck

bshephar:
recheck

slagle (changes requested):
Can you drop the 2 Depends-On?

Then this needs to be rebased and also drop the lines in the kustomize scripts where we have:

```
- op: add                                                                  
  path: /spec/nodeTemplate/ansible/ansibleVars/edpm_network_config_template
```

bshephar:
> Can you drop the 2 Depends-On?
> 
> Then this needs to be rebased and also drop the lines in the kustomize scripts where we have:
> 
> ```
> - op: add                                                                  
>   path: /spec/nodeTemplate/ansible/ansibleVars/edpm_network_config_template
> ```

I was wondering if we want an interface to change this template from install_yamls? I was considering leaving the environment variable and customizing it with the same script, but it would at least allow you to customize it using install_yamls.

Probably too much complexity to maintain though when a user can just provide their own DataPlaneNodeSet CR if they really want to customize the networks. I'll drop it for now and we can come back to it if we really need the functionality

pablintino:
@bshephar please, just ping me when this is in to update the framework. With this I can get rid off this lines: https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/hooks/playbooks/fetch_compute_facts.yml#L96C1-L97C94

bshephar:
recheck

bshephar:
recheck

bshephar:
Unrelated failure:
```
[pod/dataplane-deployment-download-cache-openstack-edpm-ipam-nlj9t/openstackansibleee] [0;31mfatal: [edpm-compute-0]: FAILED! => {"attempts": 5, "changed": false, "msg": "Failed to pull image quay.io/podified-antelope-centos9/openstack-nova-libvirt:current-podified"}[0m
```

rabi:
recheck

PR #457, opened 2023-08-03
Split out the tripleo-repos commands so we can skip downstream

marios:


devsetup/scripts/standalone.sh
```diff
@@ -23,7 +23,8 @@ IP="192.168.122.${IP_ADRESS_SUFFIX}"
 OUTPUT_DIR=${OUTPUT_DIR:-"${SCRIPTPATH}/../../out/edpm/"}
 SSH_KEY_FILE=${SSH_KEY_FILE:-"${OUTPUT_DIR}/ansibleee-ssh-key-id_rsa"}
 SSH_OPT="-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i $SSH_KEY_FILE"
-CMDS_FILE=${CMDS_FILE:-"/tmp/standalone_repos"}
+REPO_SETUP_CMDS=${REPO_SETUP_CMDS:-"/tmp/standalone_repos"}
+CMDS_FILE=${CMDS_FILE:-"/tmp/standalone_cmds"}
 CLEANUP_DIR_CMD=${CLEANUP_DIR_CMD:-"rm -Rf"}
```
marios, line 29: probably safer to also default SKIP_TRIPLEO_REPOS to "" here too
will do on next update

```

SKIP_TRIPLEO_REPOS=${SKIP_TRIPLEO_REPOS:=""}
```

marios: actually default 'false'

devsetup/scripts/standalone.sh
```diff
@@ -80,8 +87,12 @@ scp $SSH_OPT standalone/ceph.sh root@$IP:/tmp/ceph.sh
 scp $SSH_OPT standalone/openstack.sh root@$IP:/tmp/openstack.sh
 
 # Running
-ssh $SSH_OPT root@$IP "bash /tmp/repo-setup.sh"
-# separate the two commands so we can properly detect whether there is
-# a failure while deploying standalone
-ssh $SSH_OPT root@$IP "rm -f /tmp/repo-setup.sh"
-${CLEANUP_DIR_CMD} $CMDS_FILE
+if [[ -z ${SKIP_TRIPLEO_REPOS} || ${SKIP_TRIPLEO_REPOS} == "false" ]]; then
+    ssh $SSH_OPT root@$IP "bash /tmp/repo-setup.sh"
```
raukadah, line 91: In this pr, we have added two vars $REPO_SETUP_CMDS and $CMDS_FILE.
$REPO_SETUP_CMDS will be used in RDO upstream zuul jobs to enable repos using triple repos and content from $REPO_SETUP_CMDS via repo-setup.sh.

$CMDS_FILE will be used in downstream zuul jobs, to enable custom rhel repos via standalone-deploy.sh

and if we set SKIP_TRIPLEO_REPOS to true then, it will be skipped.

Looks good.

Can we initialize `SKIP_TRIPLEO_REPOS=''"` so that this if conditions always run for upstream and it will avoid to set extra var in the job definition?

devsetup/scripts/standalone.sh
```diff
@@ -80,8 +87,12 @@ scp $SSH_OPT standalone/ceph.sh root@$IP:/tmp/ceph.sh
...
+    # separate the two commands so we can properly detect whether there is
+    # a failure while deploying standalone
+    ssh $SSH_OPT root@$IP "rm -f /tmp/repo-setup.sh"
+fi
+ssh $SSH_OPT root@$IP "bash /tmp/standalone-deploy.sh"
+
+${CLEANUP_DIR_CMD} $REPO_SETUP_CMDS
```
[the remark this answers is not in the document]
marios: ack ok adding

raukadah (approved):

PR #460, opened 2023-08-04
edpm-compute-baremetal.sh use the --create, --cleanup pattern

steveb:
Also properly set the node count variables so that counts greater than 1 work.


slagle (approved):

rabi:
recheck

abays:
recheck

rabi:
recheck

Makefile
```diff
@@ -610,6 +610,8 @@ edpm_deploy_baremetal_prep: export EDPM_TOTAL_NODES=${DATAPLANE_TOTAL_NODES}
 edpm_deploy_baremetal_prep: export EDPM_NETWORK_CONFIG_TEMPLATE=${DATAPLANE_NETWORK_CONFIG_TEMPLATE}
 edpm_deploy_baremetal_prep: export EDPM_SSHD_ALLOWED_RANGES=${DATAPLANE_SSHD_ALLOWED_RANGES}
 edpm_deploy_baremetal_prep: export EDPM_CHRONY_NTP_SERVER=${DATAPLANE_CHRONY_NTP_SERVER}
+edpm_deploy_baremetal_prep: export EDPM_REGISTRY_URL=${DATAPLANE_REGISTRY_URL}
+edpm_deploy_baremetal_prep: export EDPM_CONTAINER_TAG=${DATAPLANE_CONTAINER_TAG}
```
abays, line 614: `make edpm_deploy_baremetal` doesn't work right now without these lines added

abays: I am proposing this same fix in another PR to try to get it merged sooner to unblock the baremetal Zuul job: https://github.com/openstack-k8s-operators/install_yamls/pull/468.  So if this PR merges first, I'll just close that one.

rabi: Hmm...`edpm_deploy_baremetal` target is not used in the zuul job. That's only for deploying with existing BMHs.

abays: Ah, okay.  So something else might be causing repeated failures there.

devsetup/Makefile
```diff
@@ -36,7 +36,7 @@ BM_PROVISIONING_INTERFACE  ?=enp6s0
 BM_NETWORK_NAME            ?=default
 BM_NETWORK_IPADDRESS       ?=192.168.122.1
 BM_INSTANCE_NAME_PREFIX    ?=edpm-compute
-BM_NODE_COUNT              ?=1
+BM_NODE_COUNT              ?=2
```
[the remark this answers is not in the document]
steveb: In that case how about we set DATAPLANE_TOTAL_NODES to 1 so that the defaults work for local development?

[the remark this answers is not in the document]
rabi: Yeah, that one is shared by both `edpm_deploy` and `edpm_deploy_baremetal`. It should have been set to 1 by default.  `edpm_deploy` target uses another var `DATAPLANE_SINGLE_NODE` which is set to `true`[1]  and hence the confusion. We probably need to clean these up later to use consistent set of vars.  


We don't use  `edpm_deploy_baremetal` in CI (instead use `edpm_baremetal` target which does everything together) and therefore broken after some recent changes.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L302
[2] https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/Makefile#L101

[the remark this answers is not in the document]
steveb: OK, I have removed this commit

rabi:
recheck

rabi:
recheck

rabi (approved):
/lgtm

rabi:
All jobs green in multiple runs, merging.

rabi:
> @rabi: [rabi](https://github.com/rabi) unauthorized: /override is restricted to Repo administrators, approvers in top level OWNERS file, and the following github teams:.

I forgot I'm not a OWNER :smile:

PR #467, opened 2023-08-08
Pin Podman for Wallaby Standalone deployment

jistr:
We hit an issue where the TripleO container-puppet-<SERVICE> container weren't creating any config files. Takashi found out this was caused by regression in environment variable handling in Podman. The STEP_CONFIG variable, which in our case often contains newlines, was getting trimmed to the first line only.

The change happened somewhere between podman-4.6.0-1.el9.x86_64 (working fine) and podman-4.6.0-3.el9.x86_64 (broken). We pin to the working package for now to work around the issue.


jistr:
Testing...

fao89 (approved):

jistr:
Worked in local testing.

PR #474, opened 2023-08-09
Speed up standalone deployment

fao89:


devsetup/scripts/standalone.sh
```diff
@@ -56,6 +56,12 @@ sudo dnf update -y
 EOF
 fi
 
+if [[ -e /run/systemd/resolve/resolv.conf ]]; then
+    HOST_PRIMARY_RESOLV_CONF_ENTRY=$(cat /run/systemd/resolve/resolv.conf | grep ^nameserver | grep -v '192.168' | head -n1)
```
jistr, line 60: The variable will contain "nameserver <IP>" (a full line from resolv.conf) but i think in the code below we want just "<IP>" format. I think we should add something like `| awk '{ print $2; }'` at the end of the pipe.

devsetup/scripts/standalone.sh
```diff
@@ -56,6 +56,12 @@ sudo dnf update -y
 EOF
 fi
 
+if [[ -e /run/systemd/resolve/resolv.conf ]]; then
+    HOST_PRIMARY_RESOLV_CONF_ENTRY=$(cat /run/systemd/resolve/resolv.conf | grep ^nameserver | grep -v '192.168' | head -n1 | cut -d' ' -f2)
+else
+    HOST_PRIMARY_RESOLV_CONF_ENTRY=$(cat /etc/resolv.conf | grep ^nameserver | grep -v '192.168' | head -n1 | cut -d' ' -f2)
+fi
+
```
jistr, line 64: @cescgina Can you please try to check if this will yield a good value in a CI host? The variable should contain a DNS IP that the host is configured with to resolve names in the outside world (as opposed to resolve names e.g. from CRC).

cescgina: @jistr tried the second command on a CI host, it retruns `127.0.0.1`, and `/run/systemd/resolve/resolv.conf` does not exist there

PR #479, opened 2023-08-15
dataplane: Install repo-setup without installing git or pip

steveb:
This switches the repo-setup install to download a github tarball and using the pip bundled with python to install repo-setup in a venv.

This means no packages need to be installed before the repos are set up. This aligns with the repo-setup recipe in
devsetup/scripts/edpm-compute-repos.sh.


rabi (commented):
/lgtm

slagle (approved):

PR #480, opened 2023-08-16
Don't run certmanager with openstack target

rabi:
`crc_bmo_setup` would install certmanager. We don't need to install it by default.


stuggi:
we'd need cert-manager per default as soon we add TLS ... which is soon. the plan is to have TLS enabled per default.

rabi:
> we'd need cert-manager per default as soon we add TLS ... which is soon. the plan is to have TLS enabled per default.

Can we add it back along with TLS support rather than installing it by default now? I would also not expect it to be used when TLS is disabled.

PR #481, opened 2023-08-16
Allow using different controlplane IP/interface with nncp

rabi:
This would allow us to use BM network as ctlplane in a cluster along with the configured IPs in the workers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -325,6 +325,10 @@ CEPH_IMG            ?= quay.io/ceph/demo:latest
 NNCP_INTERFACE      ?= enp6s0
 NNCP_TIMEOUT		?= 240s
 NNCP_CLEANUP_TIMEOUT	?= 120s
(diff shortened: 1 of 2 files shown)


fao89 (commented):
+2

rabi:
> > This would allow us to use BM network as ctlplane in a cluster along with the configured IPs in the workers.
> 
> So the idea is to piggy-back on the OCP cluster network? Just trying to understand exactly what you mean.

Yes, that's one of the uses with dev-scripts or IPI installer deployed clusters to leverage BM network. But it can also be used for any other network configured with OCP cluster. We should not be hard-coding ctlplane ips. Those can be different as configured in netconfig[1]

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/config/samples/network_v1beta1_netconfig.yaml#L12

abays (approved):
/lgtm

PR #482, opened 2023-08-16
Bypass containers-prepare-parameters and repo-setup files for downstream

marios:
The file is instead generated by d/stream play - see [1][2]

[1] https://review.rdoproject.org/r/q/topic:oooci_data_plane_adoption
[2] https://code.engineering.redhat.com/gerrit/q/topic:oooci_data_plane_adoption

### DIFF
--- a/devsetup/scripts/standalone.sh
+++ b/devsetup/scripts/standalone.sh
@@ -40,6 +40,7 @@ fi
 # And when running it from our own systems outside of the Red Hat network we can use any available server:
 # export NTP_SERVER=pool.ntp.org
 
+if [[ ! -f $REPO_SETUP_CMDS ]]; then
 cat <<EOF > $REPO_SETUP_CMDS
 set -ex
 sudo dnf remove -y epel-release
@@ -53,6 +54,7 @@ sudo -E tripleo-repos -b wallaby current-tripleo-dev ceph --stream
 sudo dnf repolist
 sudo dnf update -y
 EOF
+fi
 
 if [[ -e /run/systemd/resolve/resolv.conf ]]; then
     HOST_PRIMARY_RESOLV_CONF_ENTRY=$(cat /run/systemd/resolve/resolv.conf | grep ^nameserver | grep -v '192.168' | head -n1 | cut -d' ' -f2)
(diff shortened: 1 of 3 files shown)


pablintino:
recheck

raukadah (commented):
+2 checked the downstream testproject . Downstream and upstream are working with containers-prepare-parameters

raukadah (approved):

devsetup/scripts/standalone.sh
```diff
@@ -95,6 +97,9 @@ scp $SSH_OPT standalone/deployed_network.yaml root@$IP:/tmp/deployed_network.yam
 scp $SSH_OPT standalone/network.sh root@$IP:/tmp/network.sh
 scp $SSH_OPT standalone/ceph.sh root@$IP:/tmp/ceph.sh
 scp $SSH_OPT standalone/openstack.sh root@$IP:/tmp/openstack.sh
+if [[ -f /home/zuul/containers-prepare-parameters.yaml ]]; then
```
cescgina, line 100: I think we should replace here `/home/zuul` by `$HOME`, as people are using the standalone target to test adoption in scenarios different than CI

marios: ack will do thanks

devsetup/scripts/standalone.sh
```diff
@@ -77,6 +79,16 @@ export NTP_SERVER=${NTP_SERVER:-"clock.corp.redhat.com"}
 export EDPM_COMPUTE_CEPH_ENABLED=${EDPM_COMPUTE_CEPH_ENABLED:-true}
 export CEPH_ARGS="${CEPH_ARGS:--e \$HOME/deployed_ceph.yaml -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm-rbd-only.yaml}"
 
+if [[ -f $HOME/containers-prepare-parameters.yaml ]]; then
+    echo "Using existing containers-prepare-parameters.yaml - contents:"
+    cat $HOME/containers-prepare-parameters.yaml
+else
+    openstack tripleo container image prepare default \
+        --output-env-file $HOME/containers-prepare-parameters.yaml
```
marios, line 87: the ceph command uses /root/containers-prepare-parameters.yaml as it runs as root user.

so will also need to copy it there after this generation

https://logserver.rdoproject.org/02/49902/1/check/data-plane-adoption-github-rdo-centos-9-crc-single-node/847f76c/ci-framework-data/logs/ci_make_002_run_standalone.log

The output file /home/zuul/containers-prepare-parameters.yaml can not be created. Error: [Errno 2] No such file or directory: '/home/zuul/containers-prepare-parameters.yaml'

cescgina: Also we need to escape `$HOME` because we need that to point to the the home of the vm and not of the host

marios: thanks @cescgina nice catch will add

marios: no actually, at this level we want it to pickup from the host.. it is doing the copy with ssh into the VM this is outside the heredoc

marios: you *are* right, i was looking at the other snippet below line 110  ... which is outside the heredoc

this one indeed is in the script that is executed on the vm ... adding \

PR #483, opened 2023-08-16
More EDPM baremetal customization options

abays:
Exposes a couple more customization options for baremetal EDPM deployment

PR #488, opened 2023-08-21
Add kustomization for root password

rabi:
This would help us debug `edpm_baremetal` target used in CI jobs.


abays:
I did something similar in the main `scripts` dir for EDPM baremetal, but maybe it was the wrong place: https://github.com/openstack-k8s-operators/install_yamls/pull/483

rabi:
> I did something similar in the main `scripts` dir for EDPM baremetal, but maybe it was the wrong place: #483

Yeah, that target is not used in CI with crc. That's for OCP cluster with CBO and BMHs.

abays (approved):
/lgtm

PR #491, opened 2023-08-22
Use latest CentOS Stream 9 image

kajinamit:
... instead of a specific version. An old version can be removed from the mirror after a few new versions are released.


cjeanner:
/lgtm

If people want a specific image, they can override it using an env var. FYI, we're doing the same kind of change in the CI Framework here: https://github.com/openstack-k8s-operators/ci-framework/pull/464

fao89 (approved):

PR #492, opened 2023-08-22
Fix wrong logic to check existence of .ssh dir

kajinamit:
The current logic assumes $HOME/.ssh is a file but it should be a directory.

This also simplifies the selinux relabeling using restorecon because the context explicitly defined is the default one.


kajinamit:
recheck

fao89 (approved):

PR #493, opened 2023-08-25
Improved VLAN customization for NNCP/NetAttach

abays:
Minor ability to customize VLANs used in `NNCP`s and `NAD`s.  For example:

```
NETWORK_VLAN_START=30 NETWORK_VLAN_STEP=10 make openstack
```

...would use `30`, `40`, `50`, etc for `internalapi`, `storage`, `tenant`, etc.  Which would then be paired with a custom `NetConfig` CR with those VLANs passed to `make openstack_deploy`:

```
NETCONFIG_CR=/path/to/custom/netconfig.yaml make openstack_deploy
```

Recent forays into baremetal Podified deployments have unveiled the need for such customization.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,8 @@ CRC_STORAGE_RETRIES ?= 3
 # network isolation
 NETWORK_ISOLATION   ?= true
 NETWORK_MTU         ?= 1500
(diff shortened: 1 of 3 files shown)


abays:
recheck

PR #494, opened 2023-08-26
scripts: Add Additional Kustomization For Baremetal Deployment

vkhitrin:
Adding an option to inject additional EDPM variables for baremetal
deployments using Kustomize.
The following variables were added:
  - EDPM_BAREMETAL_KERNEL_ARGS, uses `edpm-kernel` role to modify kernel
    arugments
  - EDPM_BAREMETAL_KERNEL_HUGEPAGES, uses `edpm-kernel` role to
    configure hugepages
  - EDPM_BAREMETAL_TUNED_PROFILE, uses `edpm_tuned`
    role to configure a tuned profile
  - EDPM_BAREMETAL_TUNED_ISOLATED_CORES, uses `edpm_tuned` role to
    configured tuned isolated cores
  - EDPM_BAREMETAL_NETWORK_CONFIG_OVERRIDE, uses `edpm_network_config`
    role to allow providing custom network configuration

### DIFF
--- a/scripts/gen-edpm-baremetal-kustomize.sh
+++ b/scripts/gen-edpm-baremetal-kustomize.sh
@@ -51,11 +51,25 @@ if [ -z "$EDPM_BMH_NAMESPACE" ]; then
     echo "Please set EDPM_BMH_NAMESPACE"; exit 1
 fi
 
+# This statement is first since we don't want to start creating CR if provided file is not correct
+if [ ! -z "$EDPM_BAREMETAL_NETWORK_CONFIG_OVERRIDE" ]; then
+    if [ ! -f "$EDPM_BAREMETAL_NETWORK_CONFIG_OVERRIDE" ]; then
+        echo "Please esnure file ${EDPM_BAREMETAL_NETWORK_CONFIG_OVERRIDE} exists"
+        exit 1
+    fi
+    NETWORK_CONFIG_JSON=$(cat "${EDPM_BAREMETAL_NETWORK_CONFIG_OVERRIDE}" | jq -c -r . 2>/dev/null)
+    if [ ! "$NETWORK_CONFIG_JSON" ]; then
+        echo "Please esnure file ${EDPM_BAREMETAL_NETWORK_CONFIG_OVERRIDE} contains a valid JSON"
+        exit 1
+    fi
+fi
+
 NAME=${KIND,,}
 
 if [ ! -d ${DEPLOY_DIR} ]; then
     mkdir -p ${DEPLOY_DIR}
 fi
+
 
 pushd ${DEPLOY_DIR}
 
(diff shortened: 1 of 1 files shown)


vkhitrin:
This might not be the ideal place to store these parameters. They may also be worthwhile to have in virtualized deployments.

In general, is there any mechanism that allows passing additional variables to EPDM?

vkhitrin:
@rabi, @fao89, when working on a baremetal NFV deployment, I injected even more variables not mentioned in this PR.
Perhaps I should refactor this to allow generic input that will populate variables instead of hard-codding Kustomizations.

rabi:
> @rabi, @fao89, when working on a baremetal NFV deployment, I injected even more variables not mentioned in this PR. Perhaps I should refactor this to allow generic input that will populate variables instead of hard-codding Kustomizations.

IMO, we should not add all these to `install_yamls`. All NFV kustomizations for downstream testing should preferably be in a downstream repo. If we still want to keep these in `install_yamls`, they should be in separate kustomization scripts.

vkhitrin:
@rabi, what about a requirement of allowing to Kustomize further unrelated to NFV use cases?
I think having a mechanism where we would be able to provide further Kustomizations via `install_yamls` is beneficial.

Also, on the scope of this project, as part of the `ci-framework` that it intended to be used, which wraps around this project, having to maintain internal repositories will cause our squad to deviate, which was already a pain point in TripleO deployments.

Jaganathancse:
> @rabi, what about a requirement of allowing to Kustomize further unrelated to NFV use cases? I think having a mechanism where we would be able to provide further Kustomizations via `install_yamls` is beneficial.
> 
> Also, on the scope of this project, as part of the `ci-framework` that it intended to be used, which wraps around this project, having to maintain internal repositories will cause our squad to deviate, which was already a pain point in TripleO deployments.

@vkhitrin As Rabi mentioned, no need to do the changes in install_yaml. we can have custom OpenstackDataplane CR for NFV use cases and also we can have specific kustomization for our requirements in downstream repo's. 
I am also working on OpenstackDataplane Samples for NFV feature roles. will discuss and finalize.

rabi:
> I think having a mechanism where we would be able to provide further Kustomizations via `install_yamls` is beneficial.

We also have to keep them maintainable and not make them messy with tons of vars, I don't mind if we've kuystomization scripts for specific scenarios maintained in install_yamls and then a var to pick one for a scenario.

vkhitrin:
@rabi, sorry for the delay.

My question is, would it be reasonable to drop the content of this patch and rewrite it in a way where we will use only a single generic variable that will contain a JSON that will contain all extra Ansible variables for EDPM:
```
export EDPM_ANSIBLE_CUSTOM_VARS='{"edpm_key1": "value1", "edpm_key2": "value2", edpm_key3: "value3}'
```

And then we will have a generic scrip that will iterrate through this JSON and build the Kustomization?

vkhitrin:
@ rabi, can you share a recommendation on proceeding with my last comment?
From how I see things,
1. We will use `ci-framework` to invoke `install_yamls`.
2. We will want to use `install_yamls` natively without maintaining a fork.
3. We must provide custom Kustomizations (mainly Ansible variables) to configure various hardware and topologies.

Is the way forward will be through `ci-framework` https://ci-framework.readthedocs.io/en/latest/roles/edpm_kustomize.html?

rabi:
> Is the way forward will be through ci-framework

Probably yes. Looks like there is some work happening to leverage multiple kustomizations vi ci-framework.  The following[1][2] could be relevant, though the kustomizations or the scripts generating them can still live in `install_yamls`.

[1] https://issues.redhat.com/browse/OSP-28087
[2] https://github.com/openstack-k8s-operators/ci-framework/pull/511

pablintino:
> > Is the way forward will be through ci-framework
> 
> Probably yes. Looks like there is some work happening to leverage multiple kustomizations vi ci-framework. The following[1][2] could be relevant, though the kustomizations or the scripts generating them can still live in `install_yamls`.
> 
> [1] https://issues.redhat.com/browse/OSP-28087 [2] [openstack-k8s-operators/ci-framework#511](https://github.com/openstack-k8s-operators/ci-framework/pull/511)

@vkhitrin @rabi I've just spotted this PR. Correct, we are working on providing support for multiple kustomizations in ci-framework, and yes, it'll be perfectly fine if install_yamls generates them in the CR dir, we will pick them UP.

Expect that feature to be in in the early days of next week.

PR #497, opened 2023-08-29
External RGW support in NG

fmount:
It's possible to configure `swift` with an `external RGW` backend.
This change allows the `Ceph` Pod to support a configured `external RGW` instance that can serve swift related requests.

This patch implements two main functions:

1. `config_ceph`: it's used to inject `config-keys` within the `Ceph mgr` database (we only have RGW related keys at the moment);
2. `config_rgw`: a post entry in the main switch case has been added; with this patch, the post action is only used to restart the RGW instance and reload its configuration

Finally, the `Makefile` `Ceph` target is now calling both `config` and `post` actions from the script.


fmount:
hold-the-node

viroel:
/test install-yamls-deploy

lets see if node hold step will run now

scripts/gen-ceph-kustomize.sh
```diff
@@ -205,6 +217,46 @@ function create_volume {
     oc rsh -n $NAMESPACE ceph ceph fs volume create cephfs >/dev/null || true
 }
 
+function config_ceph {
+    # Define any config option that should be set in the mgr database
+    # via associative arrays and inject to the Ceph Pod
+    # Define and set config options
+    echo "Apply Ceph config keys"
+    declare -A config_keys=(
+        ["rgw_keystone_url"]="http://keystone-public-openstack.$DOMAIN"
```
fmount, line 226: As per the way endpoints are exposed, identity (keystone) will create a GenericRoute [1] based on the `instance.ServiceName`, `EndpointType` and `Namespace`.
I'm hardcoding the `Namespace` here as it's the default in both development and Prow jobs, otherwise it can be an additional parameter, which I can introduce in the next PS.

[1]  https://github.com/openstack-k8s-operators/lib-common/blob/modules/test/v0.1.1/modules/common/route/route.go#L53

fmount: Done.

PR #498, opened 2023-08-30
Add kustomization to run repo-setup service in baremetal job

rabi:
We need to setup repos downstream and there is hack in ci-framework to use this service.


abays (approved):
/lgtm

PR #500, opened 2023-08-30
Add support for multicontrollers

averdagu:
Adopting from an OSP 17.1 wallaby deployment formed by 3 controllers + 2 computes requires a
lot of manually changing hardcoded values on a few different files to adapt to the network schema of
wallaby deployment. With this pull request I intend to make it parameterizable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,24 @@ endif
 METADATA_SHARED_SECRET   ?= 1234567842
 HEAT_AUTH_ENCRYPTION_KEY ?= 767c3ed056cbaa3b9dfedb8c6f825bf0
 
+# POD networks
+POD_INTERNALAPI_VLAN ?= 20
+POD_STORAGE_VLAN ?= 21
+POD_TENANT_VLAN ?= 22
+POD_INTERNALAPI_NET ?= 172.17.0
+POD_STORAGE_NET ?= 172.18.0
+POD_TENANT_NET ?= 172.19.0
+
+# multiple controller deployment (MCD)
+MCD_ENABLED 	?= true
+MCD_INTERNALAPI_VLAN ?= 20
+MCD_STORAGE_VLAN ?= 30
+MCD_TENANT_VLAN ?= 50
+MCD_INTERNALAPI_NET ?= 172.17.1
+MCD_STORAGE_NET ?= 172.17.3
+MCD_TENANT_NET ?= 172.17.2
+MCD_DATA_NET ?= 192.168.24
+
 # Allows overriding the cleanup command used in *_cleanup targets.
 # Useful in CI, to allow injectin kustomization in each operator CR directory
 # before the resource gets deployed. If it's not possible to inject Kustomizations/CRs
@@ -327,7 +345,21 @@ MANILA_KUTTL_NAMESPACE  ?= manila-kuttl-tests
 CEPH_IMG            ?= quay.io/ceph/demo:latest-reef
 
 # NNCP
-NNCP_INTERFACE      ?= enp6s0
+## Interfaces
+NNCP_INTERFACE ?= enp6s0
+NNCP_INTERFACE_DATA ?= $(if $(findstring true, $(MCD_ENABLED)),enp7s0,$(NNCP_INTERFACE))
+NNCP_INTERFACE_MANAGEMENT ?= $(if $(findstring true, $(MCD_ENABLED)),enp8s0,$(NNCP_INTERFACE))
+NNCP_INTERFACE_EXTERNAL ?= $(if $(findstring true, $(MCD_ENABLED)),enp9s0,$(NNCP_INTERFACE))
+## API
+NNCP_INTERNAL_API_VLAN ?= $(if $(findstring true, $(MCD_ENABLED)),$(MCD_INTERNALAPI_VLAN),$(POD_INTERNALAPI_VLAN))
+NNCP_STORAGE_VLAN ?= $(if $(findstring true, $(MCD_ENABLED)),$(MCD_STORAGE_VLAN),$(POD_STORAGE_VLAN))
+NNCP_TENANT_VLAN ?= $(if $(findstring true, $(MCD_ENABLED)),$(MCD_TENANT_VLAN),$(POD_TENANT_VLAN))
+## Networks
+NNCP_INTERNAL_API_NET ?= $(if $(findstring true, $(MCD_ENABLED)),$(MCD_INTERNALAPI_NET),$(POD_INTERNALAPI_NET))
+NNCP_STORAGE_NET ?= $(if $(findstring true, $(MCD_ENABLED)),$(MCD_STORAGE_NET),$(POD_STORAGE_NET))
+NNCP_TENANT_NET ?= $(if $(findstring true, $(MCD_ENABLED)),$(MCD_TENANT_NET),$(POD_TENANT_NET))
+NNCP_DATA_NET ?= $(MCD_DATA_NET)
+## Others
 NNCP_TIMEOUT		?= 240s
 NNCP_CLEANUP_TIMEOUT	?= 120s
 NNCP_CTLPLANE_IP_ADDRESS_PREFIX     ?=192.168.122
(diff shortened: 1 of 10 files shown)


averdagu:
Need to test it on an actual environment, submitted PR for initial feedback

devsetup/Makefile
```diff
@@ -83,6 +84,41 @@ crc_cleanup: ## Destroys the CRC env, but does NOT clear ( --clear-cache ) the c
...
+	virsh --connect=qemu:///system attach-interface --domain crc --type network --source external --model virtio --mac $$MAC_ADDRESS --config;
+	sleep 1
+	virsh --connect=qemu:///system destroy crc
+	timeout --foreground ${TIMEOUT_VALUE} bash -c 'until $$(virsh --connect=qemu:///system list --all | grep crc | grep -q "shut off"); do sleep 1; done'
+	virsh --connect=qemu:///system start crc
+	timeout --foreground ${TIMEOUT_VALUE} bash -c 'until $$(virsh --connect=qemu:///system list --all | grep crc | grep -q "running"); do sleep 1; done'
+
+.PHONY: remove_nics_crc
+remove_nics_crc:
+	-MAC_ADDRESS=$(shell virsh --connect=qemu:///system dumpxml crc | grep "source network='data'" -B1 | grep mac | cut -d"'" -f2); \
```
booxter, line 108: this should probably be a loop?

devsetup/Makefile
```diff
@@ -83,6 +84,41 @@ crc_cleanup: ## Destroys the CRC env, but does NOT clear ( --clear-cache ) the c
 	sudo ${CLEANUP_DIR_CMD} /etc/pki/ca-trust/source/anchors/crc-router-ca.pem
 	sudo update-ca-trust
 
+.PHONY: add_nics_crc
+add_nics_crc: remove_nics_crc
+	# Add data interface
+	MAC_ADDRESS=$(shell echo -n 52:54:00; dd bs=1 count=3 if=/dev/random 2>/dev/null | hexdump -v -e '/1 "-%02X"' | tr '-' ':'); \
```
booxter, line 90: how did you define the mac prefix? should we put it in a variable?

devsetup/Makefile
```diff
@@ -83,6 +84,41 @@ crc_cleanup: ## Destroys the CRC env, but does NOT clear ( --clear-cache ) the c
...
+	virsh --connect=qemu:///system attach-interface --domain crc --type network --source data --model virtio --mac $$MAC_ADDRESS --config;
+	sleep 1
+	# Add management interface
+	MAC_ADDRESS=$(shell echo -n 52:54:00; dd bs=1 count=3 if=/dev/random 2>/dev/null | hexdump -v -e '/1 "-%02X"' | tr '-' ':'); \
```
booxter, line 94: this code should be a loop since it's nearly identical

devsetup/Makefile
```diff
@@ -83,6 +84,41 @@ crc_cleanup: ## Destroys the CRC env, but does NOT clear ( --clear-cache ) the c
...
+	-MAC_ADDRESS=$(shell virsh --connect=qemu:///system dumpxml crc | grep "source network='data'" -B1 | grep mac | cut -d"'" -f2); \
+	virsh --connect=qemu:///system detach-interface --domain crc --type network --mac $$MAC_ADDRESS --config
+	sleep 1
+	-MAC_ADDRESS=$(shell virsh --connect=qemu:///system dumpxml crc | grep "source network='management'" -B1 | grep mac | cut -d"'" -f2); \
+	virsh --connect=qemu:///system detach-interface --domain crc --type network --mac $$MAC_ADDRESS --config
+	sleep 1
+	-MAC_ADDRESS=$(shell virsh --connect=qemu:///system dumpxml crc | grep "source network='external'" -B1 | grep mac | cut -d"'" -f2); \
+	virsh --connect=qemu:///system detach-interface --domain crc --type network --mac $$MAC_ADDRESS --config
+	sleep 1
+	virsh --connect=qemu:///system destroy crc
```
booxter, line 117: you seem to repeat code here and above. should be a target `crc_restart`?

devsetup/Makefile
averdagu: The mac prefix is the same used by the target `crc_attach_default_interface`

devsetup/Makefile
averdagu: I find it harder than expected to create a loop on the Makefile itself where every iteration you need to assign a new variable.
For example, given the same code as add_crc_nics but using loop:
```
.PHONY: test_add_nics_crc
test_add_nics_crc: test_rem_nics_crc
        for network in data management external; do \
                MAC_ADDRESS=$(shell echo -n 52:54:00; dd bs=1 count=3 if=/dev/random 2>/dev/null | hexdump -v -e '/1 "-%02X"' | tr '-' ':'); \
                echo "virsh --connect=qemu:///system attach-interface --domain crc --type network --source $$network --model virtio --mac $$MAC_ADDRESS --config;"; \
                sleep 1; \
        done
```
When executed this will happen:
```

for network in data management external; do \
        MAC_ADDRESS=52:54:00:13:72:0E; \
        echo "virsh --connect=qemu:///system attach-interface --domain crc --type network --source $network --model virtio --mac $MAC_ADDRESS --config;"; \                                                                                  
        sleep 1; \
done
virsh --connect=qemu:///system attach-interface --domain crc --type network --source data --model virtio --mac 52:54:00:13:72:0E --config;                                                                                                   
virsh --connect=qemu:///system attach-interface --domain crc --type network --source management --model virtio --mac 52:54:00:13:72:0E --config;                                                                                             
virsh --connect=qemu:///system attach-interface --domain crc --type network --source external --model virtio --mac 52:54:00:13:72:0E --config;                                                                                               
```
I agree that both targets (add/remove nics) could be handled using a loop, but I guess the best way should be running an script

devsetup/Makefile
averdagu: This was on my scope, to set it as a target and only call it if some changes were made on the VM. I guess I'll handle this on a script

devsetup/Makefile
booxter: Yes. Which makes me think that maybe keeping it in a variable and then reusing it here. (And in other places) is a good idea.

devsetup/Makefile
booxter: A script would be nice I think

Makefile
```diff
@@ -1779,8 +1859,21 @@ nncp_cleanup: ## unconfigured nncp configuration on worker node and deletes the
...
+netattach: export STORAGE_VLAN=${NNCP_STORAGE_VLAN}
+netattach: export TENANT_VLAN=${NNCP_TENANT_VLAN}
+netattach: export INTERNALAPI_NET=${NNCP_INTERNALAPI_NET}
+netattach: export STORAGE_NET=${NNCP_STORAGE_NET}
+netattach: export TENANT_NET=${NNCP_TENANT_NET}
+netattach: export INTERNALAPI_IP_START=${NETATTACH_INTERNALIP_IP_START}
+netattach: export INTERNALAPI_IP_END=${NETATTACH_INTERNALIP_IP_END}
+netattach: export STORAGE_IP_START=${NETATTACH_STORAGE_IP_START}
+netattach: export STORAGE_IP_END=${NETATTACH_STORAGE_IP_END}
+netattach: export TENANT_IP_START=${NETATTACH_STORAGE_IP_START}
```
booxter, line 1875: I think there's a copy paste from STORAGE lines above here...

Makefile
```diff
@@ -1796,6 +1889,21 @@ netattach_cleanup: ## Deletes the network-attachment-definitions
...
+metallb: export STORAGE_VLAN=${NNCP_STORAGE_VLAN}
+metallb: export TENANT_VLAN=${NNCP_TENANT_VLAN}
+metallb: export INTERNALAPI_NET=${NNCP_INTERNALAPI_NET}
+metallb: export STORAGE_NET=${NNCP_STORAGE_NET}
+metallb: export TENANT_NET=${NNCP_TENANT_NET}
+metallb: export INTERNALAPI_IP_START=${METALLB_INTERNALIP_IP_START}
+metallb: export INTERNALAPI_IP_END=${METALLB_INTERNALIP_IP_END}
+metallb: export STORAGE_IP_START=${METALLB_STORAGE_IP_START}
+metallb: export STORAGE_IP_END=${METALLB_STORAGE_IP_END}
+metallb: export TENANT_IP_START=${METALLB_STORAGE_IP_START}
```
booxter, line 1905: same?

devsetup/scripts/crc_del_multi_interfaces.sh
```diff
@@ -0,0 +1,24 @@
...
+
+for network in data management external; do
+  MAC_ADDRESS=$(virsh --connect=qemu:///system dumpxml crc | grep "source network='$network'" -B1 | grep mac | cut -d"'" -f2);
+  if [ "x$MAC_ADDRESS" != "x" ]; then
+    NEED_REBOOT=1
+    virsh --connect=qemu:///system detach-interface --domain crc --type network --mac $$MAC_ADDRESS --config;
+  fi
+  sleep 1
+done
+if [[ $NEED_REBOOT == 1 ]]; then
```
booxter, line 18: You could probably have a separate script to reboot, then reuse it in both scripts for add and del

booxter (changes requested):

booxter:
I think this PR should include some documentation (in devsetup dir?) explaining how this should be used for multi node case.

Makefile
averdagu: Ah, you mean STORAGE_IP_[START|END]?
With this variables I mean the storage vlan.
http://dani.foroselectronica.es/openstack-tripleo-networking-layout-696/ (Vlan 30)

devsetup/scripts/crc_del_multi_interfaces.sh
averdagu: Agree, in order to not duplicate code, will do it in next patch :)

Makefile
booxter: I mean, why do you use STORAGE_IP_* for TENANT_IP_*? Are these not separate networks? I see in your link:

```
StorageNetCidr: 172.17.3.0/24
StorageNetworkVlanID: 30

TenantNetCidr: 172.17.2.0/24
TenantNetworkVlanID: 50
```

averdagu: Aahh, sorry, I'm blind. Indeed is a copy-paste error, will fix it :)

Makefile
averdagu: Indeed. will fix it on next patch

devsetup/scripts/crc_add_multi_interfaces.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if [ -z "$TIMEOUT" ]; then
+    echo "Please set TIMEOUT."
+    exit 1
+fi
+
+NEED_REBOOT=0
+
+for network in data management external; do
+  MAC_ADDRESS=$(echo -n 52:54:00; dd bs=1 count=3 if=/dev/random 2>/dev/null | hexdump -v -e '/1 "-%02X"' | tr '-' ':'); \
```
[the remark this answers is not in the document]
averdagu: I agree, both my script and this function does exactly the same, so it would be beneficial to call the script. Will do it on next patch.

scripts/gen-netatt.sh
```diff
@@ -15,30 +15,52 @@
...
+echo TENANT_NET ${TENANT_NET}
+echo INTERNALAPI_IP_START ${INTERNALAPI_IP_START}
+echo INTERNALAPI_IP_END ${INTERNALAPI_IP_END}
+echo STORAGE_IP_START ${STORAGE_IP_START}
+echo STORAGE_IP_END ${STORAGE_IP_END}
+echo TENANT_IP_START ${TENANT_IP_START}
+echo TENANT_IP_END ${TENANT_IP_END}
 
 cat > ${DEPLOY_DIR}/ctlplane.yaml <<EOF_CAT
 apiVersion: k8s.cni.cncf.io/v1
```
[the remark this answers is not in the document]
averdagu: I think there are different things.
In this case during the gen-netatt.sh the NetworkAttachmentDefinition resources are being created. This will determine the IPs that will have the pods that have an interface on any of those networks. So at the end this script will generate openshift resources.

The newtork-attachment-definition.sh is useful to modify the crc VM (but gen-netatt.sh will be used regardless if we use crc or dev-metal3 deployment)

PR #502, opened 2023-08-31
make standalone_snapshot wait after async device detach request

ashu-011:
The device detach before snapshot is asynchronous and if it doesn't complete by the time the snapshot command is executed, the snapshotting will fail. This change insert sleep 3 between these two. Closes https://github.com/openstack-k8s-operators/data-plane-adoption/issues/103


karelyatin (approved):
I hit the below issue and running same command later worked so seems this patch should help

$ make standalone_snapshot
virsh --connect=qemu:///system detach-device-alias edpm-compute-0 fs0 --live || true
Device detach request sent successfully

virsh --connect=qemu:///system snapshot-create-as --atomic --domain edpm-compute-0 --name clean
error: Requested operation is not valid: migration with virtiofs device is not supported

PR #504, opened 2023-09-04
Remove the edpm repo setup script.

steveb:
This replicates the `repo-setup` ansible service which is currently patched in [1] and is proposed to be a default service in the sample[2].

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-edpm-kustomize.sh#L94-L96
[2] https://github.com/openstack-k8s-operators/dataplane-operator/pull/383/files#diff-cd9f66e0beba929fa8eb93e256ecbd6b2c05922459cf15b0f648bab36e3fe8dfR58


devsetup/Makefile
```diff
@@ -156,8 +156,7 @@ edpm_compute: ## Create EDPM compute VM
 
 .PHONY: edpm_compute_repos
 edpm_compute_repos: ## Configure EDPM compute VM with repositories
-	$(eval $(call vars))
-	scripts/edpm-compute-repos.sh ${EDPM_COMPUTE_SUFFIX}
+	@echo "DEPRECATED: this target has no effect and will be removed. Repo setup is done by the 'repo-setup' ansible service"
```
steveb, line 159: Maybe this should also exit with an error, opinions welcome

steveb:
recheck openshift starts returning internal server errors

devsetup/Makefile
fao89: an error would be nice, it would catch the attention of people who are running it inside a script

slagle (commented):
What's the state of this given [e1c673a9d3919083b75ce80d871e37b2d9898b2d](https://github.com/openstack-k8s-operators/dataplane-operator/commit/e1c673a9d3919083b75ce80d871e37b2d9898b2d)  in dataplane-operator? We no longer want this PR here in install_yamls correct?

steveb:
> [e1c673a9d3919083b75ce80d871e37b2d9898b2d](https://github.com/openstack-k8s-operators/dataplane-operator/commit/e1c673a9d3919083b75ce80d871e37b2d9898b2d)

This was backing out a mistake on my part. The repo-setup service is specific to install_yamls, and is patched into the service list [here](https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-edpm-kustomize.sh#L72-L74), so this duplicate script can be removed

devsetup/Makefile
steveb: This can't happen yet though because CI jobs call this, so this will need to land first then a ci-framework change will need to stop calling this target

fao89 (approved):

PR #505, opened 2023-09-04
use tripleo-repos form current-tripleo on stable wallaby

SeanMooney:
This change updates the location where the python3-tripleo-repos
rpm is installed from to use the promoted content from the stable
wallaby line.

This corrects the fact the rpm is nolonger avaiable at
https://trunk.rdoproject.org/centos9/component/tripleo/current/
currenlty, ignoring the tempory issue this should also be more
stable as we are installing stable wallaby anyway and tripleo
master is nolonger used for development. stable/wallaby is the
maintained branch so we should align to that.

### DIFF
--- a/devsetup/scripts/standalone.sh
+++ b/devsetup/scripts/standalone.sh
@@ -45,7 +45,7 @@ set -ex
 sudo dnf remove -y epel-release
 sudo dnf update -y
 sudo dnf install -y vim git curl util-linux lvm2 tmux wget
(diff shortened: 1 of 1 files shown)


karelyatin (commented):
/lgtm tripleo projects are dropped from bobcat with https://review.rdoproject.org/r/c/rdoinfo/+/48758 so tripleo-repos rpm no longer a thing now ther, so it's right thing to use the rpm from wallaby

fao89 (approved):

jistr (approved):
Thanks

PR #515, opened 2023-09-06
Drop *_SINGLE_NODE vars and use only *_TOTAL_NODES

slagle:
The logic of these vars overlapped. Additionally, the handling of
$DATAPLANE_SINGLE_NODE relied on 2 nodes existing in the
dataplane-operator samples. Instead, just use $DATAPLANE_TOTAL_NODES to
configure the right number of nodes.


Makefile
```diff
@@ -572,7 +571,6 @@ openstack_deploy_cleanup: namespace netconfig_deploy_cleanup ## cleans up the se
 edpm_deploy_prep: export KIND=OpenStackDataPlaneNodeSet
 edpm_deploy_prep: export EDPM_ANSIBLE_SECRET=${DATAPLANE_ANSIBLE_SECRET}
 edpm_deploy_prep: export EDPM_ANSIBLE_USER=${DATAPLANE_ANSIBLE_USER}
-edpm_deploy_prep: export EDPM_SINGLE_NODE=${DATAPLANE_SINGLE_NODE}
 edpm_deploy_prep: export EDPM_COMPUTE_IP=${DATAPLANE_COMPUTE_IP}
 edpm_deploy_prep: export EDPM_COMPUTE_1_IP=${DATAPLANE_COMPUTE_1_IP}
```
fao89, line 575: we can drop this now

bshephar (approved):

fao89 (approved):
not holding the PR for a stale var

Makefile
slagle: Thanks. I fixed that in a follow up https://github.com/openstack-k8s-operators/install_yamls/pull/516

PR #518, opened 2023-09-07
Do not drop node ansibleVars

fao89:
https://github.com/openstack-k8s-operators/openstack-operator/pull/456#issuecomment-1709263844

### DIFF
--- a/scripts/gen-edpm-kustomize.sh
+++ b/scripts/gen-edpm-kustomize.sh
@@ -73,8 +73,9 @@ patches:
     - op: replace
       path: /spec/nodeTemplate/nodes/edpm-compute-0/ansible/ansibleHost
       value: ${EDPM_COMPUTE_IP}
-    - op: remove
-      path: /spec/nodeTemplate/nodes/edpm-compute-0/ansible/ansibleVars
+    - op: replace
+      path: /spec/nodeTemplate/nodes/edpm-compute-0/ansible/ansibleVars/ctlplane_ip
+      value: ${EDPM_COMPUTE_IP}
     - op: replace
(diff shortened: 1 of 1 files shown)


bshephar (approved):

scripts/gen-edpm-kustomize.sh
```diff
@@ -73,8 +73,9 @@ patches:
     - op: replace
       path: /spec/nodeTemplate/nodes/edpm-compute-0/ansible/ansibleHost
       value: ${EDPM_COMPUTE_IP}
-    - op: remove
-      path: /spec/nodeTemplate/nodes/edpm-compute-0/ansible/ansibleVars
+    - op: replace
+      path: /spec/nodeTemplate/nodes/edpm-compute-0/ansible/ansibleVars/ctlplane_ip
```
rabi, line 77: You don't need this if you're using IPAM and have networks list with fixedIP for controlplane[1]

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanenodeset_with_ipam.yaml#L47

fao89: yes, it was just a workaround: https://github.com/openstack-k8s-operators/openstack-operator/pull/456#issuecomment-1709263844
I'll revisit it once the openstack-operator gets green

rabi: Well it would never be required. That error was because for some strange reason we drop CtlPlane network in https://github.com/openstack-k8s-operators/dataplane-operator/commit/b51e247e58a104318662ca0ff8a42ca9c7eb6573#diff-e0d0ba0c862b888782b52f819dac0c0a570f1b3b5228d49ab2da27c9e86f752cR215-R220

PR #519, opened 2023-09-07
Remove requirement for 2 node default

bshephar:
This change implements similar logic to what is defined in the other EDPM gen script:
https://github.com/openstack-k8s-operators/install_yamls/commit/1b377bc39f91957e55e06a725fd3a4c0b95046c5

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -8,6 +8,7 @@
       - ^devsetup/scripts/bmaas/*
       - ^devsetup/scripts/edpm-compute-bmaas.sh
       - ^devsetup/scripts/gen-edpm-bmaas-kustomize.sh
+      - ^devsetup/scripts/gen-edpm-baremetal-kustomize.sh
       - ^devsetup/scripts/gen-ansibleee-ssh-key.sh
     irrelevant-files: &openstack_if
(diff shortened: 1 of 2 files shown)


zuul.d/jobs.yaml
```diff
@@ -9,6 +9,7 @@
       - ^devsetup/scripts/edpm-compute-bmaas.sh
       - ^devsetup/scripts/gen-edpm-bmaas-kustomize.sh
       - ^devsetup/scripts/gen-ansibleee-ssh-key.sh
+      - ^scripts/gen-edpm-baremetal-kustomize.sh
```
rabi, line 12: I think it would be `^devsetup/scripts/gen-edpm-baremetal-kustomize.sh`

fao89 (approved):

fao89:
as it depends on https://github.com/openstack-k8s-operators/openstack-operator/pull/456

rabi:
> @fao89: Overrode contexts on behalf of fao89: rdoproject.org/github-check

Why're we overriding and force merging everything? We can surely test https://github.com/openstack-k8s-operators/openstack-operator/pull/456 with Depends-On :/

fao89:
we were merging yesterday and I went with the flow, but you are right

PR #521, opened 2023-09-07
Set deployStrategy true by default

bshephar:


rabi:
You don't need to do this. This is set later on in https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml#L184

rabi (changes requested):
This is intentional in the job to wait for the BM nodes to be provisioned before we start deploying.

PR #523, opened 2023-09-07
Use ansible var edpm_network_config_template for setting nic config template

slagle:
Instead of setting networkConfig.temlate, which must be the template
contents, and a not a file name,  just set the ansibleVar
edpm_network_config_template with the file name.


cjeanner:
/lgtm

We have a proof of work on that other PR here:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/398

PR #524, opened 2023-09-08
Fix edpm_network_config_template ansible var

rabi:
This is a string and not a map.


cjeanner:
/lgtm

Are the various expected "types" documented anywhere?

slagle (approved):
Thanks.

It turns out this isn't actually used anywhere since ci_framework just deletes the whole key. Thanks for fixing it though.

slagle:
> Thanks.
> 
> It turns out this isn't actually used anywhere since ci_framework just deletes the whole key. Thanks for fixing it though.

I am incorrect about that. I see it's still used in the old nested job.

PR #526, opened 2023-09-11
Use antelope repos using repo-tool for EDPM deployment

raukadah:
Currently repo-setup current-podified-dev generates the master current and
current-podified delorean repos  and we are using antelope openstack services containers.

We are going to drop current-podified-dev interface from repo-setup as it get copied
from tripleo land. Developers can use current, podified-ci-testing and current-podified
tag.

It would be good to use antelope based repos and openstack service containers to
avoid issues during testing.


karelyatin (commented):
/lgtm

PR #534, opened 2023-09-12
Update for using OpenStackDataPlaneDeployment

slagle:
Deploying of the NodeSet is triggered with an
OpenStackDataPlaneDeployment CR instead of the deployStrategy field. The
OpenStackDataPlaneDeployment CR, configured with
$OPENSTACK_DATAPLANEDEPLOYMENT, and identified by
$DATAPLANE_DEPLOYMENT_CR, is copied into the $DEPLOY_DIR for
dataplane-operator resources.

It will be created during the edpm_deploy target when kustomize is run
against the $DEPLOY_DIR, triggering the deployment at the same step as
before.

The edpm kustomize scripts are updated to not set any fields under
deployStrategy, as that is no longer used.

See: https://github.com/openstack-k8s-operators/dataplane-operator/pull/378
Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/474

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -288,8 +288,12 @@ DATAPLANE_REPO                                   ?= https://github.com/openstack
 DATAPLANE_BRANCH                                 ?= main
 OPENSTACK_DATAPLANENODESET                       ?= config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml
 OPENSTACK_DATAPLANENODESET_BAREMETAL             ?= config/samples/dataplane_v1beta1_openstackdataplanenodeset_baremetal_with_ipam.yaml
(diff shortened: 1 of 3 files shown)


pablintino:
recheck

bshephar:
We will need to also remove `deployStrategy` from this file:
https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/gen-edpm-baremetal-kustomize.sh#L57-L59

Maybe our issue is coming from the fact that this other file is being used.

pablintino:
recheck

fao89:
recheck

fao89 (approved):

bshephar (approved):

PR #540, opened 2023-09-15
Replace networkConfig

bshephar:
As part of a wider effort to consolidate configuration around ansibleVars without unnecessary abstractions, this change moves the networkConfig template from a bespoke API abstraction via `spec.nodeTemplate.networkConfig.template` to the relevant Ansible variable: `edpm_network_config_template`.


slagle (commented):
@bshephar Can we close this PR? I don't think it's needed anymore.

PR #544, opened 2023-09-21
Update EDPM timeout

fao89:


devsetup/README.md
```diff
@@ -84,52 +84,7 @@ make download_tools
...
-make edpm_play_cleanup
-# Will delete VM's!:
-make edpm_compute_cleanup
-```
-
-In case additional compute node VMs are deployed, run:
-```
-make edpm_compute_cleanup EDPM_COMPUTE_SUFFIX=1
-```
+See [Deploy dev env using CRC, edpm nodes with isolated networks](../README.md#deploy-dev-env-using-crc-edpm-nodes-with-isolated-networks)
```
slagle, line 87: `../README.md` doesn't say anything about having to run `make edpm_compute_repos` using the devsetup Makefile.

slagle (changes requested):

devsetup/README.md
fao89: because it is not needed anymore, as the repo-setup is service is being added

fao89: I've removed this change, focus only on timeout for this PR

slagle (approved):

PR #548, opened 2023-09-25
Ensure only one cirros image gets created

weinimo:
When run multiple times the edpm-deploy-instance.sh would create multiple images named "cirros", which creates trouble later on.


weinimo:
I just noticed that other commands in this script also don't check if resources already exist before they create them. Any objection against adding checks for those commands too?

fao89:
it looks good to me, @raukadah could you please take a look?

fao89 (approved):

PR #557, opened 2023-09-29
EDPM node - additional networks / NICs

hjensas:
Make it possible to add additional networks (NICs) to the edpm nodes.

Use a JSON string containing a list of networks to add. Each element must define the keys ``type`` (libvirt network type) and ``name`` (libvirt network name).


hjensas:
recheck

steveb (approved):

fao89 (approved):

PR #562, opened 2023-10-02
devsetup - bmaas network - route libvirt networks

hjensas:
The CRC network, libvirt default network and the crc-bmaas network is configured in nat mode and the default libvirt firewall rules blocks traffic between the networks.

This change adds firewall rules in `CRC_BMAAS_FORWARD` and `CRC_BMAAS_POSTROUTING` chains so that traffic between the libvirt networks `crc`, `crc-bmaas` and `default` is routed without nat. Traffic to other networks are still behind nat.


hjensas:
Closing this - since the CRC is connected to the default libvirt network we end up with asynchronous routing issues ... Will have to take some other approach.

PR #564, opened 2023-10-04
devsetup - add scripts to route libvirt networks

hjensas:
By default libvirt does not route traffic between virtual networks in nat forwarding mode.

Add a script to create/cleanup routes between libvirt networks by adding rules based on input/output interface and source/destination ip subnets to FORWARD and POSTROUTING.

Add make targets to set up/tear down routing between the `crc`, `crc-bmaas` and `default` network. And include these targets in `bmaas` and `bmaas_cleanup`.

PR #565, opened 2023-10-05
devsetup - Standalone ironic resolv apps-crc.testing

hjensas:
To enable network connection to the virtual RedFish BMC running in crc configure the primary nameserver to use the crc address (192.168.130.11) when compute driver is ironic.

For this to work we need to set up routing for the libvirt networks, which is done in [PR# 564](https://github.com/openstack-k8s-operators/install_yamls/pull/564) and a host route must be added to the edpm standalone node. The route can be added in EDPM_COMPUTE_ADDITIONAL_NETWORKS as shown in below.

```bash
cat << EOF > /tmp/addtional_nets.json
[
  {
    "type": "network",
    "name": "crc-bmaas",
    "standalone_config": {
      "type": "linux_bridge",
      "name": "baremetal",
      "mtu": 1500,
      "ip_subnet": "172.20.1.0/24",
      "allocation_pools": [
        {
          "start": "172.20.1.4",
          "end": "172.20.1.250"
        }
      ],
      "host_routes": [
        {
	  "destination": "192.168.130.0/24",
          "nexthop": "172.20.1.1"
	}
      ]
    }
  }
]
EOF
export EDPM_COMPUTE_ADDITIONAL_NETWORKS=$(jq -c . /tmp/addtional_nets.json)
```

### DIFF
--- a/devsetup/scripts/standalone.sh
+++ b/devsetup/scripts/standalone.sh
@@ -138,6 +138,13 @@ while [[ $(ssh -o BatchMode=yes -o ConnectTimeout=5 $SSH_OPT root@$IP echo ok) !
 done
 
 # Render Jinja2 files
+if [ ${COMPUTE_DRIVER} == 'ironic' ]; then
+    PRIMARY_RESOLV_CONF_ENTRY="192.168.130.11"  # Use crc for dns to enable resolving apps-crc.testing
+    # Add a hosts entry for the network address to speed up iptables
+    ssh $SSH_OPT root@$IP "echo ${IP%.*}.0 .localdomain >> /etc/hosts"
+else
+    PRIMARY_RESOLV_CONF_ENTRY=${HOST_PRIMARY_RESOLV_CONF_ENTRY}
+fi
 J2_VARS_FILE=$(mktemp --suffix=".yaml" --tmpdir="${MY_TMP_DIR}")
 cat << EOF > ${J2_VARS_FILE}
 ---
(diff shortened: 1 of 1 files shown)


fao89 (approved):

devsetup/scripts/standalone.sh
```diff
@@ -138,6 +138,13 @@ while [[ $(ssh -o BatchMode=yes -o ConnectTimeout=5 $SSH_OPT root@$IP echo ok) !
 done
 
 # Render Jinja2 files
+if [ ${COMPUTE_DRIVER} == 'ironic' ]; then
+    PRIMARY_RESOLV_CONF_ENTRY="192.168.130.11"  # Use crc for dns to enable resolving apps-crc.testing
+    # Add a hosts entry for the network address to speed up iptables
+    ssh $SSH_OPT root@$IP "echo ${IP%.*}.0 .localdomain >> /etc/hosts"
```
jistr, line 144: If this /etc/hosts entry is the only thing we need to speed up iptables on Standalone, then perhaps we can use this always and go back to using `192.168.122.1` (gateway) as the DNS server always? At least on Fedora, deploying CRC will amend the systemd-resolved config so that it first proxies to the CRC host. (That's why iptables became slow in the standalone deployment.)

I think we'll need to resolve the apps-crc.testing names generally to connect the compute to the control plane anyway...

jistr: The whole "HOST_PRIMARY_RESOLV_CONF_ENTRY" dance was added to speed up iptables, it should not be needed for anything else.

hjensas: Yes this "hacky" entry in hosts solves the same problem with iptables reverse lookups slowness.

+1 to change it in a follow up, drop the `HOST_PRIMARY_RESOLV_CONF_ENTRY` and just add a similar entry in /etc/hosts always.

jistr (approved):

PR #568, opened 2023-10-09
Support for multi-stack openstack standalone

bogdando:
This is mostly required for testing a multi-stack (multi-cell) Nova compute adoption, but also might be useful for DCN adoption

After node0 deployed, extract required data for another
Heat stacks deployments.
See https://docs.openstack.org/project-deploy-guide/tripleo-docs/wallaby/deployment/standalone.html#deploy-the-remote-compute-node

Use that extracted data, when deploying consequent nodes by:
  make openstack EDPM_COMPUTE_SUFFIX=1

To make this working:
* Change the host name to match the extra nodes indexes (index 0 is 'standalone').
* Put env vars into an env file and scp it to the target standalone
   host, so that standalone-deploy.sh and openstack.sh could source
   it instead of repeating declaring the same vars (which used to
   not be working fine as overrides were getting lost in the process)
* export net-map-ip and fix netconfig:
  *  For multi-node/multi-stack standalone deployments,
      do not define VIPs for additional nodes in additional
      Heat stacks. Also, make them using RoleNetIpMap data
      exported from the "central" standalone stack.
* Install python-jinja2 for make standalone target
  before sourcing common funcs that do use of it.
                                                                                                                                                                                 
Unrelated change: rm delorean repos, before attempting to
install packages (fixes "dirty" envs with broken state 
after unfortunate triple-repos executions)

### DIFF
--- a/devsetup/scripts/standalone.sh
+++ b/devsetup/scripts/standalone.sh
@@ -35,6 +35,11 @@ SSH_OPT="-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i $SSH_KEY
 REPO_SETUP_CMDS=${REPO_SETUP_CMDS:-"${MY_TMP_DIR}/standalone_repos"}
 CMDS_FILE=${CMDS_FILE:-"${MY_TMP_DIR}/standalone_cmds"}
 SKIP_TRIPLEO_REPOS=${SKIP_TRIPLEO_REPOS:="false"}
+CLEANUP_DIR_CMD=${CLEANUP_DIR_CMD:-"rm -Rf"}
+EDPM_COMPUTE_VCPUS=${COMPUTE_VCPUS:-8}
+EDPM_COMPUTE_RAM=${COMPUTE_RAM:-20}
+EDPM_COMPUTE_DISK_SIZE=${COMPUTE_DISK_SIZE:-70}
+EDPM_COMPUTE_CEPH_ENABLED=${COMPUTE_CEPH_ENABLED:-true}
 
 if [[ ! -f $SSH_KEY_FILE ]]; then
     echo "$SSH_KEY_FILE is missing"
@@ -134,6 +139,7 @@ EOF
 fi
 
 while [[ $(ssh -o BatchMode=yes -o ConnectTimeout=5 $SSH_OPT root@$IP echo ok) != "ok" ]]; do
+    sleep 5
     true
 done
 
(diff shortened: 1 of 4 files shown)


bogdando:
this fails to deploy os-net-config for a 2nd stack as there is no br-ctlplane interface

hjensas:
What is the workflow for deploying mulit-node / multi-stack standalone?
Should it be separate make targets?

 make standalone  # Deploy the controller with VIPs
 make standalone_[cell?|secondary?]  # Deploy other nodes

In that case the second make target could export a boolean var an we can use that to determine if VIPs should be enabled etc?

devsetup/standalone/openstack.sh
```diff
@@ -92,7 +93,108 @@ if [ "$EDPM_COMPUTE_CEPH_ENABLED" = "true" ] ; then
...
+        cd $HEAT_LATEST
+    fi
+
+    openstack stack output show standalone EndpointMap --format json \
+    | jq '{"parameter_defaults": {"EndpointMapOverride": .output_value}}' \
+    > ${EDGE}_endpoint-map.json
+
+    openstack stack output show standalone EndpointMap --format json \
+    | jq '{"parameter_defaults": {"RoleNetIpMap": .output_value}}' \
+    > ${EDGE}_net-ip-map.json
```
bogdando, line 138: not sure if this is needed, if so, we should update the tripleo docs as well

https://docs.openstack.org/project-deploy-guide/tripleo-docs/wallaby/deployment/standalone.html#extract-deployment-information-from-the-controller-node

devsetup/scripts/standalone.sh
```diff
@@ -180,6 +202,18 @@ scp $SSH_OPT $HOME/.ssh/id_ecdsa.pub root@$IP:/root/.ssh/id_ecdsa.pub
 if [[ -f $HOME/containers-prepare-parameters.yaml ]]; then
     scp $SSH_OPT $HOME/containers-prepare-parameters.yaml root@$IP:/root/containers-prepare-parameters.yaml
 fi
+# For multi-stack deployment, get extracted data from main node index 0 and copy to the target node
+if [[ "$EDPM_COMPUTE_SUFFIX" != "0"  ]]; then
```
bogdando, line 206: these conditional blocks will be moved into different make standalone_stack targets

devsetup/standalone/openstack.sh
```diff
@@ -92,7 +93,108 @@ if [ "$EDPM_COMPUTE_CEPH_ENABLED" = "true" ] ; then
     CEPH_ARGS=${CEPH_ARGS:-"-e ~/deployed_ceph.yaml -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm-rbd-only.yaml"}
     ENV_ARGS+=" ${CEPH_ARGS}"
 fi
+
+# NOTE: For other nodes than node0, deploy it as extra Heat stacks,
```
bogdando, line 97: these conditional blocks will be moved into different make standalone_stack targets

PR #572, opened 2023-10-10
Add explicit exit 0 to standalone script

cescgina:
Explicitely return exit code 0 when we reach the end of the script,
if the script ends with a test chained to a command with an '&&' and
the test is not true, the script will return 0 and cause an error in CI


cescgina:
Testing in https://review.rdoproject.org/r/c/testproject/+/49765

jistr (approved):

PR #573, opened 2023-10-11
devsetup - handle if xmllint does not return an IP

fultonj:
Fixes issue from 071252c37b7bdfe5ae8eb88bdef8d77be300898b

On a RHEL8 hypervisor a firstboot script contained the full XML output of `virsh net-dumpxml` instead of an IP address. If a valid IP address is not returned from a call to virsh net-dumpxml and xmllint, then set EDPM_COMPUTE_NETWORK_IP to a default of 192.168.122.1 as a fallback.

PR #574, opened 2023-10-11
Fix xmllint read from stdin on RHEL8

hjensas:
On RHEL 8 the '-' to tell xmllint to read from stdion must be the last argument on the commandline for the xpath select pattern to work.

```bash
$ virsh net-dumpxml default | xmllint --xpath 'string(/network/ip/@address)' -
192.168.124.1
```
^^ Works with '-' as last arg.

```bash  
$ virsh net-dumpxml default | xmllint - --xpath 'string(/network/ip/@address)'
<?xml version="1.0"?>
<network>
  <name>default</name>
  <ip address="192.168.124.1" netmask="255.255.255.0">
<% - SNIP - %>
  </ip>
</network>
```
^^ Fails with '-' as first arg.

This should fix the issue described in https://github.com/openstack-k8s-operators/install_yamls/pull/573


abays (approved):
This is the format I would expect.  We use it elsewhere, i.e. `oc kustomzie <some dir> | oc apply -f -` [1].

/lgtm

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/441eca4d115e1a153fe2041dff664ba578f2a5a5/scripts/operator-deploy-resources.sh#L31

PR #575, opened 2023-10-11
Replacing edpm_chrony_ntp_servers with timesync_ntp_servers

jpodivin:
The edpm_chrony role is getting removed, this variable has to be renamed in order to reflect the new reality.

Depends-On: openstack-k8s-operators/edpm-ansible#373


scripts/gen-edpm-baremetal-kustomize.sh
```diff
@@ -53,7 +53,7 @@ patches:
       path: /spec/services/0
       value: repo-setup
     - op: replace
-      path: /spec/nodeTemplate/ansible/ansibleVars/edpm_chrony_ntp_servers
+      path: /spec/nodeTemplate/ansible/ansibleVars/edpm_ntp_servers
       value:
         - ${EDPM_CHRONY_NTP_SERVER}
```
fultonj, line 58: LGTM except one thing.

On line 58 I'd replace this:
```
  - ${EDPM_CHRONY_NTP_SERVER}
```
with this
```
  - {hostname: ${EDPM_CHRONY_NTP_SERVER}}
```
Same for line 76. 

That way you don't need unnecessary extra complexity to append the hostname before the call to `linux-systemroles.timesync`. Also, people familiar with the `linux-systemroles.timesync` (since it's becoming a standard within Red Hat) can pass what they know. Also, there's no point in maintaining a semblance of backwards compatibility with the old way as it hasn't really shipped and for today's users, install_yamls hides it. I say a semblance because it's not really backwards compatible as the variable name is changing.

WDYT @jpodivin @rebtoor ?

jpodivin: If there is really no need for backwards compatibility I can do that. Although I would at least like to keep the variable name, so it's obvious what kind of time servers it's setting up.

Makefile
```diff
@@ -330,7 +330,7 @@ DATAPLANE_TOTAL_NODES                            ?=1
 DATAPLANE_RUNNER_IMG                             ?=quay.io/openstack-k8s-operators/openstack-ansibleee-runner:${OPENSTACK_K8S_TAG}
 DATAPLANE_NETWORK_INTERFACE_NAME                 ?=eth0
 DATAPLANE_SSHD_ALLOWED_RANGES                    ?=['192.168.122.0/24']
-DATAPLANE_CHRONY_NTP_SERVER                      ?=pool.ntp.org
+DATAPLANE_NTP_SERVER                      ?=pool.ntp.org
```
rebtoor, line 333: can you fix indentation here?

scripts/gen-edpm-baremetal-kustomize.sh
rebtoor: Why we should keep backward compatibility? I mean, the old role will be deleted, i think it's better to keep 1:1 compatibility with systemrole variables, as @fultonj said.

fultonj (commented):
I'm +2 on this provided all concerns below are addressed.

https://github.com/openstack-k8s-operators/edpm-ansible/pull/373#pullrequestreview-1694636108

rebtoor (approved):

rebtoor (approved):

PR #576, opened 2023-10-11
Adding a NetworkAttachment for the designate services to support

dkehn:
their necessity to communicate with DNS servers.


dkehn:
recheck

dprince:
So I think 'DesignateApiNetwork' defaults to internal_api there. So do we need a custom network for designate here?

dprince (approved):

PR #579, opened 2023-10-12
Create dataplane services in dataplane_kuttl_prep

rabi:
We want to get rid of the reconciliation of services in nodeset controller.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1474,6 +1474,8 @@ dataplane_kuttl_prep: dataplane_kuttl_cleanup
 	devsetup/scripts/gen-ansibleee-ssh-key.sh
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR}
 	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(DATAPLANE_BRANCH),-b ${DATAPLANE_BRANCH}) ${DATAPLANE_REPO} "${OPERATOR_NAME}-operator" && popd
+	DEPLOY_DIR=${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services KIND=OpenStackDataPlaneService bash scripts/gen-edpm-services-kustomize.sh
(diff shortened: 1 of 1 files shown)


rabi:
recheck 

fatal: [localhost]: FAILED! => {\"changed\": false, \"msg\": \"Failure downloading https://golang.org/dl/go1.19.10.linux-amd64.tar.gz, The read operation timed out\"}",

Makefile
```diff
@@ -1474,6 +1474,8 @@ dataplane_kuttl_prep: dataplane_kuttl_cleanup
 	devsetup/scripts/gen-ansibleee-ssh-key.sh
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR}
 	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(DATAPLANE_BRANCH),-b ${DATAPLANE_BRANCH}) ${DATAPLANE_REPO} "${OPERATOR_NAME}-operator" && popd
+	DEPLOY_DIR=${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services KIND=OpenStackDataPlaneService bash scripts/gen-edpm-services-kustomize.sh
+	oc kustomize ${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services | oc apply -f -
```
slagle, line 1478: I don't think this should be a user required task. Where are they going to get the YAML files from? I'm asking what the product documented steps would look like. I'd prefer to leave the service reconciliation in dataplane-operator.

slagle (changes requested):

Makefile
gibizer: I foresee that some of the service CRs that reconciled by the dataplane-operator are won't be used by the user. E.g. the Service/nova will be created manually to pass extra config snippets like cpu_shared_set or [pci]device_spec to the nova-compute config. I foresee similar user customization needs in other openstack services as well. So I think some of the service where we don't expect any user configuration by default can remain under the dataplane-operator reconciliation, but the ones that almost always needs user input does not make sense to reconciled by the dataplane-operator.

PR #582, opened 2023-10-12
devsetup - download_tools - add pkg python3-jinja2

hjensas:
Install the python3-jinja2 package as part of the download_tools ansible role in devsetup so that standalone.sh does not fail with import error.


lewisdenny:
recheck
Due to a CI Blocker

bogdando (approved):

fao89 (approved):

PR #583, opened 2023-10-12
devsetup - Standalone ironic - Clean "metadata"

hjensas:
Set IronicCleaningDiskErase to "metadata" in the Standalone deployment. The default "erase" takes a lot longer, and there is no need to use full erase in an adoption test env.


lewisdenny:
recheck
Due to a CI Blocker

PR #584, opened 2023-10-13
Modifications for BGP deployment

luis5tb:
This is to automate most of the steps to deploy in an env that is also connected to a spine-leaf setup (with 2 ToR for each compute)

### DIFF
--- a/devsetup/scripts/gen-edpm-compute-node-bgp.sh
+++ b/devsetup/scripts/gen-edpm-compute-node-bgp.sh
@@ -1,0 +1,267 @@
+#!/bin/bash
+#
+# Copyright 2022 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+set -ex
+export LIBVIRT_DEFAULT_URI=qemu:///system
+# expect that the common.sh is in the same dir as the calling script
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+CRC_POOL=${CRC_POOL:-"$HOME/.crc/machines/crc"}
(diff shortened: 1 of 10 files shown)


devsetup/scripts/bgp-interfaces-setup.sh
```diff
@@ -0,0 +1,57 @@
...
+  <address type='pci' domain='0x0000' bus='0x00' slot='0x02' function='0x7'/>
+</controller>
+EOF
+
+cat << EOF > /tmp/net1.xml
+<interface type='network'>
+  <source network='default'/>
+  <model type='virtio'/>
+  <address type='pci' domain='0x0000' bus='0x06' slot='0x00' function='0x0'/>
+</interface>
```
stuggi, line 33: should this be the same as in https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/Makefile#L107 ? would it make sense to make this script generic to run by default what we currently have in the Makefile and in the BGP case also add the other?

devsetup/scripts/gen-edpm-compute-node-bgp.sh
```diff
@@ -0,0 +1,266 @@
...
+EDPM_COMPUTE_SUFFIX=${1:-"0"}
+EDPM_COMPUTE_NAME=${EDPM_COMPUTE_NAME:-"edpm-compute-${EDPM_COMPUTE_SUFFIX}"}
+if [ ${STANDALONE} = "true" ]; then
+    EDPM_COMPUTE_VCPUS=${EDPM_COMPUTE_VCPUS:-8}
+    EDPM_COMPUTE_RAM=${EDPM_COMPUTE_RAM:-20}
+    EDPM_COMPUTE_DISK_SIZE=${EDPM_COMPUTE_DISK_SIZE:-70}
+fi
+EDPM_COMPUTE_VCPUS=${EDPM_COMPUTE_VCPUS:-8}
+EDPM_COMPUTE_RAM=${EDPM_COMPUTE_RAM:-24}
+EDPM_COMPUTE_DISK_SIZE=${EDPM_COMPUTE_DISK_SIZE:-60}
```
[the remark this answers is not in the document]
luis5tb: right, I'll use the defaults one here

devsetup/scripts/bgp-interfaces-setup.sh
luis5tb: I think the problem we had was that we needed to add extra pci slots. You are suggesting to add the pci slots in the BGP case too, right?

stuggi: yes, add the pci slots in case of bgp, but not in the default one

stuggi (approved):
/lgtm

PR #587, opened 2023-10-16
Revert "Configure ip on crc baremetal bridge"

hjensas:
The actual result of this is a `- ip: null` is set on the crc-bmaas interface becuse the BMAAS_BRIDGE_IPADDRESS is not exported when calling baremetal-bridge.sh.

I don't think we need this, provisioning with bmo works without it.

This reverts commit 352c3b8d14681c77ef987dc24c2a280b210df41f.


rabi (approved):
/lgtm

yeah we don't use `crc-bmaas` interface for baremetal provisioning anymore. This is not required.

stuggi (approved):
/lgtm

PR #588, opened 2023-10-16
Rename NAD for BMAAS: s/baremetal-net/baremetal/

hjensas:
When trying to attach to OVN there is an error when creating the ovs bridge:
```
  ovs-vsctl: Error detected while setting up
   'br-baremetal-net': could not add network device
   "br-baremetal-net to ofproto (Invalid argument).
   See ovs-vswitchd log for details.
```
    
Let's rename it, so that it fits in the 15 char limit for interface name when ovn-operator adds the 'br-' prefix.

### DIFF
--- a/devsetup/scripts/bmaas/network-attachement-definition.sh
+++ b/devsetup/scripts/bmaas/network-attachement-definition.sh
@@ -12,8 +12,8 @@ NETWORK_NAME=${NETWORK_NAME:-"$DEFAULT_NETWORK_NAME"}
 function usage {
     echo
     echo "options:"
-    echo "  --create        Create baremetal-net NetworkAttachmentDefinition"
-    echo "  --cleanup       Delete baremetal-net NetworkAttachmentDefinition"
+    echo "  --create        Create barametal NetworkAttachmentDefinition"
+    echo "  --cleanup       Delete barametal NetworkAttachmentDefinition"
     echo
 }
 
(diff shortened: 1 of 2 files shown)


devsetup/scripts/bmaas/network-attachement-definition.sh
```diff
@@ -29,13 +29,13 @@ function create {
 apiVersion: k8s.cni.cncf.io/v1
 kind: NetworkAttachmentDefinition
 metadata:
-  name: baremetal-net
+  name: baremetal
   namespace: openstack
```
karelyatin, line 33: can rename in Line 15 and 16 too.

karelyatin:
> use

Actually the reason here is not dash in the name but interface name limit to 15 characters which surpasses when ovn creates a bridge by prefixing br-.

hjensas:
> > use
> 
> Actually the reason here is not dash in the name but interface name limit to 15 characters which surpasses when ovn creates a bridge by prefixing br-.

ah, ok I will update the commit message to say that as well.

devsetup/scripts/bmaas/network-attachement-definition.sh
hjensas: Done, thanks!

karelyatin:
> > > use
> > 
> > 
> > Actually the reason here is not dash in the name but interface name limit to 15 characters which surpasses when ovn creates a bridge by prefixing br-.
> 
> ah, ok I will update the commit message to say that as well.
> 
> Is there an actual reason for ovn-operator to add the 'br-' prefix? It could just use the physical network name defined in the mapping without adding a prefix?

I don't think there is a reason from functional point but seems it was done for consistency/historical usage, but @slawqo may know more.

karelyatin (commented):
/lgtm

stuggi (approved):
/lgtm

PR #589, opened 2023-10-17
[DNM] Add a sleep after standalone vm creation

cescgina:
To workaround some indentified network issue in the extracted crc
adoption job, wait for a few minutes until the network is functional.
Only when the SLEEP_TIME variable is set, the sleep should be executed.


cescgina:
Closing this as it doesn't really solve the problem and found a better solution elsewhere.

PR #591, opened 2023-10-19
bmaas add MetalLB L2Advertisement + Address pool

hjensas:
When testing ironic it may make sense to expose the internal endpoint for Ironic on the baremetal
provisioning network.

This change add's make target bmaas_metallb which creates the IPAddressPool and L2Advertisement resources.

Also some cleanup and docs changes:
* Move the execution the nmstate target to the bmaas_crc_baremetal_bridge target.
* Change the execution of input target to instead run the namespace target, and run this with the bmaas_baremetal_net_nad target.
* Mention that MetalLB is configured in doc's and add a table descriping how the 172.20.1.0/24 subnet is divided into pools.

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -42,6 +42,8 @@ BMAAS_INSTANCE_NET_MODEL ?= virtio
 BMAAS_NETWORK_NAME ?= crc-bmaas
 BMAAS_NETWORK_IPADDRESS ?= 172.20.1.1
 BMAAS_NETWORK_NETMASK ?= 255.255.255.0
+BMAAS_METALLB_POOL_NAME ?= baremetal
+BMAAS_METALLB_ADDRESS_POOL ?= 172.20.1.64/26
 BMAAS_NODE_COUNT ?= 1
 BMAAS_REDFISH_USERNAME ?= admin
 BMAAS_REDFISH_PASSWORD ?= password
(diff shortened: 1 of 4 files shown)


hjensas:
recheck

It looks like [data-plane-adoption-github-rdo-centos-9-crc-single-node ](https://review.rdoproject.org/zuul/build/bfa1d338b2664c02b7da94bf23521cbb) job don't exercise the patched codepath.

stuggi (approved):
looks good to me

devsetup/Makefile
```diff
@@ -273,12 +276,26 @@ bmaas_crc_baremetal_bridge_cleanup:
...
 .PHONY: bmaas_baremetal_net_nad_cleanup
 bmaas_baremetal_net_nad_cleanup:
 	scripts/bmaas/network-attachement-definition.sh --cleanup
 
+.PHONY: bmaas_metallb
+bmaas_metallb: export NETWORK_NAME = ${BMAAS_NETWORK_NAME}
+bmaas_metallb: export POOL_NAME = ${BMAAS_METALLB_POOL_NAME}
+bmaas_metallb: export ADDRESS_POOL = ${BMAAS_METALLB_ADDRESS_POOL}
+bmaas_metallb:
+	pushd .. && make metallb && popd
```
rabi, line 291: Sorry I'm  a little late to review this. But I think we should avoid these cross Makefile target dependencies.

hjensas: I don't mind removing them, and instead document that you need to do `make metallb` manually as a pre-requisite to using the `bmaas_metallb` target. We can add a check to the script and fail `bmaas_metallb` target with a good error message instead. I went this way, because we already did it this way for `input`/`namespace` and `nmstate` targets.

PR #600, opened 2023-10-24
Ensure DataplaneServices gets the expected image

fao89:


scripts/gen-edpm-services-kustomize.sh
```diff
@@ -56,3 +56,8 @@ EOF
 kustomization_add_resources
 
 popd
+
+oc patch csv -n openstack-operators openstack-ansibleee-operator.v0.0.1 \
+    --type='json' -p='[{
+    "op":"replace", "path":"/spec/install/spec/deployments/0/spec/template/spec/containers/1/env/0",
+    "value": {"name": "RELATED_IMAGE_ANSIBLEEE_IMAGE_URL_DEFAULT", "value": "'${OPENSTACK_RUNNER_IMG}'"}}]'
```
slagle, line 63: Instead of adding this here, maybe it should just be in the Makefile? Or this should be the only thing in this script, and it ought to be renamed.

Point being, creating the services kustomization and then oc applying the directory of services does not work, b/c the NodeSet controller will reconcile them back to their original value.

Makefile
```diff
@@ -651,7 +651,10 @@ edpm_deploy_prep: edpm_deploy_cleanup ## prepares the CR to install the data pla
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR} ${DEPLOY_DIR}
 	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(DATAPLANE_BRANCH),-b ${DATAPLANE_BRANCH}) ${DATAPLANE_REPO} "${OPERATOR_NAME}-operator" && popd
 	cp devsetup/edpm/services/* ${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services
-	DEPLOY_DIR=${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services KIND=OpenStackDataPlaneService bash scripts/gen-edpm-services-kustomize.sh
+	oc patch csv -n openstack-operators openstack-ansibleee-operator.v0.0.1 \
```
slagle, line 654: we probably need to discover the csv name here, instead of hardcoding the version

slagle (approved):

fao89:
recheck

fao89:
recheck

Makefile
```diff
@@ -716,8 +718,10 @@ edpm_deploy_baremetal_prep: edpm_deploy_cleanup ## prepares the CR to install th
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR} ${DEPLOY_DIR}
 	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(DATAPLANE_BRANCH),-b ${DATAPLANE_BRANCH}) ${DATAPLANE_REPO} "${OPERATOR_NAME}-operator" && popd
 	cp devsetup/edpm/services/* ${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services
-	DEPLOY_DIR=${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services KIND=OpenStackDataPlaneService bash scripts/gen-edpm-services-kustomize.sh
-	oc kustomize ${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services | oc apply -f -
+	oc apply -f ${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services
+	oc patch $(shell oc get csv -n openstack-operators -o name | grep ansibleee) \
```
rabi, line 722: We can't do this as it as this would result in jobs not using the images built with CR in CI.

PR #606, opened 2023-10-26
Add KUTTL make targets for telemetry

paramite:
Depends-On: https://github.com/openstack-k8s-operators/telemetry-operator/pull/213

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -371,6 +371,10 @@ TELEMETRY_CR                     ?= ${OPERATOR_BASE_DIR}/telemetry-operator/${TE
 CEILOMETER_CENTRAL_DEPL_IMG      ?= unused
 CEILOMETER_NOTIFICATION_DEPL_IMG ?= unused
 SG_CORE_DEPL_IMG                 ?= unused
+TELEMETRY_KUTTL_BASEDIR   ?= ${OPERATOR_BASE_DIR}/telemetry-operator
+TELEMETRY_KUTTL_RELPATH   ?= tests/kuttl/suites
+TELEMETRY_KUTTL_CONF      ?= ${TELEMETRY_KUTTL_BASEDIR}/kuttl-test.yaml
+TELEMETRY_KUTTL_NAMESPACE ?= telemetry-kuttl-tests
 
 # BMO
 BMO_REPO                         ?= https://github.com/metal3-io/baremetal-operator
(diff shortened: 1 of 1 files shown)


paramite:
recheck

paramite:
recheck

paramite:
recheck

paramite:
recheck

abays:
recheck

abays:
/override data-plane-adoption-github-rdo-centos-9-crc-single-node

Makefile
```diff
@@ -2026,6 +2029,19 @@ telemetry_deploy_cleanup: ## cleans up the service instance, Does not affect the
...
 
+.PHONY: telemetry_kuttl_run
+telemetry_kuttl_run: ## runs kuttl tests for the telemetry operator, assumes that everything needed for running the test was deployed beforehand.
+	kubectl-kuttl test --config ${TELEMETRY_KUTTL_CONF} ${TELEMETRY_KUTTL_DIR} --namespace ${NAMESPACE}
+
+.PHONY: telemetry_kuttl
+telemetry_kuttl: export NAMESPACE = ${TELEMETRY_KUTTL_NAMESPACE}
+# Set the value of $TELEMETRY_KUTTL_NAMESPACE if you want to run the telemetry
+# kuttl tests in a namespace different than the default (telemetry-kuttl-tests)
+telemetry_kuttl: input deploy_cleanup telemetry telemetry_deploy_prep
```
abays, line 2040: If Telemetry operator depends on other operators (such as MariaDB, Keystone, etc), then you'll need to address those dependencies here as pre-targets.  i.e. `kuttl_common_prep`.  For example [1].

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/c0a3767928bc188fcd9af004e977c55b6f2d251d/Makefile#L1611

Makefile
```diff
@@ -2026,6 +2029,19 @@ telemetry_deploy_cleanup: ## cleans up the service instance, Does not affect the
...
+	make wait
```
cescgina, line 2041: seems like it's missing ` $(eval $(call vars,$@,telemetry))` before this line to run correctly

cescgina (commented):
I tried running the tests, but while it finds the `autoscaling` test suite, it does not run any tests because the test files are inside a test folder, but kuttl expects them inside `autoscaling`. So, `path/tests/suites/autoscaling/01-assert.yaml` instead of `path/tests/suites/autoscaling/tests/01-assert.yaml`

paramite:
Thanks for hints. Fixed dependencies (added kuttl_common_prep and heat) and added testDirs to config, because telemetry-operator gonna have more than one test suite.

paramite:
Currently tests work, just timeouts too early during Prometheus init. Patch in operator [1] is already submitted.

[1] https://github.com/openstack-k8s-operators/telemetry-operator/pull/220

cescgina (commented):
Changes lgtm, however I could not get the tests to pass unless the namespace fields in the assert file was removed. Looks like the assert is expecting the resources to be deployed in a namespace called `telemetry-kuttl-autoscaling`, apply this commit the tests passed for me: https://github.com/cescgina/telemetry-operator/commit/a2d693c5638fbd91f40d2ae005f7b56a39773e86

paramite:
Ah, that's good point. Thanks for that. Do you mind submitting PR with this?

paramite:
Your PR is merged. So what now? Do I need to do something else with this PR?

cescgina:
> Your PR is merged. So what now? Do I need to do something else with this PR?

PR lgtm, IMO it can be merged now, but I don't have approver rights, @abays can you take a look when free?

PR #609, opened 2023-11-03
Support using manila/native cephfs with standalone

gouthampacha:
This is useful for a adoption test/dev environment.


gouthampacha:
recheck

unsure how this would cause a failure to deploy/adopt heat: 

https://review.rdoproject.org/zuul/build/d0ea74c8cbb246f38c21e1a5fbad237e/log/controller/data-plane-adoption-tests-repo/data-plane-adoption/tests/logs/test_with_ceph_out_2023-11-03T03:22:29EDT.log#392-393

fmount:
@marios @jistr is that ok to enable manila by default in standalone? this will help to support by default the manila adoption in the testsuite (this assumes we have Ceph as a backend)

jistr:
I don't feel strongly one way or another. On one hand i don't think it's a problem, on the other hand we'll have to draw the line somewhere regarding what we deploy in the "all in one" and what should go into per-DFG jobs... I'm thinking we could perhaps do this and if we hit stability trouble with too many services being tested in a single job, we may try to walk it back and split into per-service jobs? But my vote should be weak in matters of setting CI patterns.

Curious what @marios and @cescgina think :)

fmount:
> I don't feel strongly one way or another. On one hand i don't think it's a problem, on the other hand we'll have to draw the line somewhere regarding what we deploy in the "all in one" and what should go into per-DFG jobs... I'm thinking we could perhaps do this and if we hit stability trouble with too many services being tested in a single job, we may try to walk it back and split into per-service jobs? But my vote should be weak in matters of setting CI patterns.
> 
> Curious what @marios and @cescgina think :)

Thanks @jistr for your thoughts.
We already added a `MANILA_ENABLED` bool, so it would be easy to create jobs w/o Manila if required or if we're going to hit issues w/ the standalone deployment.

jistr:
For completeness i'll mention here that we just discussed this at the Adoption call and the consensus was in the spirit of "let's just include it and in case it causes any issues we can revisit the decision".

cescgina:
> I don't feel strongly one way or another. On one hand i don't think it's a problem, on the other hand we'll have to draw the line somewhere regarding what we deploy in the "all in one" and what should go into per-DFG jobs... I'm thinking we could perhaps do this and if we hit stability trouble with too many services being tested in a single job, we may try to walk it back and split into per-service jobs? But my vote should be weak in matters of setting CI patterns.
> 
> Curious what @marios and @cescgina think :)

I'm fine with this change, if problematic it's easy enough to turn off, and CI does not seem to have a problem. One thing to keep in mind is that the standalone vm does not have a ton of resources, being nested. I don't think this will be a problem, since we're currently switching to a different layout without any nested vms.

fmount (approved):
/lgtm

PR #612, opened 2023-11-06
[dev-preview2] Run extracted crc adoption job in PRs

cescgina:
Move to use the extracted crc adoption job instead of single-node one. Create a new child job, to set openstack-operator as required project. Cherry picked from commit ae39de3dc4033

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -19,3 +19,9 @@
       - LICENSE
       - OWNERS
       - .*/*.md
+- job:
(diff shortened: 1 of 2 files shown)


fao89:
recheck

cescgina:
Closing this as it's not really needed. I opened it because CI was failing in a dev-preview2 backport, but it turned out to be targeting the wrong branch. The dev-preview2 content is out of sync with data-plane-adoption main, as evindence in the test failure where the wrong label is used to check the openstackclient pod. I don't think it's worth it at this point to fix that (and maybe other issues) just to run data-plane adoption in dev-preview2.

PR #614, opened 2023-11-07
NAT64 - SNO (Single-node-OpenShift) libvirt lab

hjensas:
This add's scripts to set up NAT64+DNS64 to enable deploying an IPv6 only lab environment which will work for users that don't have actual IPv6 internet available.

### DIFF
--- a/devsetup/README.md
+++ b/devsetup/README.md
@@ -294,3 +294,42 @@ nodes:
   ports:
   - address: 52:54:00:8a:ea:14
 ```
+
+### IPv6 LAB
+
+#### Create the IPv6 LAB
+
+Export vars:
+```bash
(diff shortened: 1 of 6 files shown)


devsetup/README.md
```diff
@@ -294,3 +294,62 @@ nodes:
...
+```bash
+oc login -u admin -p 12345678 https://api.sno.lab.example.com:6443
+```
+
+Change to the install_yamls root directory:
+```bash
+cd ..
+```
+
+```bash
```
hjensas, line 342: The readme is not complete, but I think we can merge and update in follow ups.

hjensas:
recheck

stuggi (approved):
/lgtm

PR #616, opened 2023-11-09
Condition libvirt service restart

hjensas:
On later Fedora and EL/Centoes Stream 9 uses modular libvirt, in this case we should not start|restart libvirtd.service.

Condition the restart based on "is-enabled" for the service. If libvirtd.service is disbled, instead restart virtqemud.service.

Related: https://github.com/openstack-k8s-operators/ci-framework/issues/751


hjensas:
recheck

hjensas:
recheck

fao89 (approved):

PR #617, opened 2023-11-09
[BGP] Ensure the loopback can be reconfigured

luis5tb:
It seems even though the lo state is UNKNOWN, the proper nncp configuration is to set it to 'state: up' so that new configurations are processed


stuggi (approved):
/lgtm

PR #621, opened 2023-11-10
Avoid overwriting metallb's mp-bgp route-maps

luis5tb:
There are 2 route maps created by metallbm, one for ipv4 and one for ipv6. And the previous configuration was overwritting the one for ipv4.


stuggi (approved):
/lgtm

PR #622, opened 2023-11-10
Parameterize "attach_default_interface" targets

hjensas:
The target for 'crc_attach_default_interface' is currently hard-coded to the crc instance and the default network.

Refactor to use parameters for instance name and network to attach to.

Also, libvirt does not allow using "mac" for ipv6 ip-dhcp-host - so change the reservation to use instance name only.

### DIFF
--- a/devsetup/scripts/interfaces-setup.sh
+++ b/devsetup/scripts/interfaces-setup.sh
@@ -8,7 +8,7 @@ fi
 
 if [ -n "$BGP" ]; then
     # We cannot add PCI slots with the VM running
-    sudo virsh destroy crc
+    sudo virsh destroy $INSTANCE_NAME
 
     cat << EOF > /tmp/pci1.xml
 <controller type='pci' index='7' model='pcie-root-port'>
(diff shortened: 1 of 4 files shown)


abays:
recheck

abays:
I saw this after I installed a fresh CRC in an environment that had previously used `install_yamls` and the old `make crc_attach_default_interface`:

```
$ make attach_default_interface
bash scripts/interfaces-setup-cleanup.sh
bash scripts/interfaces-setup.sh

error: Failed to update network default
error: Requested operation is not valid: there is an existing dhcp host entry in network 'default' that matches "<host mac='unknown' name='crc' ip='192.168.122.10'/>"
```

I then did `make attach_default_interface_cleanup` and then re-ran `make attach_default_interface` and the message no longer appeared.

hjensas:
> I saw this after I installed a fresh CRC in an environment that had previously used `install_yamls` and the old `make crc_attach_default_interface`:
> 
> ```
> $ make attach_default_interface
> bash scripts/interfaces-setup-cleanup.sh
> bash scripts/interfaces-setup.sh
> 
> error: Failed to update network default
> error: Requested operation is not valid: there is an existing dhcp host entry in network 'default' that matches "<host mac='unknown' name='crc' ip='192.168.122.10'/>"
> ```
> 
> I then did `make attach_default_interface_cleanup` and then re-ran `make attach_default_interface` and the message no longer appeared.

hmm, I will do some more testing.
My intent was that the old `crc_attach_default_interface` targets should still work for compatibility.

hjensas:
> > I saw this after I installed a fresh CRC in an environment that had previously used `install_yamls` and the old `make crc_attach_default_interface`:
> > ```
> > $ make attach_default_interface
> > bash scripts/interfaces-setup-cleanup.sh
> > bash scripts/interfaces-setup.sh
> > 
> > error: Failed to update network default
> > error: Requested operation is not valid: there is an existing dhcp host entry in network 'default' that matches "<host mac='unknown' name='crc' ip='192.168.122.10'/>"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I then did `make attach_default_interface_cleanup` and then re-ran `make attach_default_interface` and the message no longer appeared.
> 
> hmm, I will do some more testing. My intent was that the old `crc_attach_default_interface` targets should still work for compatibility.

I was not able to reproduce this - there is a sleep for 5 seconds after detaching the interface. My guess is that in rare cases it can take more time for libvirt to update the network and in that case we would see the error above.

abays:
> > > I saw this after I installed a fresh CRC in an environment that had previously used `install_yamls` and the old `make crc_attach_default_interface`:
> > > ```
> > > $ make attach_default_interface
> > > bash scripts/interfaces-setup-cleanup.sh
> > > bash scripts/interfaces-setup.sh
> > > 
> > > error: Failed to update network default
> > > error: Requested operation is not valid: there is an existing dhcp host entry in network 'default' that matches "<host mac='unknown' name='crc' ip='192.168.122.10'/>"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I then did `make attach_default_interface_cleanup` and then re-ran `make attach_default_interface` and the message no longer appeared.
> > 
> > 
> > hmm, I will do some more testing. My intent was that the old `crc_attach_default_interface` targets should still work for compatibility.
> 
> I was not able to reproduce this - there is a sleep for 5 seconds after detaching the interface. My guess is that in rare cases it can take more time for libvirt to update the network and in that case we would see the error above.

Fair enough.  It's easy to work-around if encountered, anyhow.

abays (approved):
/lgtm

devsetup/scripts/interfaces-setup.sh
```diff
@@ -28,29 +28,29 @@ EOF
...
+    sudo virsh attach-device $INSTANCE_NAME --file /tmp/pci2.xml --persistent
     sleep 10
 
-    sudo virsh start crc
+    sudo virsh start $INSTANCE_NAME
     # wait for the VM to be up
     sleep 240
 fi
 
+virsh --connect=qemu:///system net-update default add-last ip-dhcp-host --xml "<host name='$INSTANCE_NAME' ip='$IP_ADDRESS'/>" --config --live
```
rabi, line 46: Looks like this has broken the baremetal job[1]. Let's see with https://github.com/openstack-k8s-operators/install_yamls/pull/627 that seems to fix it locally for me.

[1] https://review.rdoproject.org/zuul/builds?job_name=cifmw-crc-podified-edpm-baremetal&job_name=openstack-baremetal-operator-crc-podified-edpm-baremetal&job_name=install-yamls-crc-podified-edpm-baremetal&skip=0

PR #623, opened 2023-11-13
Fix incorrect syntax in standalone

gouthampacha:
This error was being ignored; and the manila
services were not being configured.


abays (approved):
/lgtm

PR #624, opened 2023-11-14
Parameterize "bmaas_attach_network" targets

hjensas:
BMAAS_OCP_INSTANCE_NAME defines which libvirt VM
to attach the bmaas network to.


abays (approved):
/lgtm

PR #627, opened 2023-11-15
Use $MAC_ADDRESS when updating default network

rabi:
The default interface gets different ip at times and looks like it breaks the baremetal job.

### DIFF
--- a/devsetup/scripts/interfaces-setup.sh
+++ b/devsetup/scripts/interfaces-setup.sh
@@ -43,8 +43,8 @@ EOF
     sleep 240
 fi
 
-virsh --connect=qemu:///system net-update default add-last ip-dhcp-host --xml "<host name='$INSTANCE_NAME' ip='$IP_ADDRESS'/>" --config --live
 MAC_ADDRESS=$(echo -n 52:54:00; dd bs=1 count=3 if=/dev/random 2>/dev/null | hexdump -v -e '/1 "-%02X"' | tr '-' ':')
+virsh --connect=qemu:///system net-update default add-last ip-dhcp-host --xml "<host name='$INSTANCE_NAME' mac='$MAC_ADDRESS' ip='$IP_ADDRESS'/>" --config --live
 virsh --connect=qemu:///system attach-interface $INSTANCE_NAME --source $NETWORK_NAME --type network --model virtio --mac $MAC_ADDRESS --config --persistent
 
 sleep 5
--- a/devsetup/scripts/interfaces-setup-cleanup.sh
+++ b/devsetup/scripts/interfaces-setup-cleanup.sh
@@ -9,7 +9,7 @@ fi
 MAC_ADDRESS=$(virsh --connect=qemu:///system dumpxml $INSTANCE_NAME | xmllint --xpath "string(/domain/devices/interface/source[@network=\"$NETWORK_NAME\"]/../mac/@address)" -)
 if [ -n "${MAC_ADDRESS}" ]; then
     virsh --connect=qemu:///system detach-interface $INSTANCE_NAME network --mac $MAC_ADDRESS
-    virsh --connect=qemu:///system net-update $NETWORK_NAME delete ip-dhcp-host "<host name='$INSTANCE_NAME'/>" --config --live
+    virsh --connect=qemu:///system net-update $NETWORK_NAME delete ip-dhcp-host "<host name='$INSTANCE_NAME' mac='$MAC_ADDRESS'/>" --config --live
     sleep 5
 fi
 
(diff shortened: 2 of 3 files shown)


abays (approved):
/lgtm

devsetup/scripts/interfaces-setup.sh
```diff
@@ -43,8 +43,8 @@ EOF
     sleep 240
 fi
 
-virsh --connect=qemu:///system net-update default add-last ip-dhcp-host --xml "<host name='$INSTANCE_NAME' ip='$IP_ADDRESS'/>" --config --live
 MAC_ADDRESS=$(echo -n 52:54:00; dd bs=1 count=3 if=/dev/random 2>/dev/null | hexdump -v -e '/1 "-%02X"' | tr '-' ':')
+virsh --connect=qemu:///system net-update default add-last ip-dhcp-host --xml "<host name='$INSTANCE_NAME' mac='$MAC_ADDRESS' ip='$IP_ADDRESS'/>" --config --live
```
hjensas, line 47: hmm, so using the mac here breaks on IPv6. Livrit does not allow using the mac as identifier, this is why I removed it. We statically configure the IP in the NNCP later, so I figured this reservation was only there to try ensure 192.168.122.10 was not allocated to some other node?

```
NETWORK_ISOLATION_IP_ADDRESS ?= 192.168.122.10
attach_default_interface: export IP_ADDRESS=${NETWORK_ISOLATION_IP_ADDRESS}
< - snip - >
attach_default_interface: attach_default_interface_cleanup ## Attach default libvirt network to CRC
	bash scripts/interfaces-setup.sh
```
NNCP_CTLPLANE_IP_ADDRESS_PREFIX     ?=192.168.122
NNCP_CTLPLANE_IP_ADDRESS_SUFFIX     ?=10

Here is where it is set static on the interface:
https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-nncp.sh#L148-L154

Do we have job's that don't apply the NNCP?

rabi: > Livrit does not allow using the mac as identifier 

You mean for ipv6? We may need to add to check for ipv6 and use DUID instead (?)

> So I figured this reservation was only there to try ensure 192.168.122.10 was not allocated to some other node?

I don't see lease for 192.168.122.10 if not using mac after running #make crc_attach_default_interface though. 

[stack@osp-df-2 devsetup]$ make crc_attach_default_interface
make attach_default_interface
make[1]: Entering directory '/home/stack/work/install_yamls/devsetup'
bash scripts/interfaces-setup-cleanup.sh
bash scripts/interfaces-setup.sh
Updated network default persistent config and live state
Interface attached successfully

make[1]: Leaving directory '/home/stack/work/install_yamls/devsetup'
[stack@osp-df-2 devsetup]$ sudo virsh net-dhcp-leases default
 Expiry Time           MAC address         Protocol   IP address          Hostname             Client ID or DUID
---------------------------------------------------------------------------------------------------------------------
 2023-11-15 07:00:09   52:54:00:b1:ce:5f   ipv4       192.168.122.47/24   crc-8cf2w-master-0   01:52:54:00:b1:ce:5f

> Do we have job's that don't apply the NNCP?


I don't think there is anything job specific here. NNCP should configure static IP. But it probably behaves weird at times.

hjensas: Yes, libvirt wants the DUID - but there is no way of knowing what that is :(. I think RHCOS uses DUID-LLT or DUID-UUID - the latter is based on machine-id which is generated on first boot - so we can't know in advance what to put in dhcp config.

* DUID-LLT (Type 1): using Link-Layer address plus Time (32-bits, seconds since 01.01.2000 00:00 UTC modulo 2^32)
* DUID-UUID (Type 4): using a Universally Unique Identifier (UUID)

The hostname `crc-8cf2w-master-0` looks generated as well, and no way to set that upfront with crc afict.

I think the solution might be to stop using the default network - and instead create our own network without any DHCP for this network isolation bridge. (I am working on changes for this) That way we won't have any IP before the NNCP is applied. If we absolutely need DHCP we can also manage that outside of libvirt if we get off the "default" network, dnsmasq actually supports using mac for both ipv4 and ipv6 so really this is a libvirt limitation.

rabi: > I think the solution might be to stop using the default network - and instead create our own network without any DHCP for this network isolation bridge

We switched to use default network to avoid the issue of restarting crc after attaching a new network and issues associated with pull-secret removed from disk in CI and crc not coming up.

hjensas: > We switched to use default network to avoid the issue of restarting crc after attaching a new network and issues associated with pull-secret removed from disk in CI and crc not coming up.

Yes, we can attach one additional network without restart because additional nets require additional PCI slots which cannot be hot-plugged. My plan is to attach one additional network as well, just not the "default" network at all. My plan is to  use that one network for both provisioning and network isolation vlans - so we should'nt need a crc restart.

PR #629, opened 2023-11-15
Run nncp target before crc_bmo_setup

rabi:
This would ensure ironic using the nncp configured static ip for the default interface rather than the dhcp configured one when attaching the interface.

Also reverts the change to use $MAC_ADDRESS when updating default network to unblock ipv6 work.


abays (approved):
/lgtm

hjensas:
Aha, I figured we had done this already.

/lgtm
/approved

PR #631, opened 2023-11-20
libvirt URI - no socket on modular libvirt

hjensas:
On my Fedora 39 - sushy-emulator is not able to connect to libvirt on the hypervisor when socket is specified in the connection URI.

This works:
  qemu+ssh://${USER}@${ADDRESS}/system
This fails:
  qemu+ssh://${USER}@${ADDRESS}/system?socket=/var/run/libvirt/libvirt-sock

Add a condition checking if libvirtd.service is-active, and only use the socket when that is the case.


rabi (approved):
/lgtm

PR #635, opened 2023-11-24
Fix parameterised $NETWORK_NAME in interfaces-setup.sh

hjensas:
The ip-dhcp-host command still hard codes the network name to default. We should use the $NETWORK_NAME variable.

### DIFF
--- a/devsetup/scripts/interfaces-setup.sh
+++ b/devsetup/scripts/interfaces-setup.sh
@@ -44,7 +44,7 @@ EOF
 fi
 
 MAC_ADDRESS=$(echo -n 52:54:00; dd bs=1 count=3 if=/dev/random 2>/dev/null | hexdump -v -e '/1 "-%02X"' | tr '-' ':')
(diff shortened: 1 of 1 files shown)


hjensas:
recheck

rabi:
recheck

hjensas:
recheck

hjensas:
recheck

hjensas:
recheck

rabi:
recheck

hjensas:
recheck

abays:
All the Zuul jobs have passed independently over many attempted batch runs.  Let's override to stop rolling the dice in the hopes for a perfect run with no hiccups.

rabi:
> All the Zuul jobs have passed independently over many attempted batch runs. Let's override to stop rolling the dice in the hopes for a perfect run with no hiccups.

There was kind of an agreement to not use override and instead fix the jobs. I think we have no other option but to revert https://github.com/openstack-k8s-operators/edpm-image-builder/pull/19 to make the baremetal job stable.

hjensas:
recheck

PR #641, opened 2023-11-28
[FIPS]Change ssh migration key to ECDSA

gibizer:
The e8b0fcaa7a13a0d4656b76e019d25d11d648fd13 switched on the FIPS crypto profile in the pre-deployed EDPM VMs. Even though the recent FIPS standard[1] allows ED25519 ssh keys, the current FIPS profile in centos9 stream does not allow it. So the current migration ssh key isn't usable. This PR changes the key type to ECDSA until the profile is updated.

[1] https://csrc.nist.gov/pubs/fips/186-5/final


fao89 (approved):

gibizer:
Tested locally cold migration works with this PR

PR #644, opened 2023-11-30
Remove ansibleVars from the NodeSet node spec

bshephar:
In an effort to simplify the dataplane samples. We will remove this from the sample file, which renders this kustomize operation unnecessary. See for reference:
    https://github.com/openstack-k8s-operators/dataplane-operator/pull/501

### DIFF
--- a/scripts/gen-edpm-kustomize.sh
+++ b/scripts/gen-edpm-kustomize.sh
@@ -54,8 +54,6 @@ patches:
     - op: replace
       path: /spec/nodes/edpm-compute-0/ansible/ansibleHost
       value: ${EDPM_COMPUTE_IP}
-    - op: remove
-      path: /spec/nodes/edpm-compute-0/ansible/ansibleVars
     - op: replace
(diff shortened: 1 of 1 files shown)


slagle (changes requested):
The actual change is ok, even though it's a little hard to completely tell :smile: 

Can you not adjust the style of the whole script as well? Maybe your editor did it automatically, but it doesn't match the existing style of the other scripts even. If that is something you'd like to do, I'd do it in a separate PR, and consistently across all scripts so we can gather feedback on it.

bshephar:
> The actual change is ok, even though it's a little hard to completely tell 😄
> 
> Can you not adjust the style of the whole script as well? Maybe your editor did it automatically, but it doesn't match the existing style of the other scripts even. If that is something you'd like to do, I'd do it in a separate PR, and consistently across all scripts so we can gather feedback on it.

Ah damn, yeah I think it was a LSP auto edit on save there. I'll amend that.

bshephar:
recheck

PR #648, opened 2023-12-01
Add support for deployment on community distro okd

amoralej:
This patch adds support to run deploy openstack operators in okd openshift distro. The main differences are related to the installation of external operators metallb, nmstate and cert-manager as openshift installation is based on the redhat-operators catalog which is not available in okd.

In order to implement it, I am adding a new parameter OKD which defaults to 'false' and should be 'true' to enable okd. I am trying to isolate the differences in separate scripts to avoid overcomplicating the scripts for deployment on openshift.


fao89:
recheck

fao89 (approved):

PR #650, opened 2023-12-04
Use edpm_deploy_prep target for generating crs with kustomization

raukadah:
Currently edpm_deploy_prep targets generates the edpm crs with kustomization and also applys the services and ansibleee env crs.

The prep target should be only generating the needed crs and kustomization.

It will help the ci-framework to use the generated crs and kustomization by install_yamls and apply them in transparent way.

Note: It also moves the ansibeleee and nova ssh key generation to edpm_deploy_generate_keys target and called via edpm_prep via conditional.
    
Since edpm_prep target is used in ci-framework so we need to keep calling edpm_deploy_generate_keys target to avoid breakage.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/907

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -362,6 +362,7 @@ DATAPLANE_KUTTL_DIR                              ?= ${OPERATOR_BASE_DIR}/datapla
 DATAPLANE_KUTTL_NAMESPACE                        ?= dataplane-kuttl-tests
 BM_CTLPLANE_INTERFACE                            ?= enp1s0
 BM_ROOT_PASSWORD_SECRET                          ?=
+GENERATE_SSH_KEYS				 ?= true
 
 # Manila
 MANILA_IMG              ?= quay.io/openstack-k8s-operators/manila-operator-index:${OPENSTACK_K8S_TAG}
@@ -650,6 +651,12 @@ openstack_deploy_cleanup: namespace netconfig_deploy_cleanup ## cleans up the se
 	oc kustomize ${DEPLOY_DIR} | oc delete --ignore-not-found=true -f - || true
 	${CLEANUP_DIR_CMD} ${OPERATOR_BASE_DIR}/openstack-operator ${DEPLOY_DIR}
 
+.PHONY: edpm_deploy_generate_keys
+edpm_deploy_generate_keys:
+	$(eval $(call vars,$@,dataplane))
+	devsetup/scripts/gen-ansibleee-ssh-key.sh
+	bash scripts/gen-edpm-nova-migration-ssh-key.sh
+
 .PHONY: edpm_deploy_prep
 edpm_deploy_prep: export KIND=OpenStackDataPlaneNodeSet
 edpm_deploy_prep: export EDPM_ANSIBLE_SECRET=${DATAPLANE_ANSIBLE_SECRET}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -677,8 +677,6 @@ edpm_deploy_prep: edpm_deploy_cleanup ## prepares the CR to install the data pla
 	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(DATAPLANE_BRANCH),-b ${DATAPLANE_BRANCH}) ${DATAPLANE_REPO} "${OPERATOR_NAME}-operator" && popd
 	cp devsetup/edpm/services/* ${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services
 	DEPLOY_DIR=${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services KIND=OpenStackDataPlaneService bash scripts/gen-edpm-services-kustomize.sh
-	oc kustomize ${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services | oc apply -f -
-	oc apply -f devsetup/edpm/config/ansible-ee-env.yaml
```
pablintino, line 681: Some Make shells still deploy stuff like `gen-edpm-nova-migration-ssh-key.sh` and `gen-ansibleee-ssh-key.sh`

pablintino:
I left a comment about some code that is still creating resources.

Makefile
raukadah: Moved to edpm_deploy step

raukadah:
recheck

raukadah:
recheck

raukadah:
recheck

fao89:
recheck

raukadah:
recheck

raukadah:
recheck

raukadah:
recheck

raukadah:
recheck dpa job failure

raukadah:
recheck

raukadah:
recheck

Makefile
```diff
@@ -676,16 +683,15 @@ edpm_deploy_prep: edpm_deploy_cleanup ## prepares the CR to install the data pla
 	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR} ${DEPLOY_DIR}
 	pushd ${OPERATOR_BASE_DIR} && git clone ${GIT_CLONE_OPTS} $(if $(DATAPLANE_BRANCH),-b ${DATAPLANE_BRANCH}) ${DATAPLANE_REPO} "${OPERATOR_NAME}-operator" && popd
 	cp devsetup/edpm/services/* ${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services
-	oc apply -f ${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services
 	oc patch $(shell oc get csv -n ${OPERATOR_NAMESPACE} -o name | grep ansibleee) \
 	-n ${OPERATOR_NAMESPACE} --type='json' \
```
pablintino, line 687: Uhmm cant this be moved to the non _prep too? This target should work even without an OCP running

Makefile
```diff
@@ -718,20 +726,21 @@ edpm_deploy_baremetal_prep: edpm_deploy_cleanup ## prepares the CR to install th
...
-	bash scripts/gen-edpm-nova-migration-ssh-key.sh
+ifeq ($(GENERATE_SSH_KEYS), true)
+	make edpm_deploy_generate_keys
+endif
 
 .PHONY: edpm_deploy_baremetal
 edpm_deploy_baremetal: input edpm_deploy_baremetal_prep ## installs the dataplane instance using kustomize. Runs prep step in advance. Set DATAPLANE_REPO and DATAPLANE_BRANCH to deploy from a custom repo.
 	$(eval $(call vars,$@,dataplane))
+	oc apply -f ${OPERATOR_BASE_DIR}/${OPERATOR_NAME}-operator/config/services
+	oc patch $(shell oc get csv -n openstack-operators -o name | grep ansibleee) \
```
rabi, line 740: Why're we patching the  runner image here? This would override what's set in the CI.

PR #651, opened 2023-12-04
Update kustomize to v5.0.3

fultonj:


fultonj:
recheck

abays:
recheck

abays (approved):
/lgtm

abays:
I think this will work as-is, but for thoroughness, should we update this [1] as well?

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/8643f4b43522c6acd14d7735b0830ded1438a7ce/devsetup/roles/download_tools/defaults/main.yaml#L15

PR #652, opened 2023-12-05
Use NAMESPACE variable for nova-migration-key

bshephar:
This change adds the NAMESPACE variable to the nova-migration-key generatation script rather than hard-coding the openstack namespace.


fao89:
recheck

fao89 (approved):

PR #654, opened 2023-12-06
Add the ability to deploy multiple Ceph Pods

fmount:
We're working on a feature that allows to deploy an arbitrary numbers of `GlanceAPI` [1].
To ease the test and to be closer to an `edge` scenario, an approach would be to have the ability to deploy multiple `Ceph` `Pods`: they bring different `Secrets` that can be propagated to a subset of `ctlplane` components.
This patch introduces the new `CEPH_CLUSTERS` variable that is used within the bash script to (eventually) deploy multiple `Ceph` `Pods`.

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/384

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -386,6 +386,7 @@ MANILA_KUTTL_NAMESPACE  ?= manila-kuttl-tests
 
 # Ceph
 CEPH_IMG            ?= quay.io/ceph/demo:latest-reef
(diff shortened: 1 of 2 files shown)


fmount:
Closing for now because we don't have capacity to work on this.

PR #656, opened 2023-12-10
Allow sushy-tools image to be overridden

steveb:
This will allow sushy-tools feature work which ci-framework will use.


abays (approved):
/lgtm

PR #657, opened 2023-12-11
Use virtual-media in CI

rabi:
Rather than using the same ctlplane network for provisioning let's use virtual media BMC addressing option.

### DIFF
--- a/devsetup/scripts/edpm-compute-baremetal.sh
+++ b/devsetup/scripts/edpm-compute-baremetal.sh
@@ -70,7 +70,7 @@ metadata:
     app: openstack
 spec:
   bmc:
-    address: redfish+http://sushy-emulator.apps-crc.testing/redfish/v1/Systems/${!uuid_var}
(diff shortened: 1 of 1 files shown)


hjensas (approved):
Cool, I did not think this would work - I was looking at doing this and saw this note in sushy-emulator [docs](https://docs.openstack.org/sushy-tools/latest/user/dynamic-emulator.html#virtual-media-boot):
```
Note

ISO files to boot from must be UEFI-bootable, libvirtd should be running on the same machine with sushy-emulator.
```

But, it looks like that information is not correct. It boots from `BdsDxe: loading Boot0001 "UEFI QEMU DVD-ROM QM00001 " from PciRoot(0x0)/Pci(0x1F,0x2)/Sata(0x0,0xFFFF,0x0)` here:
https://logserver.rdoproject.org/57/657/9296bc1fce89b34face520edaa7bc8bd3a7cf597/github-check/install-yamls-crc-podified-edpm-baremetal/d32440e/ci-framework-data/logs/bmaas_console_logs/edpm-compute-01_no_ansi_2023-12-11-04:25:36.log

abays (approved):
/lgtm

PR #660, opened 2023-12-11
Allow sushy emulator to use openstack

steveb:
This change allows sushy-tools to be started with openstack as the driver instead of libvirt. With the current default values there is no change from the current libvirt behaviour.

With this change, a configured sushy-emulator pod can be started which is managing openstack instances instead of libvirt VMs.

Further feature work is required on sushy-tools for this to be actually useful. This is the beginning of reworking of baremetal jobs to become multinode to improve reliability.

### DIFF
--- a/devsetup/scripts/bmaas/sushy-emulator.sh
+++ b/devsetup/scripts/bmaas/sushy-emulator.sh
@@ -18,26 +18,36 @@ MY_TMP_DIR="$(mktemp -d)"
 trap 'rm -rf -- "$MY_TMP_DIR"' EXIT
 
 NODE_NAME_PREFIX=${NODE_NAME_PREFIX:-"crc-bmaas"}
-LIBVIRT_USER=${LIBVIRT_USER:-"sushyemu"}
 NAMESPACE=${SUSHY_EMULATOR_NAMESPACE:-"sushy-emulator"}
-SSH_ALGORITHM=rsa
-SSH_KEY_FILE=bmaas-ssh-key-id_rsa
-SSH_KEY_SIZE=4096
-CRC_NETWORK_NAME=crc
+DRIVER=${SUSHY_EMULATOR_DRIVER:-"libvirt"}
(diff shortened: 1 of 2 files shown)


devsetup/scripts/bmaas/sushy-emulator.sh
```diff
@@ -18,26 +18,36 @@ MY_TMP_DIR="$(mktemp -d)"
...
+    LIBVIRT_IP_ADDRESS=$(nmcli connection show "${CRC_NETWORK_NAME}" | grep ipv4.addresses | cut -d / -f 1 | awk '{ print $2 }')
+    INSTANCES=$(virsh --connect=qemu:///system list --all --uuid --name | grep "${NODE_NAME_PREFIX}" | awk '{ printf "\"" $1 "\" " }' | sed -e 's/" "/", "/' -e 's/^"/["/' -e 's/" $/"]/')
+    LIBVIRT_URI="'qemu+ssh://${LIBVIRT_USER}@${LIBVIRT_IP_ADDRESS}/system${LIBVIRT_SOCKET}'"
+    EMULATOR_OS_CLOUD="None"
+elif [ $DRIVER = "openstack" ]; then
+    OS_CLIENT_CONFIG_FILE=${SUSHY_EMULATOR_OS_CLIENT_CONFIG_FILE:-/etc/openstack/clouds.yaml}
+    INSTANCES=$(openstack --os-cloud=${SUSHY_EMULATOR_OS_CLOUD} server list --name "edpm-compute.*" -f json -c ID | jq -c [.[].ID])
+    LIBVIRT_URI="None"
+    EMULATOR_OS_CLOUD="'${SUSHY_EMULATOR_OS_CLOUD}'"
+fi
```
hjensas, line 50: We may want to add a final else here, or earlier in the script and error if $DRIVER is not one of the supported drivers.

hjensas (approved):
A nit in comments, I think we can merge as is and follow up.

fao89 (approved):

PR #665, opened 2023-12-14
Further increase edpm vm rng rate, bytes

steveb:
Kernel panics due to FIPS entropy test failures are still happening[1], but with less frequency. The improvement occured when the rng rate was increased. This change increases the frequency some more, and also increases the bytes that can be consumed by the guest vm. We will observe if this reduces the number of kernel panics further.

[1] https://review.rdoproject.org/zuul/build/fdde6cf319a248d7850276f8c5b9aedc/log/ci-framework-data/logs/bmaas_console_logs/edpm-compute-01_console.log


abays (approved):
/lgtm

PR #658, opened 2023-12-11
Allow passing custom config to DataPlaneService/nova

gibizer:
The new optional DATAPLANE_EXTRA_NOVA_CONFIG_FILE ENV variable allows adding extra configuration to the nova-compute service on the EDPM nodes. It create a nova-custom DataPlaneService passing the provided file via a config map and includes that service in the NodeSet.

e.g.:
```
$ cat > ./nova-extra.conf
[foo]
bar = 1
$
$ DATAPLANE_EXTRA_NOVA_CONFIG_FILE=./nova-extra.conf make edpm_wait_deploy
```

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/947 (merged)

PR #667, opened 2023-12-15
[edpm_wait_deploy]Fix waiting for Ready condition

gibizer:
Since #658 landed the dataplane output dir has DataPlaneService resource generated and that does not have Ready condition. So the edpm_wait_deploy target is changed not to wait for all resources in the output dir but to only wait for the Deployment and the NodeSet readyness.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -763,13 +763,13 @@ edpm_deploy_baremetal: input edpm_deploy_baremetal_prep ## installs the dataplan
 .PHONY: edpm_wait_deploy_baremetal
 edpm_wait_deploy_baremetal: edpm_deploy_baremetal ## waits for dataplane readiness. Runs prep step in advance. Set DATAPLANE_REPO and DATAPLANE_BRANCH to deploy from a custom repo.
 	$(eval $(call vars,$@,dataplane))
-	oc kustomize ${DEPLOY_DIR} | oc wait --for condition=Ready --timeout=$(BAREMETAL_TIMEOUT) -f -
+	oc kustomize ${DEPLOY_DIR} | yq '. | select(.kind == "OpenStackDataPlaneNodeSet"), select(.kind == "OpenStackDataPlaneDeployment")' | oc wait --for condition=Ready --timeout=$(BAREMETAL_TIMEOUT) -f -
 	$(MAKE) edpm_nova_discover_hosts
 
 .PHONY: edpm_wait_deploy
 edpm_wait_deploy: edpm_deploy ## waits for dataplane readiness. Runs prep step in advance. Set DATAPLANE_REPO and DATAPLANE_BRANCH to deploy from a custom repo.
 	$(eval $(call vars,$@,dataplane))
-	oc kustomize ${DEPLOY_DIR} | oc wait --for condition=Ready --timeout=$(DATAPLANE_TIMEOUT) -f -
+	oc kustomize ${DEPLOY_DIR} | yq '. | select(.kind == "OpenStackDataPlaneNodeSet"), select(.kind == "OpenStackDataPlaneDeployment")' | oc wait --for condition=Ready --timeout=$(DATAPLANE_TIMEOUT) -f -
 	$(MAKE) edpm_nova_discover_hosts
 
 .PHONY: edpm_register_dns


fmount:
Same here: it requires https://github.com/openstack-k8s-operators/data-plane-adoption/pull/221 to land to unblock the `data-plane-adoption` job (/cc @cescgina)

fmount:
recheck

Makefile
```diff
@@ -763,13 +763,17 @@ edpm_deploy_baremetal: input edpm_deploy_baremetal_prep ## installs the dataplan
 .PHONY: edpm_wait_deploy_baremetal
 edpm_wait_deploy_baremetal: edpm_deploy_baremetal ## waits for dataplane readiness. Runs prep step in advance. Set DATAPLANE_REPO and DATAPLANE_BRANCH to deploy from a custom repo.
 	$(eval $(call vars,$@,dataplane))
-	oc kustomize ${DEPLOY_DIR} | oc wait --for condition=Ready --timeout=$(BAREMETAL_TIMEOUT) -f -
+	oc kustomize ${DEPLOY_DIR}
+	oc wait --for condition=Ready OpenStackDataPlaneDeployment --all --timeout=$(BAREMETAL_TIMEOUT)
+	oc wait --for condition=Ready OpenStackDataPlaneNodeSet --all --timeout=$(BAREMETAL_TIMEOUT)
```
abays, line 768: Do we need the `oc kustomize ${DEPLOY_DIR}`?  I'm trying to understand what purpose it serves without passing its result to the old `oc wait`.

Also, I guess since these EDPM deploy targets ultimately depend on (and execute) `edpm_deploy_cleanup`, the cluster normally won't have any extra `OpenStackDataPlaneDeployments` or `OpenStackDataPlaneNodeSets` lingering.  If the user manually created additional ones, however, things would get stuck here with the `--all` flag.  But I think it's fine for now.

gibizer: > Also, I guess since these EDPM deploy targets ultimately depend on (and execute) edpm_deploy_cleanup, the cluster normally won't have any extra OpenStackDataPlaneDeployments or OpenStackDataPlaneNodeSets lingering. If the user manually created additional ones, however, things would get stuck here with the --all flag. But I think it's fine for now.

Yeah, I either wait for every CR of these types or I need a way to know the names of the CRs to wait for. The latter seemed more complicated / fragile as any sample can be passed on install_yamls. Let me see if I can come up with something reasonable. as I anyhow need to respin this.

gibizer:
> @gibizer the pr link in the description looks unrelated to "not have Ready condition"?

It is related. The #658 started copying the DataPlaneService/nova definition to the output dir to kustomize it (rename it to nova-custom and inject an additional ConfigMap). So now the output dir contains a resource that has no Ready condition and hence the oc wait fails.

Makefile
gibizer: What about a nice yq filter on the kustomize output?  :)

Makefile
```diff
@@ -763,13 +763,13 @@ edpm_deploy_baremetal: input edpm_deploy_baremetal_prep ## installs the dataplan
 .PHONY: edpm_wait_deploy_baremetal
 edpm_wait_deploy_baremetal: edpm_deploy_baremetal ## waits for dataplane readiness. Runs prep step in advance. Set DATAPLANE_REPO and DATAPLANE_BRANCH to deploy from a custom repo.
 	$(eval $(call vars,$@,dataplane))
-	oc kustomize ${DEPLOY_DIR} | oc wait --for condition=Ready --timeout=$(BAREMETAL_TIMEOUT) -f -
+	oc kustomize ${DEPLOY_DIR} | yq '. | select(.kind == "OpenStackDataPlaneNodeSet"), select(.kind == "OpenStackDataPlaneDeployment")' | oc wait --for condition=Ready --timeout=$(BAREMETAL_TIMEOUT) -f -
```
abays, line 766: Should we add `yq` to the dev-setup's `download_tools` role?

abays (commented):
+1.  Tested in my environment.

Makefile
gibizer: I've added it recently https://github.com/openstack-k8s-operators/install_yamls/commit/eecf8bc83022ea129324fccc73aedb0bfa2ddcb3

gibizer:
recheck cert manager failed

PR #669, opened 2023-12-16
Add NetworkAttachmentDefinition for OVN-Octavia bridge

beagles:
Octavia Amphora Controller services need to be able to communicate with amphora instances connected to an dedicated tenant network. This bridge is necessary to provide connectivity between the pods.


fao89:
recheck

PR #671, opened 2023-12-18
[edpm_deploy_instance] Handle already exist resources

karelyatin:
Some resources was missing the check for already existing resource.

Also created a basic security group with required icmp/ssh rules instead of modifying the default security group.


abays (approved):
/lgtm

PR #673, opened 2023-12-19
Remove NodeSet env section

bshephar:
We modify the /spec/env section of the OpenStackDataPlaneNodeSet in ci-framework. Having it done in both places is confusing from a ci perspective. This change removes the kustomize section associated with /spec/env and we can instead update this to suite our needs from a dev perspective within the cluster. From a CI perspective, we can modify it via the ci-framework.


bshephar:
recheck

lewisdenny:
Patch adding this into ci-framework: https://github.com/openstack-k8s-operators/ci-framework/pull/944

bshephar:
recheck

bshephar:
recheck

fao89 (approved):

PR #676, opened 2023-12-21
Fix Ceph Pod multus annotations to work with 4.14

fmount:
Moving from `OCP` `4.13` to `4.14` has a few issues related to the `Ceph` Pod [1]. The problem seems to be the `empty value` assigned by default to the `multus annotation`, that is now removed from the `Pod` by default.
Only when a valid annotation is passed (see `make ceph_help` for more details), `kustomize` builds the `Ceph` Pod accordingly.

[1] https://github.com/openshift/release/pull/46911

Fixes: [OSPRH-2711](https://issues.redhat.com/browse/OSPRH-2711)


abays (approved):
/lgtm

PR #677, opened 2023-12-21
Fix NNCP for IPV6_ENABLED

hjensas:
The bridge must be set up on the interface no matter if it is IPv4 or IPv6. Move the snippet outside the if condition.

### DIFF
--- a/scripts/gen-nncp.sh
+++ b/scripts/gen-nncp.sh
@@ -291,8 +291,17 @@ EOF_CAT
     #
     cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
     - description: Configuring Bridge ${BRIDGE_NAME} with interface ${INTERFACE}
+      name: ${BRIDGE_NAME}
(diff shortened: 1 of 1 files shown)


scripts/gen-nncp.sh
```diff
@@ -323,6 +323,9 @@ EOF_CAT
         cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
       ipv6:
         enabled: false
+EOF_CAT
+    fi
+    cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
       mtu: ${INTERFACE_MTU}
```
karelyatin, line 329: may be better to move this block to L295, and drop duplicated mtu and state?

karelyatin (changes requested):
comment inline

scripts/gen-nncp.sh
hjensas: Makes sense, Done.

karelyatin (approved):

PR #679, opened 2024-01-03
Fix dataplane_kuttl

fao89:
dataplane_kuttl_prep is needed as dependency for CI: https://github.com/openshift/release/blob/master/ci-operator/step-registry/openstack-k8s-operators/kuttl/openstack-k8s-operators-kuttl-commands.sh#L100


fao89:
recheck

slagle (approved):

PR #680, opened 2024-01-03
Add Barbican encryption key to inputs

dmendiza:
This patch adds the Barbican Simple Crypto Backend encryption key (KEK) to the input target.  The key is stored in the osp-secret along with all other secrets.

Jira: [OSPRH-2390](https://issues.redhat.com//browse/OSPRH-2390)


dmendiza:
recheck

abays (approved):
/lgtm

PR #690, opened 2024-01-15
Consolidate adoption CI jobs

cescgina:
Make sure that all repos other than data-plane-adoption use the same
adoption job. Before this change, install_yamls used a different one
than ci-framework or openstack-operator.


marios (approved):
lgtm

PR #695, opened 2024-01-17
IPv6 kustomization for OpenstackControlplane

hjensas:
Edit the OpenstackControlplane for IPv6 or IPv4/IPv6 single/dual-stack.


stuggi (approved):
/lgtm

PR #702, opened 2024-01-23
Don't patch runner image in edpm_deploy_baremetal

rabi:
This breaks the baremetal jobs as this will override the images generated with edpm-ansible PR. Not changed `edpm_deploy` as CI does not use that target and may break something else.


karelyatin:
/lgtm

If it's required for dev setup can be taken care in follow ups

lewisdenny (approved):

karelyatin:
Fails with same issue:-
TASK [osp.edpm.edpm_users : Ensure user is present on the host [ openvswitch ]] ***
Tuesday 23 January 2024  07:23:59 +0000 (0:00:01.033)       0:00:04.461 ******* 
[0;31mfatal: [edpm-compute-0]: FAILED! => {"changed": false, "msg": "usermod: user openvswitch is currently used by process 1190\n", "name": "openvswitch", "rc": 8}[0m

We can override the failures once we confirm the baremetal job in other dependent PR succeeds

rabi:
> If it's required for dev setup can be taken care in follow ups

https://github.com/openstack-k8s-operators/install_yamls/pull/703 will do that. If we want, we can merge that instead.

karelyatin:
/approve

job passed in dependent patch

PR #703, opened 2024-01-23
Patch runner image conditionally

rabi:
We use the same targets in CI and use the images built with PR. If someone wants to use a custom image they can change the $DATAPLANE_RUNNER_IMG to use that image.


rabi:
recheck RETRY_LIMIT

fao89 (approved):

PR #705, opened 2024-01-24
Standalone: Deploy cinder-backup

Akrog:


Akrog:
recheck

jistr:
This should be tested with the adoption job but it's not getting triggered. Can you try adding `- ^devsetup/standalone/*` here:

https://github.com/openstack-k8s-operators/install_yamls/blob/main/zuul.d/projects.yaml#L16

Akrog:
> This should be tested with the adoption job but it's not getting triggered. Can you try adding `- ^devsetup/standalone/*` here?
> 
> https://github.com/openstack-k8s-operators/install_yamls/blob/main/zuul.d/projects.yaml#L16

The zuul change in the PR doesn't seem trigger the job. Maybe it needs to be merged to be in effect?

fao89:
> > This should be tested with the adoption job but it's not getting triggered. Can you try adding `- ^devsetup/standalone/*` here?
> > https://github.com/openstack-k8s-operators/install_yamls/blob/main/zuul.d/projects.yaml#L16
> 
> The zuul change in the PR doesn't seem trigger the job. Maybe it needs to be merged to be in effect?

it was triggered: https://review.rdoproject.org/zuul/build/01435481e3c148a0b7fa5c23dd1a309c

fao89 (approved):

PR #706, opened 2024-01-24
Add a backend for Glance kuttl tests

fmount:
`Glance` kuttl tests are already testing the instance deployment with a `File` backend. 
However, with the `Glance` patch [1], we are adding a constraint to the glance `split` layout, and a valid backend (other than file) is required.
This patch adds `Swift` as `Glance` backend to validate the `Split` scenario with a backend different than `Ceph`.

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/417


cschwede:
recheck

fao89 (approved):

PR #710, opened 2024-01-30
Ensure RabbitmqCluster CRD installed for dataplane kuttl

rabi:
infra-operator requires the RabbitmqCluster CRD, otherwise it fails with error as it now watches for it[1].

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/controllers/rabbitmq/transporturl_controller.go#L353


bshephar (approved):

rabi:
recheck

PR #713, opened 2024-01-31
Standalone: idempotently redeploy ceph, use CEPH_IP

bogdando:
Allow changing existing standalone tripleo ceph deployment in-place. For example, if it has failed on some step, and cannot be retried, because of "failed to set up loop device: Device or resource busy"

Allow standalone tripleo ceph to be deployed on custom IP, which by default is bound to edpm node suffix.

Also bypass 'Cannot infer CIDR network' by using
extra args and --yes --force options (otherwise one cannot use --skip-mon-network)
See https://github.com/openstack/tripleo-ansible/blob/master/ tripleo_ansible/roles/tripleo_cephadm/tasks/bootstrap.yaml#L65


bogdando:
would be nice to have, albeit no time to troubleshoot this. I'm swithcing back to use default suffixes.

PR #716, opened 2024-02-05
Remove redundant standalone_vm var for data plane adoption CI

marios:
This removes the standalone_vm var which was necessary when we had more than one type of DPA CI job (nested vm and non nested standalone deployment). See [1] for more context.

[1] https://issues.redhat.com/browse/OSPRH-4450

### DIFF
--- a/devsetup/scripts/standalone.sh
+++ b/devsetup/scripts/standalone.sh
@@ -25,10 +25,6 @@ COMPUTE_DRIVER=${2:-"libvirt"}
 EDPM_COMPUTE_ADDITIONAL_NETWORKS=${3:-'[]'}
 EDPM_COMPUTE_NAME=${EDPM_COMPUTE_NAME:-"edpm-compute-${EDPM_COMPUTE_SUFFIX}"}
 EDPM_COMPUTE_NETWORK=${EDPM_COMPUTE_NETWORK:-default}
-STANDALONE_VM=${STANDALONE_VM:-"true"}
-if [[ ${STANDALONE_VM} == "true" ]]; then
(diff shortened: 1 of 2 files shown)


devsetup/scripts/standalone.sh
```diff
@@ -25,10 +25,6 @@ COMPUTE_DRIVER=${2:-"libvirt"}
 EDPM_COMPUTE_ADDITIONAL_NETWORKS=${3:-'[]'}
 EDPM_COMPUTE_NAME=${EDPM_COMPUTE_NAME:-"edpm-compute-${EDPM_COMPUTE_SUFFIX}"}
 EDPM_COMPUTE_NETWORK=${EDPM_COMPUTE_NETWORK:-default}
-STANDALONE_VM=${STANDALONE_VM:-"true"}
-if [[ ${STANDALONE_VM} == "true" ]]; then
-    EDPM_COMPUTE_NETWORK_IP=$(virsh net-dumpxml ${EDPM_COMPUTE_NETWORK} | xmllint --xpath 'string(/network/ip/@address)' -)
```
cescgina, line 30: isn't `EDPM_COMPUTE_NETWORK_IP` undefined after this change?

cescgina: even though we don't use it for adoption CI, I think this is still needed for dev environments

cescgina:
recheck crc died

devsetup/scripts/standalone.sh
marios: ok yes indeed for the 'make standalone' case which creates the vm.

in this case i will abandon this one.

i will revise https://review.rdoproject.org/r/c/rdo-jobs/+/51547 so that we will remove the skip_standalone_vm we use in the play conditionals but keep this one standalone_vm passed into this side.

marios:
per discussion abandoning

PR #717, opened 2024-02-07
Add certmanager to dataplane_kuttl_tests

vakwetu:
We need to ensure that the certmanager operators are installed and running so that we can do kuttl tests for TLS certs


bshephar (approved):

fao89:
recheck

fao89 (approved):

PR #719, opened 2024-02-08
Move yq parameters

paramite:
Currently `make openstack` fails with yq installed via pip (yq 3.2.3). This patch make it work with it again.


paramite:
recheck

fao89:
recheck

paramite:
recheck

abays:
recheck

abays:
The baremetal job seems to consistently fail with this error:

```
 "error: must build at directory: not a valid directory: evalsymlink failure on '/home/zuul/ci-framework-data/artifacts/manifests/operator/baremetal-operator/config' : lstat /home/zuul/ci-framework-data/artifacts/manifests/operator: no such file or directory"
 ```

**Comment by paramite** [comment]
Studying the logs more deeply it fails with `Error: unknown shorthand flag: 'y' in -y`. Apparently version used by CI does not have -y option which unfortunately is required by version available in PIP. How is the job getting `yq`, so that I can get the same version in my lab? I guess we can give up on this PR until CI gets updated version.

**Comment by fao89** [comment]
I believe it gets yq from here: https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/roles/download_tools/tasks/main.yaml#L90

**Comment by ciecierski** [comment]
Hello,
@fao89 it is indeed [1] what causes problems. We should use yq from pip, not source. The yq from pip doesn't recognize  -y flag, but it works without it just fine.

[1]https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/roles/download_tools/tasks/main.yaml#L116

**Comment by abays** [comment]
This old PR was merged yesterday and broke CI:

```
    Error: unknown shorthand flag: 'y' in -y
    Usage:
      yq eval [expression] [yaml_file1]... [flags]
```

https://softwarefactory-project.io/zuul/t/rdoproject.org/build/d5cc04e9203e4801b7f4d8bbfefeef1f

PR #720, opened 2024-02-09
Correct order of default route on crc node

lewisdenny:
Starting with OCP 4.14, nmstate/network manager is beginning to create
the `ospbr` bridge default route as "static" which causes it to have a
higher priority over the external `ens3` interface.

This causes external connectivity to drop after the nncp resource is
applied and prevents Zuul from connecting to collect logs and fails the
job.

There has always been two default routes but it hasn't caused a problem
until OCP 4.14 where the priority changed.

JIRA:[OSPRH-4675](https://issues.redhat.com//browse/OSPRH-4675)
JIRA: [OSPCIX-182](https://issues.redhat.com/browse/OSPCIX-182)

### DIFF
--- a/scripts/gen-nncp.sh
+++ b/scripts/gen-nncp.sh
@@ -141,6 +141,7 @@ EOF_CAT
       - destination: 0.0.0.0/0
         next-hop-address: ${GATEWAY}
         next-hop-interface: ${BRIDGE_NAME}
+        metric: 101
 EOF_CAT
     fi
     if [ -n "$IPV6_ENABLED" ]; then


bshephar:
Sounds like something we would also need to document. Is there a Jira issue open? We can probably explore the wider impact to our validated architectures. Since our goal is to release on 4.18 (I believe), we would need to explicitly warn users about this when they are creating their NNCPs.

lewisdenny:
Testing here: https://code.engineering.redhat.com/gerrit/c/testproject/+/450036

If it fixes our issue I'll create a var for that subnet.

lewisdenny:
Now after nmstate runs both ocp4.12 and ocp4.14 route layout looks the same and connectivity isn't broken:

ocp4.12:
```
[core@crc-74q6p-master-0 ~]$ ip r
default via 10.0.199.254 dev ens3 proto dhcp src 10.0.199.216 metric 100
default via 192.168.122.1 dev ens7 proto static metric 101
```

ocp4.14
```
[core@crc-pjmnl-master-0 ~]$ ip r
default via 10.0.199.254 dev ens3 proto dhcp src 10.0.199.241 metric 100
default via 192.168.122.1 dev ospbr proto static metric 101
```

lewisdenny:
I would like to remove the extra default route but nmstate is being weird about it so I'll reach out to the nmstate folks and do a follow up PR. For now this route order unblocks our issue with 4.14 in CI.

scripts/gen-nncp.sh
```diff
@@ -141,6 +141,7 @@ EOF_CAT
       - destination: 0.0.0.0/0
```
bshephar, line 141: So, does this need to stay `0.0.0.0/0`? We can't just use the ctlplane subnet for it?

lewisdenny: For now yes, when trying to use the ctlplane subnet nmstate operator fails to apply the config sometimes due to an unrelated IPV6 route test ping.

I left some details on that in the Jira but I would like to resolve that in a follow up patch to unblock 4.14 for now.

lewisdenny: For some context, when nmstate operator applies the policy it then generates a list of probes to test. This comes from the generated output of `nmstatectl show` and tries to ping each gateway. If one of the pings fail the policy is rolled back and failed, I need to touch base with the nmstate folks.

bshephar (approved):

marios (approved):
seems reasonable

PR #726, opened 2024-02-13
Add NetworkAttachmentDefinition for OVN-Octavia bridge

beagles:
Octavia Amphora Controller services need to be able to communicate with amphora instances connected to an dedicated tenant network. This bridge is necessary to provide connectivity between the pods.

(note also corrects accidental removal of the designate network attachment)


Akrog (approved):
This looks good to me. It follows the same pattern as the other VLANs to provide pod to pod communication regardless of where the pods are.
Maybe we should have a function that abstract all the common stuff, but that should hold this patch.

PR #727, opened 2024-02-14
Run baremetal job for any changes to scripts

rabi:
We use different nodesets and networking for pre-provisioned and baremetal jobs. We should run baremetal job for any changes to things like `gen-nncp.sh`, so as to avoid issues like https://issues.redhat.com/browse/OSPCIX-185.

JIRA: https://issues.redhat.com/browse/OSPCIX-185


karelyatin (approved):
/lgtm

PR #728, opened 2024-02-14
Cleanup pre-existing nncp CR yamls

Akrog:
If we run `make openstack`, then recreate the CRC VM and rerun `make openstack` it will fail.

The reason for this that we are creating the nncp yaml filenames based on the WORKER which changes on each deployment.

Since it changes we end up with multiple yamls that are leftover from previous runs.

We could use a fixed name for the yamls, but we would run into the same scenario if number of WORKERS changes between runs.

This patch just removes all nncp yamls from the destination directory before generating them.


abays (approved):
/lgtm

PR #733, opened 2024-02-15
increase default ceph timeout to 300 seconds

SeanMooney:
This script bumpts the time we wait for ceph to be
ready form 30 seconds to 300 to ensure we dont
time out in ci.

### DIFF
--- a/scripts/gen-ceph-kustomize.sh
+++ b/scripts/gen-ceph-kustomize.sh
@@ -35,7 +35,12 @@ fi
 
 pushd ${DEPLOY_DIR}
 
-CEPH_TIMEOUT=${CEPH_TIMEOUT:-30}
+# this should not take 300 second however we have seen it take > 30 seconds
+# in ci as it takes 18 seconds just to pull the image so we set this to 300
+# as a limit we should never hit. IF this is timing out at 300 seconds
+# bumping the timeout again is likely not the correct action and something
+# else is wrong.
+CEPH_TIMEOUT=${CEPH_TIMEOUT:-300}
 CEPH_HOSTNETWORK=${CEPH_HOSTNETWORK:-true}
 CEPH_POOLS=("volumes" "images" "backups" "cephfs.cephfs.meta" "cephfs.cephfs.data")
 CEPH_DAEMONS="osd,mds,rgw"


scripts/gen-ceph-kustomize.sh
```diff
@@ -35,7 +35,12 @@ fi
 
 pushd ${DEPLOY_DIR}
 
-CEPH_TIMEOUT=${CEPH_TIMEOUT:-30}
+# this should not take 300 second however we have seen it take > 30 seconds
+# in ci as it takes 18 seconds just to pull the image so we set this to 300
+# as a limit we should never hit. IF this is timing out at 300 seconds
+# bumping the timeout again is likely not the correct action and something
+# else is wrong.
+CEPH_TIMEOUT=${CEPH_TIMEOUT:-300}
```
fmount, line 43: I can approve this patch, however I have two comments:
1. it shouldn't take 300 seconds;
2. this change invalidates the bump we did in `Prow` jobs (the effect would be "decreasing" the timeout, which doesn't make sense) [1];
3. `ci-framework` allows to customize these values, so I think you can try your job and gather some data before attempting this bump (also it would help to have a more realistic timeout) 

Maybe I'm overlooking the main purpose here, but I'm wondering if we want to first give a try to bump this in the jobs provided by[1][2].
I also think that you're right when you say that 30 secs is a very low value, but w/ 90 secs in the upstream CI (mostly Prow) I've never seen a high amount of failures (/cc @viroel who can give us some numbers).  

[1] https://github.com/openshift/release/blob/master/ci-operator/step-registry/openstack-k8s-operators/deploy/openstack-k8s-operators-deploy-commands.sh#L11
[2] https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/centos-9/ceph_backends.yml#L10

SeanMooney: yep it does i found that after ward im going to close this for now we can bump it later. for me it was takeing about 50 second in the job for ceph to actully be deployed and ready.

i have refactored to using the ceph_backends.yml in my job https://github.com/openstack-k8s-operators/nova-operator/pull/685/files#diff-108978819c05ae183d88ec87959c2341a94cfc3f9465e3aeee82d554217b4f58R178

SeanMooney:
this is overridable, i orginally opened this as it was not clear how to do that with the ci framework but i found that after dinner last night so this is not needed

PR #604, opened 2023-10-25
Drop edpm_baremetal/edpm_baremetal_cleanup targets

rabi:
These are not used anymore in CI and the documentation points to using `edpm_baremetal_compute` and `edpm_deploy_baremetal` instead.

PR #553, opened 2023-09-28
Delete devsetup gen-edpm-baremetal-kustomize.sh

steveb:
The devsetup edpm_baremetal target now calls to the parent directory make target edpm_deploy_baremetal. This means all of the duplicated variables can be deleted from devsetup/Makefile.

Eventually edpm_deploy_baremetal target can be deleted and the documentation and CI job cifmw-crc-podified-edpm-baremetal can call edpm_deploy_baremetal directly, but this will require some cross-repo dance including setting DATAPLANE_GROWVOLS_ARGS from the job config.

Also document EDPM baremetal deployment

PR #737, opened 2024-02-20
[baremetal] Fix setup with default values

karelyatin:
With current default disk-size=20 and edpm_growvols default
growvols_args='/=8GB /tmp=1GB /var/log=10GB /var/log/audit=2GB /home=1GB /var=100%'
deployment fails with insufficient space.

The issue got triggered after [1] which removed override from
install-yamls. To get it working readd defaults that works
with 20G disk-size.

Also DATAPLANE_GROWVOLS_ARGS was unused since [2], moved it
to correct target to allow custom disk layout.

[1] https://github.com/openstack-k8s-operators/install_yamls/pull/553
[2] https://github.com/openstack-k8s-operators/install_yamls/pull/604

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -85,7 +85,6 @@ BMAAS_ROUTE_LIBVIRT_NETWORKS ?= ${BMAAS_NETWORK_NAME},crc,default
 
 DATAPLANE_PLAYBOOK ?= osp.edpm.download_cache
 DATAPLANE_CUSTOM_SERVICE_RUNNER_IMG ?=quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
-DATAPLANE_GROWVOLS_ARGS    ?=
 BM_NETWORK_NAME            ?=default
 BM_INSTANCE_NAME_PREFIX    ?=edpm-compute
 BM_NODE_COUNT              ?=1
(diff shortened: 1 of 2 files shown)


devsetup/Makefile
```diff
@@ -59,7 +59,7 @@ BMAAS_OCP_INSTANCE_NAME ?= crc
 BMAAS_INSTANCE_NAME_PREFIX ?= crc-bmaas
 BMAAS_INSTANCE_MEMORY ?= 4096
 BMAAS_INSTANCE_VCPUS ?= 2
-BMAAS_INSTANCE_DISK_SIZE ?= 20
+BMAAS_INSTANCE_DISK_SIZE ?= 40
```
rabi, line 62: I don't think we should force users to use a bigger disk by default, but ensure  they have enough space in `/var` with `$DATAPLANE_GROWVOLS_ARGS`.

karelyatin: Ok so ok to readd default for DATAPLANE_GROWVOLS_ARGS which got removed by other patch?

karelyatin: @rabi so i updated it to re provide custom grow vars as before. Let me know in case it needs to be fixed differently.

rabi: sorry I missed this earlier.

rabi (commented):
/lgtm

PR #740, opened 2024-02-21
Add the crc_bmo_setup target to the baremetal target

slagle:
openstack-baremetal-operator requires the metal3 CRDs, otherwise the
operator won't start. This adds crc_bmo_setup which sets up metal3 to
the baremetal target if the CRDs are not already defined. This matches
the logic that is used to call the crc_bmo_setup target from
openstack_prep.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1968,6 +1968,9 @@ ansibleee_cleanup: ## deletes the operator, but does not cleanup the service res
 	${CLEANUP_DIR_CMD} ${OPERATOR_DIR}
 
 ##@ BAREMETAL
+
+BAREMETAL_PREP_DEPS := $(if $(findstring true,$(BMO_SETUP)), crc_bmo_setup)
+
 .PHONY: baremetal_prep
 baremetal_prep: export IMAGE=${BAREMETAL_IMG}
 baremetal_prep: ## creates the files to install the operator using olm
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1975,7 +1975,7 @@ baremetal_prep: ## creates the files to install the operator using olm
 	bash scripts/gen-olm.sh
 
 .PHONY: baremetal
-baremetal: operator_namespace baremetal_prep ## installs the operator, also runs the prep step. Set BAREMETAL_IMG for custom image.
+baremetal: operator_namespace crc_bmo_setup baremetal_prep ## installs the operator, also runs the prep step. Set BAREMETAL_IMG for custom image.
```
slagle, line 1978: i guess this means "make baremetal" would only work with crc, which is probably not what we want. open to other ideas.

slagle: i updated it to work the same way as openstack_prep, so I think it's ok now.

Makefile
```diff
@@ -1968,14 +1968,17 @@ ansibleee_cleanup: ## deletes the operator, but does not cleanup the service res
...
+
 .PHONY: baremetal_prep
 baremetal_prep: export IMAGE=${BAREMETAL_IMG}
 baremetal_prep: ## creates the files to install the operator using olm
 	$(eval $(call vars,$@,openstack-baremetal))
 	bash scripts/gen-olm.sh
 
 .PHONY: baremetal
-baremetal: operator_namespace baremetal_prep ## installs the operator, also runs the prep step. Set BAREMETAL_IMG for custom image.
+baremetal: operator_namespace $(BAREMETAL_PREP_DEPS) baremetal_prep ## installs the operator, also runs the prep step. Set BAREMETAL_IMG for custom image.
```
rabi, line 1981: Thanks! It works in CI prow jobs as multi-node compact cluster has CBO installed. Could have been  `BAREMETAL_DEPS` aligned with the target it's used.

rabi (approved):
/lgtm

PR #746, opened 2024-02-28
Shorten netattach name for octavia

beagles:
Having hyphenated names for interfaces gets in the way a couple of implementation options.


Akrog (approved):
I don´t have approving permission, but the change looks good to me.

fmount (approved):
/lgtm

PR #748, opened 2024-02-28
Adding Redis target and octavia dependency

gthiemonge:


Makefile
```diff
@@ -2515,3 +2521,30 @@ certmanager_cleanup:
...
+.PHONY: redis_deploy
+redis_deploy: input redis_deploy_prep ## installs the service instance using kustomize. Runs prep step in advance. Set INFRA_REPO and INFRA_BRANCH to deploy from a custom repo.
+	$(eval $(call vars,$@,infra))
+	make wait
+	bash scripts/operator-deploy-resources.sh
+
+.PHONY: redis_deploy_cleanup
+redis_deploy_cleanup: namespace ## cleans up the service instance, Does not affect the operator.
+	$(eval $(call vars,$@,infra))
+	oc kustomize ${DEPLOY_DIR} | oc delete --ignore-not-found=true -f -
```
beagles, line 2549: I wonder if we should just do an oc delete -f ${DEPLOY_DIR}/${REDIS_CR} here and remove it instead of wiping out the entire path. I'm just thinking of things like memcached that is also using this path.

beagles: Actually won't really work because it'll be missing the values inserted by the kustomize command like the namespace.

beagles (commented):
/lgtm

abays (approved):
/lgtm

PR #750, opened 2024-02-29
Allow kustomizing osContainerImageUrl

rabi:
This would allow to use a different image and dont have to patch csv for it.


abays (approved):
/lgtm

PR #751, opened 2024-02-29
Make server role configurable

cschwede:
Default is unchanged, but now it is possible to use a different role, eg:

OPENSTACK_DATAPLANENODESET=config/samples/dataplane_v1beta1_openstackdataplanenodeset_swift.yaml DATAPLANE_SERVER_ROLE=swift make edpm_wait_deploy


cschwede:
recheck

PR #752, opened 2024-02-29
Use release-0.5 branch for BMO

rabi:
metal3 has moved to go 1.21 recently and this has broken things as we still use go 1.20.  This aligns with bmo version we use in openstack-baremetal-operator[1].

```
cd apis; /home/zuul/ci-framework-data/artifacts/manifests/operator/baremetal-operator/tools/bin/controller-gen object:headerFile="../hack/boilerplate.go.txt" paths="./..."
metal3.io/v1alpha1/baremetalhost_validation.go:10:2: package slices is not in GOROOT (/usr/local/go/src/slices)
note: imported by a module that requires go 1.21
Error: not all generators ran successfully
```

[1] https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/main/go.mod#L9

PR #753, opened 2024-02-29
Bump golang version to 1.22.0

lpiwowar:
I'm getting the following error on the `make openstack` step. Bumping up the 1.22.0 seems to be resolving the problem. 
```
metal3.io/v1alpha1/baremetalhost_validation.go:10:2: package slices is not in GOROOT (/usr/local/go/src/slices)
note: imported by a module that requires go 1.21
```

### DIFF
--- a/devsetup/vars/default.yaml
+++ b/devsetup/vars/default.yaml
@@ -6,7 +6,7 @@ opm_version: v1.30.0
 sdk_version: v1.31.0
 
 # golang version
-go_version: 1.20.14
+go_version: 1.22.0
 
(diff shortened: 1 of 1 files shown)


lpiwowar:
recheck

gibizer:
the fix is in https://github.com/openstack-k8s-operators/install_yamls/pull/752 

we cannot bump to 1.22 just like this. we dont have compatible containers to build things with 1.22

gibizer:
this is not something we want or can have now

PR #759, opened 2024-03-04
Add storage_mgmt network to nncp for adoption

cescgina:
For adoption we need the storage_mgmt network to adopt swift before we
deploy the data-plane. This change also updates scripts/gen-netatt.sh to
make sure all networks use the same vlan ids everywhere.

### DIFF
--- a/scripts/gen-nncp.sh
+++ b/scripts/gen-nncp.sh
@@ -315,16 +315,60 @@ EOF_CAT
     fi
 
     #
-    # octavia-vlan-link VLAN interface
+    # storagemgmt VLAN interface
     #
     cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
-    - description: octavia-vlan-link vlan interface
+    - description: storagemgmt vlan interface
       name: ${INTERFACE}.$((${VLAN_START}+$((${VLAN_STEP}*3))))
       state: up
       type: vlan
       vlan:
         base-iface: ${INTERFACE}
         id: $((${VLAN_START}+$((${VLAN_STEP}*3))))
+        reorder-headers: true
+EOF_CAT
+    if [ -n "$IPV4_ENABLED" ]; then
+        cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
+      ipv4:
+        address:
+        - ip: 172.20.0.${IP_ADDRESS_SUFFIX}
+          prefix-length: 24
+        enabled: true
+        dhcp: false
+EOF_CAT
+    else
+        cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
+      ipv4:
+        enabled: false
+EOF_CAT
+    fi
+    if [ -n "$IPV6_ENABLED" ]; then
+        cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
+      ipv6:
+        address:
+        - ip: fd00:dede::${IPV6_ADDRESS_SUFFIX}
+          prefix-length: 64
+        enabled: true
+        dhcp: false
+        autoconf: false
+EOF_CAT
+    else
+        cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
+      ipv6:
+        enabled: false
+EOF_CAT
+    fi
+    #
+    # octavia-vlan-link VLAN interface
+    #
+    cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
+    - description: octavia-vlan-link vlan interface
+      name: ${INTERFACE}.$((${VLAN_START}+$((${VLAN_STEP}*4))))
+      state: up
+      type: vlan
+      vlan:
+        base-iface: ${INTERFACE}
+        id: $((${VLAN_START}+$((${VLAN_STEP}*4))))
         reorder-headers: true
 EOF_CAT
     if [ -n "$IPV4_ENABLED" ]; then
@@ -364,12 +408,12 @@ EOF_CAT
     #
     cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
     - description: designate vlan interface
-      name: ${INTERFACE}.$((${VLAN_START}+$((${VLAN_STEP}*4))))
+      name: ${INTERFACE}.$((${VLAN_START}+$((${VLAN_STEP}*5))))
       state: up
       type: vlan
       vlan:
         base-iface: ${INTERFACE}
-        id: $((${VLAN_START}+$((${VLAN_STEP}*4))))
+        id: $((${VLAN_START}+$((${VLAN_STEP}*5))))
         reorder-headers: true
 EOF_CAT
     if [ -n "$IPV4_ENABLED" ]; then
--- a/scripts/gen-netatt.sh
+++ b/scripts/gen-netatt.sh
@@ -224,6 +224,42 @@ spec:
     }
 EOF_CAT
 
+cat > ${DEPLOY_DIR}/storagemgmt.yaml <<EOF_CAT
+apiVersion: k8s.cni.cncf.io/v1
+kind: NetworkAttachmentDefinition
+metadata:
+  labels:
+    osp/net: storagemgmt
+  name: storagemgmt
+  namespace: ${NAMESPACE}
+spec:
+  config: |
+    {
+      "cniVersion": "0.3.1",
+      "name": "storagemgmt",
+      "type": "macvlan",
+      "master": "${INTERFACE}.$((${VLAN_START}+${VLAN_STEP}*3))",
+      "ipam": {
+        "type": "whereabouts",
+EOF_CAT
+if [ -n "$IPV4_ENABLED" ]; then
+    cat >> ${DEPLOY_DIR}/storagemgmt.yaml <<EOF_CAT
+        "range": "172.20.0.0/24",
+        "range_start": "172.20.0.30",
+        "range_end": "172.20.0.70"
+EOF_CAT
+elif [ -n "$IPV6_ENABLED" ]; then
+    cat >> ${DEPLOY_DIR}/storagemgmt.yaml <<EOF_CAT
+        "range": "fd00:dede::/64",
+        "range_start": "fd00:dede::30",
+        "range_end": "fd00:dede::70"
+EOF_CAT
+fi
+cat >> ${DEPLOY_DIR}/storagemgmt.yaml <<EOF_CAT
+      }
+    }
+EOF_CAT
+
 cat > ${DEPLOY_DIR}/octavia.yaml <<EOF_CAT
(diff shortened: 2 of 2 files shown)


scripts/run_openstack_kuttl.sh
```diff
@@ -0,0 +1 @@
+
```
fmount, line 1: has this been added by mistake?

fmount (commented):
wondering about the status of this patch that might be needed by [1]

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288

cescgina:
> wondering about the status of this patch that might be needed by [1]
> 
> [1] [openstack-k8s-operators/data-plane-adoption#288](https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288)

this is definitely needed for 288, it's ready for review, I need to fix the ipv6 addresses, but overall it needs review to ensure we're not breaking networking for other use cases

scripts/run_openstack_kuttl.sh
cescgina: yes,not sure what happened here, but will remove this

hjensas:
> > wondering about the status of this patch that might be needed by [1]
> > [1] [openstack-k8s-operators/data-plane-adoption#288](https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288)
> 
> this is definitely needed for 288, it's ready for review, I need to fix the ipv6 addresses, but overall it needs review to ensure we're not breaking networking for other use cases

I don't understand why the storage_mgmt network is needed by [openstack-k8s-operators/data-plane-adoption#288](https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288)? It does not attach any pods to the storage_mgmt network? Can you not use a route instead, as documented in the [adoption guide](https://openstack-k8s-operators.github.io/data-plane-adoption/dev/#_route_networks)? (AFICT from tripleo-heat-templates, there are no firewall rules blocking based on source address for swift, so a route should work?)

If you want to connect some adoption pod, or the switft pods in the new controlplane you also need to add the networkAttachmentDefinition in `scripts/gen-netatt.sh` and possibly also something in `scripts/gen-metallb-config.sh`?

cescgina:
> > > wondering about the status of this patch that might be needed by [1]
> > > [1] [openstack-k8s-operators/data-plane-adoption#288](https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288)
> > 
> > 
> > this is definitely needed for 288, it's ready for review, I need to fix the ipv6 addresses, but overall it needs review to ensure we're not breaking networking for other use cases
> 
> I don't understand why the storage_mgmt network is needed by [openstack-k8s-operators/data-plane-adoption#288](https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288)? It does not attach any pods to the storage_mgmt network? Can you not use a route instead, as documented in the [adoption guide](https://openstack-k8s-operators.github.io/data-plane-adoption/dev/#_route_networks)? (AFICT from tripleo-heat-templates, there are no firewall rules blocking based on source address for swift, so a route should work?)
> 
> If you want to connect some adoption pod, or the switft pods in the new controlplane you also need to add the networkAttachmentDefinition in `scripts/gen-netatt.sh` and possibly also something in `scripts/gen-metallb-config.sh`?
> 
> And yes, please add IPv6 the same way it is done for the other networks.

I believe this was needed because in https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288 swift is used as backend for glance, and it needs/expects to talk to the standalone swift service over the storage_mgmt network, but I could be wrong about that. What's for sure is that until we added this change, swift adoption would fail.

fmount:
> > > > wondering about the status of this patch that might be needed by [1]
> > > > [1] [openstack-k8s-operators/data-plane-adoption#288](https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288)
> > > 
> > > 
> > > this is definitely needed for 288, it's ready for review, I need to fix the ipv6 addresses, but overall it needs review to ensure we're not breaking networking for other use cases
> > 
> > 
> > I don't understand why the storage_mgmt network is needed by [openstack-k8s-operators/data-plane-adoption#288](https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288)? It does not attach any pods to the storage_mgmt network? Can you not use a route instead, as documented in the [adoption guide](https://openstack-k8s-operators.github.io/data-plane-adoption/dev/#_route_networks)? (AFICT from tripleo-heat-templates, there are no firewall rules blocking based on source address for swift, so a route should work?)
> > If you want to connect some adoption pod, or the switft pods in the new controlplane you also need to add the networkAttachmentDefinition in `scripts/gen-netatt.sh` and possibly also something in `scripts/gen-metallb-config.sh`?
> > And yes, please add IPv6 the same way it is done for the other networks.
> 
> I believe this was needed because in [openstack-k8s-operators/data-plane-adoption#288](https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288) swift is used as backend for glance, and it needs/expects to talk to the standalone swift service over the storage_mgmt network, but I could be wrong about that. What's for sure is that until we added this change, swift adoption would fail.

The current swift adoption only moves `swift-proxy` but leave `swift-storage` in the controller nodes. This means that the swift proxy in the ctlplane needs to reach the swift-storage in the controller nodes. If we do not propagate such network (by routing it or just extending the same way we do for the others), the swift adoption won't work, and the patch data-plane-adoption#288 will fail because we missed a requirement.
@hjensas if you have a solution in mind I agree we can take another path (which is out of scope for me right now), but the real question here is whether this patch is good or introduces a regression. wdyt?

scripts/gen-nncp.sh
```diff
@@ -331,7 +331,7 @@ EOF_CAT
         cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
       ipv4:
```
hjensas, line 332: We should add IPv6 here, if not I am worried this will break the IPv6 test environment.

hjensas (changes requested):
If we add the `storage_mgtm` network on the node to allow swift-proxy to reach the contollers on this network - should we not connected the `storage_mgmt` network to the switf-proxy as well? See: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/swift_adoption/tasks/main.yaml#L67

If `storage_mgmt` is adding just the NNCP - but swift-proxy is still connected only on `storage` I belive what this is doing is adding a route. We could then add that route in the infrastructure instead e.g on the hypervisor node as it is documnted in the adoption guide.

Adding @cschwede 


My main issue with this is that it does not implement IPv6 like we do for all the other networks. We should do that to avoid breaking that functionality.

scripts/gen-nncp.sh
cescgina: it has an IPv6 section, just like the other networks, it's on  line 345, it's just hidden by github UI because the line did not change. Now, I'm not sure if the subnet should have a particular value, but it's there

hjensas: sorry, I missed that. 
Is changing the IPv6 prefix for `octavia` ok?
Prior to this change it was `fd00:eeee::` it is `fd00:eded::` after this change.

The `fd00:eeee::` is also in several other files - so it is going to break octavia?
```
scripts/gen-nncp.sh:        - ip: fd00:eeee::${IPV6_ADDRESS_SUFFIX}
scripts/gen-service-kustomize.sh:          - end: fd00:eeee::250
scripts/gen-service-kustomize.sh:            start: fd00:eeee::100
scripts/gen-service-kustomize.sh:          cidr: fd00:eeee::/64
scripts/gen-netatt.sh:        "range": "fd00:eeee::/64",
scripts/gen-netatt.sh:        "range_start": "fd00:eeee::30",
scripts/gen-netatt.sh:        "range_end": "fd00:eeee::70"
out/openstack/nncp/cr/api.sno.lab.example.com_nncp.yaml:        - ip: fd00:eeee::5
out/openstack/infra/cr/kustomization.yaml:          - end: fd00:eeee::250
out/openstack/infra/cr/kustomization.yaml:            start: fd00:eeee::100
out/openstack/infra/cr/kustomization.yaml:          cidr: fd00:eeee::/64
```

hjensas:
> @hjensas one question about the creating a route approach. IIUC the adoption docs assume that both the crc and standalone instances are running in a vm inside the same hypervisor, but that is no longer the case in CI, they are running in different vms. Would it work to create a route in the crc node with the `ip` commands?

No, if this was switched to use the extracted CRC then it is probably easier to add this network to the node.

scripts/gen-nncp.sh
cescgina: I missed that, thanks, it's probably wrong, I'll swap the subnets for octavia and storage_mgmt

scripts/gen-nncp.sh
```diff
@@ -315,17 +315,61 @@ EOF_CAT
     fi
 
     #
-    # octavia-vlan-link VLAN interface
+    # storage_mgmt VLAN interface
     #
     cat >> ${DEPLOY_DIR}/${WORKER}_nncp.yaml <<EOF_CAT
-    - description: octavia-vlan-link vlan interface
+    - description: storage_mgmt vlan interface
       name: ${INTERFACE}.$((${VLAN_START}+$((${VLAN_STEP}*3))))
```
hjensas, line 322: This is swapping the VLAN ID's around as well?


In `scripts/gen-netatt.sh`:
```
      "name": "octavia",
      "type": "macvlan",
      "master": "${INTERFACE}.$((${VLAN_START}+${VLAN_STEP}*3))",

      "name": "designate",
      "type": "macvlan",
      "master": "${INTERFACE}.$((${VLAN_START}+${VLAN_STEP}*4))",
```


Now we are inserting storage_mgmt at `${VLAN_STEP}*3` and shifting octavia and disignate? But that is not updated in all the other files?

I think we should add the `storage_mgmt` network at the end and use ${VLAN_STEP}*5?

cescgina: yes, I put it there to match the vlan id from the standalone deployment, where storage_mgmt has vlan 23. We can either change the vlans in `scripts/gen-netatt.sh` or the standalone network config to keep the current order and add `storage_mgmt` at the end as you suggest. Judging by the ip ranges, the storage_mgmt seemed like it should go after the `tenant` network and before the octavia one

hjensas: IDK - it seems like we have a complete mess here?

The `NetConfig` defines `storagemgmt` as VLAN 23.
https://github.com/openstack-k8s-operators/infra-operator/blob/main/config/samples/network_v1beta1_netconfig.yaml#L45-L53

I think that is the one used based on `Makefile:NETCONFIG     ?= config/samples/network_v1beta1_netconfig.yaml`?

But in NNCP we are currently using 23 for `octavia`?

so probably changing octavia to use a diferent VLAN is the best way to go? I think that would be only changes in install_yamls repo `gen-nncp.sh` and `gen-netatt.sh`?

cescgina: yes, many moving parts here, let me update the patch to set the same vlan ids in gen-nncp and net-netatt

hjensas (approved):
I think this is good now.

fmount (approved):

PR #763, opened 2024-03-08
Get projects using the api explicitely

cescgina:
In CI we are seeing intermittently that running 'oc get project
project-name' fails because openshift calls the config.openshift.io/v1
api instead of the project.openshift.io/v1 as expected.

JIRA: https://issues.redhat.com/browse/OSPCIX-193

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -589,7 +589,7 @@ operator_namespace: ## creates the namespace specified via OPERATOR_NAMESPACE en
 	$(eval $(call vars,$@))
 	bash scripts/gen-namespace.sh
 	oc apply -f ${OUT}/${OPERATOR_NAMESPACE}/namespace.yaml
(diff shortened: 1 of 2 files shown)


cescgina:
recheck

lewisdenny (approved):
/lgtm
I'm not sure why `oc` is hitting the config api, maybe we check so early that the OCP projects API isn't up yet and it caches that so even when it comes up we still use config?

Anyway, `projects.v1.project.openshift.io` is what we want to hit.

```
[core@crc-pjmnl-master-0 ~]$ oc get projects.v1.project.openshift.io | grep openstack
openshift-openstack-infra                                         Active
openstack                                                         Active
openstack-operators                                               Active
[core@crc-pjmnl-master-0 ~]$ oc get projects.v1.config.openshift.io | grep openstack
```

fao89 (approved):

PR #764, opened 2024-03-08
Remove repo-setup service

fao89:
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1241


scripts/gen-edpm-baremetal-kustomize.sh
```diff
@@ -50,8 +50,10 @@ patches:
       path: /spec/baremetalSetTemplate/bmhNamespace
       value: ${EDPM_BMH_NAMESPACE}
     - op: add
-      path: /spec/services/0
-      value: repo-setup
+      path: /spec/nodeTemplate/ansible/ansibleVarsFrom/0
+      value:
+        configMapRef:
+          name: repo-setup-bootstrap
```
raukadah, line 56: @fao89 Can we keep the service name same as `repo-setup` otherwise it will break periodic baremetal and hci job? Both uses repo-setup service and perform downstream kustomization.
[1]. https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/edpm_deploy_baremetal/tasks/main.yml#L188
[2]. https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/hci_prepare/tasks/phase1.yml#L63

PR #765, opened 2024-03-11
Use $BMH_NAMESPACE when cleaning BMHs

rabi:
Also, we don't need to cleanup OpenStackProvisionServer as it would be cleaned up with OpenStackBaremetalSet.

### DIFF
--- a/devsetup/scripts/edpm-compute-baremetal.sh
+++ b/devsetup/scripts/edpm-compute-baremetal.sh
@@ -97,8 +97,7 @@ function cleanup {
     while oc get bmh | grep -q -e "deprovisioning" -e "provisioned"; do
         sleep 5
     done || true
-    oc delete --all bmh --ignore-not-found=true || true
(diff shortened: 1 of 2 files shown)


devsetup/scripts/edpm-compute-baremetal.sh
```diff
@@ -97,8 +97,7 @@ function cleanup {
     while oc get bmh | grep -q -e "deprovisioning" -e "provisioned"; do
         sleep 5
     done || true
-    oc delete --all bmh --ignore-not-found=true || true
-    oc delete --all openstackprovisionserver --ignore-not-found=true || true
+    oc delete --all -n bmh $NAMESPACE --ignore-not-found=true || true
```
abays, line 100: I think `-n bmh $NAMESPACE` needs to be `-n $NAMESPACE bmh` instead

rabi: Thanks for spotting that :smile:

abays (approved):
/lgtm

PR #768, opened 2024-03-11
Support deploying standalone with a different ctlplane subnet

booxter:
For adoption purposes, we would like to be able to, first, deploy standalone using a subnet that is different from the subnet used for the same vlan in podified setup; then, use tripleo mechanism to add IP addresses in the new podified subnet; finally, run EDPM adoption, carefully configuring IP addresses for each node in a set so that both addresses are configured.

During adoption phase, the tripleo setup will carry IP addresses in both subnets, 1) for old subnet - to continue running old services on the original endpoints; 2) for new subnet - to run edp deployment and to spawn new services in.

(Post-adoption, the old subnet addresses can be dropped from EDP definitions.)

To deploy standalone with a different ctlplane subnet, export CTLPLANE_IP different from the IP address extracted from net-dumpxml.


booxter:
Legit failure in CI

```
jinja2.exceptions.UndefinedError: 'additional_ctlplane_ip' is undefined
```

booxter:
recheck

devsetup/scripts/standalone.sh
```diff
@@ -117,7 +118,7 @@ export NTP_SERVER=${NTP_SERVER:-"clock.corp.redhat.com"}
 export EDPM_COMPUTE_CEPH_ENABLED=${EDPM_COMPUTE_CEPH_ENABLED:-true}
 export CEPH_ARGS="${CEPH_ARGS:--e \$HOME/deployed_ceph.yaml -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm-rbd-only.yaml -e \$HOME/nova_noceph.yaml}"
 export COMPUTE_DRIVER=${COMPUTE_DRIVER:-"libvirt"}
-export IP=${IP}
+export IP=${CTLPLANE_IP}
```
jistr, line 121: I find the flow of values somewhat confusing... First we assign CTLPLANE_IP=IP, then we assign IP=CTLPLANE_IP... Could we just define a new variable ADDITIONAL_CTLPLANE_IP and below in the ansible variables have

```
additional_ctlplane_ip: ${ADDITIONAL_CTLPLANE_IP}
```
?

jistr: And the `{% if additional_ctlplane_ip is defined %}` could perhaps become `{% if additional_ctlplane_ip is defined and additional_ctlplane_ip %}`? I'm not sure if it's necessary but it also rules out when the additional IP variable is defined but `null` or empty string.

devsetup/scripts/standalone.sh
```diff
@@ -31,6 +31,7 @@ if [[ ${STANDALONE_VM} == "true" ]]; then
 fi
 IP_ADRESS_SUFFIX=${IP_ADRESS_SUFFIX:-"$((100+${EDPM_COMPUTE_SUFFIX}))"}
 IP=${IP:-"${EDPM_COMPUTE_NETWORK_IP%.*}.${IP_ADRESS_SUFFIX}"}
+CTLPLANE_IP=${CTLPLANE_IP:-"${IP}"}
```
marios, line 34: why do you need this override. I mean, why can't you use and set "IP" to the value that you want. the non standard ctlplane network.

devsetup/standalone/net_config.j2
```diff
@@ -14,6 +14,10 @@ network_config:
   - ip_netmask: {{ ctlplane_ip }}/{{ ctlplane_cidr }}
   - ip_netmask: {{ ctlplane_ip }}/32
   - ip_netmask: {{ ctlplane_vip }}/32
+  {% if additional_ctlplane_ip is defined %}
```
marios, line 17: so do you also need to keep the 'original' 192.168.122.100 (default) as well?

booxter:
This is superseded by https://github.com/openstack-k8s-operators/install_yamls/pull/774

PR #769, opened 2024-03-12
Run standalone optionally with three Swift replicas

cschwede:
If standalone is running with a single replica, Swift replicators and rsync won't be started and data migration is therefore not possible.

This patch runs Swift optionally with three separate disks and three replicas, thus enabling replicators to run and test migration workloads. To achieve this, three additional disks are created and attached to the edpm nodes. The disk images are sparse, thus not using much space as long as there is no data stored within.

The additional disks are not only useful for the standalone node itself, but also for dataplane nodes to test with multiple disks.

The default is unchanged; this feature needs to be explicitly enabled.

### DIFF
--- a/devsetup/scripts/gen-edpm-node.sh
+++ b/devsetup/scripts/gen-edpm-node.sh
@@ -25,6 +25,7 @@ trap 'rm -rf -- "$MY_TMP_DIR"' EXIT
 EDPM_SERVER_ROLE=${EDPM_SERVER_ROLE:-"compute"}
 
 STANDALONE=${STANDALONE:-false}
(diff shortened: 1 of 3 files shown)


cschwede:
> The default is unchanged; this feature needs to be explicitly enabled.

I'm actually wondering if it might be a better option to enable this by default? As long as the additional disks are not used this adds ~600kB per VM of disk storage, and not much more. However, enabled replicators in Swift are required to properly test data migration when adopting, thus enabling this becomes a necessity in that case. And the purpose of the standalone VM is to test adoption

fao89 (approved):

PR #770, opened 2024-03-12
adress bugs found when enabling tls-e

SeanMooney:
this commit has 2 primary changes

first newer version of yq require -y or -Y
to be passed when invoking yq with -i

both -y and -Y ensure that the final output
format of yq after modifying a file in place is in yaml
formant. yq is just a thin wraper around jq and transforms
the data into json before mofifying it and in future verison
-i will error if we do not specify -y or -Y.

the second change is a follow up to the naming change of the
network attachments. while testing the tls-e feauter for the
edpm nodes i notice that some fo the inventory fields were not
generated properly when the upper case names were used.
to avoid that i have just normalised the names to lowercase.

### DIFF
--- a/scripts/gen-edpm-kustomize.sh
+++ b/scripts/gen-edpm-kustomize.sh
@@ -57,30 +57,30 @@ patches:
     - op: replace
       path: /spec/nodes/edpm-${EDPM_SERVER_ROLE}-0/networks
       value:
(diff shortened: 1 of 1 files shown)


SeanMooney:
The yq in CentOS is too old for -y...

fao89:
recheck

PR #772, opened 2024-03-13
Fix IPv6 Lab wildcard DNS

hjensas:
The wildcard for *.sno.lab.example.com is incorrect, and cause issues with internal dns lookups.


fao89 (approved):

stuggi:
recheck

PR #773, opened 2024-03-13
Make ceph device persistent

bogdando:
Persist /dev/vg2/data-lv2 as a systemd service to align ceph deployment with that ci-framework does for Ceph deployment on standalone tripleo.

Add standalone_revert.sh script to ensure the time is synchronized and /dev/vg2/data-lv2 device is recreated, after restoring VM from the clean snapshot.

Add env vars to allow ssh commands functional after revert is done (for Makefile targets standalone_deploy and standalone_revert).

### DIFF
--- a/devsetup/scripts/standalone_revert.sh
+++ b/devsetup/scripts/standalone_revert.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+# set -x
+
+export VIRSH_DEFAULT_CONNECT_URI=qemu:///system
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+EDPM_COMPUTE_SUFFIX=${1:-"0"}
+EDPM_COMPUTE_NETWORK=${EDPM_COMPUTE_NETWORK:-default}
+STANDALONE_VM=${STANDALONE_VM:-"true"}
+if [[ ${STANDALONE_VM} == "true" ]]; then
+    EDPM_COMPUTE_NETWORK_IP=$(virsh net-dumpxml ${EDPM_COMPUTE_NETWORK} | xmllint --xpath 'string(/network/ip/@address)' -)
+fi
+IP_ADRESS_SUFFIX=${IP_ADRESS_SUFFIX:-"$((100+${EDPM_COMPUTE_SUFFIX}))"}
+IP=${IP:-"${EDPM_COMPUTE_NETWORK_IP%.*}.${IP_ADRESS_SUFFIX}"}
+OUTPUT_DIR=${OUTPUT_DIR:-"${SCRIPTPATH}/../../out/edpm/"}
+SSH_KEY_FILE=${SSH_KEY_FILE:-"${OUTPUT_DIR}/ansibleee-ssh-key-id_rsa"}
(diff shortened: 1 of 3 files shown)


bogdando:
recheck rdoproject.org/github-check no logs

bogdando:
recheck rdoproject.org/github-check no logs

fao89:
recheck

devsetup/standalone/ceph.sh
```diff
@@ -31,6 +31,27 @@ sudo pvcreate /dev/loop3
...
+ExecStop=/sbin/losetup -d /dev/loop3
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
+EOF
+cat /tmp/ceph-osd-losetup.service | sudo tee /etc/systemd/system/ceph-osd-losetup.service
+sudo chmod 0644 /etc/systemd/system/ceph-osd-losetup.service
+sudo systemctl daemon-reload
+sudo systemctl enable --now ceph-osd-losetup.service
```
jistr, line 53: Why is this needed? In my env i can snapshot-restore using the make targets and the loopback device is still present. Are you snapshotting some other way?

devsetup/scripts/standalone_revert.sh
```diff
@@ -0,0 +1,20 @@
...
+    EDPM_COMPUTE_NETWORK_IP=$(virsh net-dumpxml ${EDPM_COMPUTE_NETWORK} | xmllint --xpath 'string(/network/ip/@address)' -)
+fi
+IP_ADRESS_SUFFIX=${IP_ADRESS_SUFFIX:-"$((100+${EDPM_COMPUTE_SUFFIX}))"}
+IP=${IP:-"${EDPM_COMPUTE_NETWORK_IP%.*}.${IP_ADRESS_SUFFIX}"}
+OUTPUT_DIR=${OUTPUT_DIR:-"${SCRIPTPATH}/../../out/edpm/"}
+SSH_KEY_FILE=${SSH_KEY_FILE:-"${OUTPUT_DIR}/ansibleee-ssh-key-id_rsa"}
+SSH_OPT="-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i $SSH_KEY_FILE"
+
+virsh snapshot-revert --domain edpm-compute-${EDPM_COMPUTE_SUFFIX} --snapshotname clean
+ssh $SSH_OPT root@$IP systemctl stop chronyd ';' chronyd -q  \'pool pool.ntp.org iburst\' ';' systemctl start chronyd
```
jistr, line 19: +1 on setting the clock, which doesn't seem to be done automatically after reverting from the snapshot. 

But why the manual setting with pool.ntp.org? That will get blocked inside our network. For me just `systemctl restart chronyd` seems to bring the clock up to date (and it's using the configured NTP server).

devsetup/standalone/ceph.sh
bogdando: testing shows that ensuring the unit is started after reboot is sufficient. Reverting doesn't cause problems here. So I will rework this

devsetup/scripts/standalone_revert.sh
bogdando: good point. I think we can split this off into a different patch

PR #776, opened 2024-03-14
Add extra routes to the octavia network attachment

beagles:
These routes are used internally to setup bidirectional routes between the load balancer management network in the openstack cloud and the amphora controllers.


bshephar (approved):
This `octavia.yaml` sure looks a lot like json. :)

PR #777, opened 2024-03-14
Set value for RABBITMQ_DEPL_IMG

dprince:
Set to: quay.io/podified-antelope-centos9/openstack-rabbitmq:current-podified


stuggi:
the image might not work without changes removing kolla. a test from @Deydra71 showed:

```
$ oc logs rabbitmq-server-0 -n octavia-kuttl-tests
Defaulted container "rabbitmq" out of: rabbitmq, setup-container (init)
+ sudo -E kolla_set_configs
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
```

PR #742, opened 2024-02-22
Octavia services depend on nova and neutron

beagles:
The operator is needs nova and neutron to deploy the amphora controller services.

PR #779, opened 2024-03-15
Octavia services depend on placement

Deydra71:
As nova has been added to the dependency list in #742, placement must be added as well for nova to be successfully deployed.


stuggi (approved):
/lgtm nova requires placement

PR #780, opened 2024-03-18
Change Designate subnet

bshephar:
The current Designate subnet conflicts with the default serviceNetwork in OpenShift:
https://github.com/openshift/installer/blob/0eafdbb77fd62f8311ba8d9abda145f3280c5f79/docs/user/customization.md\?plain\=1\#L56

This change moves the designate subnet we're using from 172.30.0.0/24 to 172.24.0.0/24 to avoid this conflict.

### DIFF
--- a/scripts/gen-netatt.sh
+++ b/scripts/gen-netatt.sh
@@ -277,9 +277,9 @@ spec:
       "master": "${INTERFACE}.$((${VLAN_START}+${VLAN_STEP}*4))",
       "ipam": {
         "type": "whereabouts",
-        "range": "172.30.0.0/24",
-        "range_start": "172.30.0.30",
-        "range_end": "172.30.0.70"
+        "range": "172.28.0.0/24",
+        "range_start": "172.28.0.30",
+        "range_end": "172.28.0.70"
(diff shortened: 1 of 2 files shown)


scripts/gen-netatt.sh
```diff
@@ -277,9 +277,9 @@ spec:
       "master": "${INTERFACE}.$((${VLAN_START}+${VLAN_STEP}*4))",
       "ipam": {
         "type": "whereabouts",
-        "range": "172.30.0.0/24",
-        "range_start": "172.30.0.30",
-        "range_end": "172.30.0.70"
+        "range": "172.24.0.0/24",
```
beagles, line 280: where it is on it's own vlan it might not be a problem but this is the same range as octavia

bshephar: Ah damn, let me change it. I'll take a better look at what we're already using and make sure it doesn't conflict with anything else.

bshephar: `172.28` seems to be free.

fao89 (approved):

PR #781, opened 2024-03-19
Add routes to the net-iso bridge

hjensas:
Add route so that the nat64 router knows where to forward fd00:aaaa::/64 traffic.

Add route on the host to ensure nat64 IPv6 prefix is routed to the nat64 router.

Use forwarding mode "open" when using IPv6 - this turns off all the Libvirt firewalling that otherwise get's in the way.

Alos remove some of the options, we use env vars to drive it.


morucci:
recheck

rabi (approved):
/lgtm

rabi:
recheck

rabi (approved):
/lgtm

fao89:
recheck

rabi:
recheck

fao89:
recheck

rabi:
recheck

PR #786, opened 2024-03-25
Fix bmaas multiple nodes libvirt

hjensas:
The sed experessions used to make the list of virsh node uuids a comma separated list does not work. Replace it with awk.


hjensas:
recheck

hjensas:
recheck

rabi (approved):
/lgtm

PR #788, opened 2024-03-26
Change octavia network attachment to something bridgeable

beagles:
The old host route was incorrectly colliding with the IP of the host interface for the macvlan network attachment. This modifies the target route to an IP on the high part of the range with the aim of avoiding IP conflicts between routing, host and pod IPs.

Related: #[OSPRH-5827](https://issues.redhat.com//browse/OSPRH-5827)


abays (approved):
/lgtm

PR #790, opened 2024-03-27
Standalone - Ironic - Change IPA URI

hjensas:
https://images.rdoproject.org has been retired.
Use IPA images from the https://tarballs.opendev.org instead.

### DIFF
--- a/devsetup/standalone/post_config/ironic.sh
+++ b/devsetup/standalone/post_config/ironic.sh
@@ -16,11 +16,14 @@
 set -ex
 
 export OS_CLOUD="standalone"
-IRONIC_PYTHON_AGENT_URI=${IRONIC_PYTHON_AGENT_URI:-"https://images.rdoproject.org/centos9/master/rdo_trunk/current-tripleo/ironic-python-agent.tar"}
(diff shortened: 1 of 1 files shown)


hjensas:
recheck

devsetup/standalone/post_config/ironic.sh
```diff
@@ -16,33 +16,36 @@
 set -ex
 
 export OS_CLOUD="standalone"
-IRONIC_PYTHON_AGENT_URI=${IRONIC_PYTHON_AGENT_URI:-"https://images.rdoproject.org/centos9/master/rdo_trunk/current-tripleo/ironic-python-agent.tar"}
+IRONIC_PYTHON_AGENT_URI=${IRONIC_PYTHON_AGENT_URI:-"https://tarballs.opendev.org/openstack/ironic-python-agent/dib/ipa-centos8-stable-wallaby.tar.gz"}
```
olliewalsh, line 19: why not one of the centos9 images?

olliewalsh: ack, and I guess needs to be wallaby for standalone

olliewalsh (approved):
/lgtm

devsetup/standalone/post_config/ironic.sh
hjensas: It might work with a later release, but the risk is that something changed in the API so a newer IPA may fail. Safer to use wallaby IPA on CS8. :+1:

PR #791, opened 2024-04-02
Add quotes around MAC address

hjensas:
Something changed in, so a plain MAC address in a scalar is converted to integer.

openstack baremetal create /root/ironic_nodes.yaml where yaml contains:
```
  ports:
  - address: 52:54:00:00:40:07
```

Fails with below error, adding quotes fixes the issue.
```
Unable to create the port with the specified attributes:
  {'address': 41135042407, 'node_uuid': '53ef5eef-5e36-4f41-8dbe-73618bd5fcdf'}.
  The error is: Schema error for port: 41135042407 is not of type 'string'
  Failed validating 'type' in schema['properties']['address'] (HTTP 400)
```


abays:
recheck

abays (approved):
/lgtm

PR #792, opened 2024-04-03
edpm_deploy_instance create floating ip before server creation

auniyal61:
creating and assigning floating ip takes some time, hence [edpm-deploy-instance.sh](https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/edpm-deploy-instance.sh)  often fails (but not always) and script exit with error code 1, even we have successful server creation and floating ip too exists.
this change simply move `floating ip create` before `server create`.

Also adds appropriate timeout to openstack_wait_deploy and edpm_wait_deploy cmds


auniyal61:
This patch is not updated from some time.
also, I tried the same script recently (today), and did not see issue. 
I'll close this.

PR #793, opened 2024-04-05
Patch network.operator when OVNKubernetes

karelyatin:
Latest CRC(v2.34.1)[1] switched to OpenShift 4.15
and OVNKubernetes. With default installation
MetalLB on secondary interfaces and local host
routing do not work[2][3], so we need to patch
network.operator to get these work.

[1] https://github.com/crc-org/crc/releases/tag/v2.34.1
[2] https://issues.redhat.com/browse/OSPRH-3860
[3] https://issues.redhat.com/browse/OSPRH-2457


bogdando (commented):
good catch lgtm

abays:
recheck

```
RETRY_LIMIT
```

PR #796, opened 2024-04-10
Add support for installing certmanager on OCP 4.15

danpawlik:
The command `make certmanager` does not work on CRC 2.34.1 which is based on OCP 4.15.3.
This commit is adding one more condition and set the certmanager version to 1.13 when it see OCP version 4.15.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2521,7 +2521,7 @@ swift_deploy_cleanup: ## cleans up the service instance, Does not affect the ope
 .PHONY: certmanager
 certmanager: export NAMESPACE=$(if $(findstring 4.10,$(OCP_RELEASE)),openshift-cert-manager,cert-manager)
 certmanager: export OPERATOR_NAMESPACE=$(if $(findstring 4.10,$(OCP_RELEASE)),openshift-cert-manager-operator,cert-manager-operator)
-certmanager: export CHANNEL=$(if $(findstring 4.10,$(OCP_RELEASE)),tech-preview,stable-v1.11)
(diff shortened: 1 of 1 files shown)


lewisdenny (approved):
/lgtm :)

Makefile
```diff
@@ -2521,7 +2521,7 @@ swift_deploy_cleanup: ## cleans up the service instance, Does not affect the ope
 .PHONY: certmanager
 certmanager: export NAMESPACE=$(if $(findstring 4.10,$(OCP_RELEASE)),openshift-cert-manager,cert-manager)
 certmanager: export OPERATOR_NAMESPACE=$(if $(findstring 4.10,$(OCP_RELEASE)),openshift-cert-manager-operator,cert-manager-operator)
-certmanager: export CHANNEL=$(if $(findstring 4.10,$(OCP_RELEASE)),tech-preview,stable-v1.11)
+certmanager: export CHANNEL=$(if $(findstring 4.10,$(OCP_RELEASE)),tech-preview,$(if $(findstring 4.15,$(OCP_RELEASE)),stable-v1.13,stable-v1.11))
```
karelyatin, line 2524: would be better to cleanup 4.10 checks itself here and above and in cert_manager_cleanup to be more cleaner, which is already EOL, also with other part of code it will not work on older OCPs

karelyatin: yes sure

karelyatin:
/override ci/prow/install-yamls-deploy

CI don't have 4.15 yet so it's not tested

PR #804, opened 2024-04-16
Revert "Set a domain on generated EDPM nodes"

olliewalsh:
Setting /etc/hostname to an FQDN with the wrong domain by default is worse than setting it to the short name.

The JIRA has also been moved back to refinement.

Revert for now and return to it later.

Related: [OSPRH-6187](https://issues.redhat.com//browse/OSPRH-6187)
Related: [OSPRH-6173](https://issues.redhat.com//browse/OSPRH-6173)

This reverts commit 3be98b012a4daa58ddddf3bed6dff5e9b2006185.


rabi:
> Setting /etc/hostname to an FQDN with the wrong domain by default is worse than setting it to the short name.

If for some reason you to change the hostname to a fqdn like this (for pre-provisioned nodes), you also have to set it in nodeset spec as well for the certs to be valid like we do in CI[1].

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/hooks/playbooks/fetch_compute_facts.yml#L136
https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/multinode-customizations.yml#L278

PR #805, opened 2024-04-16
[WIP] Add ceph to tripleo deployment

cescgina:


fultonj (commented):
Rather than supplying the entire `overcloud_ceph_roles.yaml` file, the script could be modified to generate it using `openstack overcloud roles generate Controller ComputeHCI > roles.yaml`. This could make the PR smaller.

devsetup/tripleo/ceph.sh
```diff
@@ -0,0 +1,58 @@
...
+if [[ ! -d "$HOME/ci-framework" ]]; then
+    git clone https://github.com/openstack-k8s-operators/ci-framework.git
+fi
+
+cd ci-framework
+# create block devices on all compute nodes
+ansible-playbook -i $INV playbooks/ceph.yml --tags block -e num_osds=3
+cd ..
+
+# TODO jgilaber, will need to reduce the amount of block devices when we
```
fultonj, line 35: It's OK to have 9 OSDs if you have 3 computes. You could reduce it later if you need to for some reason, but it's not an essential todo.

devsetup/tripleo/generate_ceph_inventory.py
```diff
@@ -0,0 +1,143 @@
...
+import argparse
+import re
+import sys
+import yaml
+
+
+def parse_opts(argv):
+    parser = argparse.ArgumentParser(
+            description='Create an ansible inventory from a yaml file similar'
+                        'to the one output by `openstack overcloud node provision`.'
```
fultonj, line 23: We need this script because "deployed ceph" works with the output of `openstack overcloud node provision` but we do not have metalsmith here. Thus, we're using the script to give us what metalsmith would have given us.

devsetup/tripleo/ceph.sh
cescgina: ack, I saw the comment in your README and though it was necessary, I'll remove it in my next update

fultonj:
> > Rather than supplying the entire `overcloud_ceph_roles.yaml` file, the script could be modified to generate it using `openstack overcloud roles generate Controller ComputeHCI > roles.yaml`. This could make the PR smaller.
> 
> yes, that was the way I had it orginally, but then I needed a few sed commands to tweak some fields, so I thought this would be simpler

I think though that this:
```
openstack overcloud roles generate Controller ComputeHCI \
 | sed -e /// \
  -e /// \
  ...
  -e /// > roles.yaml
```
is better than an extra 250 lines of yaml though. Not sure how much sed you had to do. I don't expect `openstack overcloud roles generate` to change much, but if changes and breaks the job, maybe that's something we actually want to catch?

cescgina:
> > > Rather than supplying the entire `overcloud_ceph_roles.yaml` file, the script could be modified to generate it using `openstack overcloud roles generate Controller ComputeHCI > roles.yaml`. This could make the PR smaller.
> > 
> > 
> > yes, that was the way I had it orginally, but then I needed a few sed commands to tweak some fields, so I thought this would be simpler
> 
> I think though that this:
> 
> ```
> openstack overcloud roles generate Controller ComputeHCI \
>  | sed -e /// \
>   -e /// \
>   ...
>   -e /// > roles.yaml
> ```
> 
> is better than an extra 250 lines of yaml though. Not sure how much sed you had to do. I don't expect `openstack overcloud roles generate` to change much, but if changes and breaks the job, maybe that's something we actually want to catch?

fair enough, I'll revert back to generate the roles.yaml. It wasn't too bad, just a couple of sed commands, but I'll put the PR back in draft to verify it works

fultonj:
Looks like the latest patch set added the command to generate the roles file, but the roles file is still part of the PR. Want to remove it?

cescgina:
> Looks like the latest patch set added the command to generate the roles file, but the roles file is still part of the PR. Want to remove it?

sure, will do in the next udpate, don't want to stop running job

cescgina:
recheck

cescgina:
Marking it ready for review since we got another green run in https://review.rdoproject.org/r/c/testproject/+/52874

fultonj (commented):
I'm +2 on this change.

@fmount what do you think?

fmount (commented):
/lgtm

fao89 (commented):
+2
just waiting the dependency to be merged to formally approve

cescgina:
> +2 just waiting the dependency to be merged to formally approve

@fao89 thanks for the review, we should actually merge this one before the rdo-jobs, I don't know why I put the depends-on here

PR #811, opened 2024-04-22
Wait for the other operators

rabi:
We changed to install openstack operator in dataplane kuttl tests, but we still have to wait for other operators to be ready before running the tests.


abays (approved):
/lgtm

PR #816, opened 2024-04-27
Don't use wildcard when applying custom services

rabi:
Using `*.yaml` does not work when there are multiple custom services in devsetup/edpm/services like we have downstream.

jira: https://issues.redhat.com/browse/OSPCIX-266


rabi:
recheck

raukadah:
/lgtm
Tested it downstream, It fixes the issue.

PR #817, opened 2024-04-30
Drop nncp ctlplane routes

karelyatin:
These ctlplane routes were originally added as part of [1], and since then they required various adjustment due to one or the other issues. It was pointed out these routes shouldn't be needed for the use case it was originally added for, so let's drop these.

[1] https://github.com/openstack-k8s-operators/install_yamls/pull/481

Related-Issue: [OSPRH-6307](https://issues.redhat.com//browse/OSPRH-6307)


karelyatin:
Both baremetal and non-baremetal job works fine

abays (approved):
/lgtm

PR #819, opened 2024-05-02
Clean chassis from OVN when deleting EDPM nodes

cubeek:
The cleanup of edpm nodes takes care only of virsh instances. The patch also adds calls to OVN database to remove the nodes from OVN because often in the dev environments the IPs for newly created compute nodes are reused which doesn't play well with OVN.

JIRA: [OSPRH-639](https://issues.redhat.com//browse/OSPRH-639)

### DIFF
--- a/devsetup/scripts/common.sh
+++ b/devsetup/scripts/common.sh
@@ -14,6 +14,13 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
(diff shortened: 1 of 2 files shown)


fao89 (approved):

slawqo (commented):
/lgtm thx for the patch

PR #823, opened 2024-05-07
Wait for the nodeset before applying the deploment

fao89:


fao89:
https://github.com/openstack-k8s-operators/ci-framework/pull/1657

PR #824, opened 2024-05-07
remove service-specific DatabasePasswords

zzzeek:
MariaDB database passwords are now generated by
the mariadb-operator API which creates MariaDBAccount and associated Secret objects.  The database passwords in the osp-secret object are no longer used.

This PR is late so may be deferred until after GA has released if there's concern about risk


fao89 (commented):
/approve

stuggi (approved):
/lgtm

PR #825, opened 2024-05-08
Set OSP17 TunedProfileName to throughput-performance matching OSP18

marios:
This sets the TunedProfileName for the tripleo overcloud_roles used with the OSP17.1 source deployment. See [1][2] for more info

[1] https://review.rdoproject.org/r/c/rdo-jobs/+/53075
[2] https://redhat-internal.slack.com/archives/CQXJFGMK6/p1715149288982149?thread_ts=1714661926.998599&cid=CQXJFGMK6

### DIFF
--- a/devsetup/tripleo/overcloud_roles.yaml
+++ b/devsetup/tripleo/overcloud_roles.yaml
@@ -206,7 +206,7 @@
   HostnameFormatDefault: '%stackname%-novacompute-%index%'
   RoleParametersDefault:
     FsAioMaxNumber: 1048576
-    TunedProfileName: "virtual-host"
+    TunedProfileName: "throughput-performance"
   # Deprecated & backward-compatible values (FIXME: Make parameters consistent)
   # Set uses_deprecated_params to true if any deprecated params are used.
   # These deprecated_params only need to be used for existing roles and not for


devsetup/tripleo/overcloud_roles.yaml
```diff
@@ -206,7 +206,7 @@
   HostnameFormatDefault: '%stackname%-novacompute-%index%'
   RoleParametersDefault:
     FsAioMaxNumber: 1048576
-    TunedProfileName: "virtual-host"
+    TunedProfileName: "throughput-performance"
```
marios, line 209: https://logserver.rdoproject.org/74/51274/19/check/periodic-data-plane-adoption-multinode-osp-17-to-extracted-crc-no-ceph/98f472c/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/pre-adoption-validation-openstack-pre-adoption-openstack-44nnj/logs/openstackansibleee.log

```
        * TASK [osp.edpm.edpm_pre_adoption_validation : Validate that current active tuned profile won't change] ***
[1;30mtask path: /usr/share/ansible/collections/ansible_collections/osp/edpm/roles/edpm_pre_adoption_validation/tasks/tuned.yml:42[0m
Wednesday 08 May 2024  10:14:27 +0000 (0:00:00.061)       0:00:10.677 ********* 
[0;32mok: [np0004646741] => {[0m
[0;32m    "msg": "Validating that the current profile 'throughput-performance' is the same as the new profile 'throughput-performance'"[0m
[0;32m}[0m
```

@gibizer @cescgina ^^

cescgina (approved):

gibizer (approved):
This is a valid way forward to make the test pass. Personally I would do the opposite and keep the 17 config as is and adapt the 18 NodeSet configuration instead to match the 17 value. I would do this as it is the more likely scenario happen on the field as on the field the 17 deployment configuration will not be changed before the adoption to pass the validation. But I accept that we don't want to test that scenario here.

fultonj:
FWIW I'm fine with merging this.

I believe RHEL's default is still throughput performance and unless a deployment's roles_data.yaml changed it to some other profile that's what we'd get by default.

PR #826, opened 2024-05-10
Add multi-cell tripleo setup for adoption devel

bogdando:
In order to keep the HW requirments for development
of multi-cell OSP 17.1 adoption for RHOSO 18, provide
a reduced multi-stack footprint
(which is supported in tripleo, yet not in OSP):

undercloud: 1 VM
overcloud: Controller0 ( 1 VM, no HA)
cell1: Compute0, Compute1 (CellController1) ( 2 VMs)
cell2: Compute2+CellController2 (AIO VM host)

No Ceph/HCI support yet, TBD.
Only a fixed number of cells (2 extra cells) supported yet.

Add partial local dev envs support (given VMs preprovisioned):
* Fix HOME paths to match the default user home from ssh opts
  (when it is 'zuul', the commands fail on the control node,
   when executed from another user name)
* Add missing repo setup commands and notes from standalone.sh
* Support RH registry auth env vars
* Disable undercloud install validations (to allow deployments
  on low storage envs)
* Fix tripleo network config to let it configure nodes in a local
  setup, where there is no pre-configured networks
* Unplug External network from roles file as CI infra does not
  provide a gateway for it, so tripleo networking fails ping test

Add EDPM_CONFIGURE_NETWORKING to control tripleo networking
configuration. Disable it for CI jobs. Should be enabled for
local deployments.

Add j2 bool filter support to jinja_render common func.

Required-By: https://review.rdoproject.org/r/c/rdo-jobs/+/53192

JIRA [OSPRH-6548](https://issues.redhat.com/browse/OSPRH-6548)

### DIFF
--- a/devsetup/scripts/common.sh
+++ b/devsetup/scripts/common.sh
@@ -33,18 +33,18 @@ function jinja2_render {
     j2_vars_file=$2
 
     /usr/bin/python3 -c "
-import yaml
 import jinja2
+import os
+import yaml
 
 with open('$j2_vars_file', 'r') as f:
     vars = yaml.safe_load(f.read())
 
-with open('$j2_template_file', 'r') as f:
-    template = f.read()
+loader = jinja2.FileSystemLoader(os.path.dirname('$j2_template_file'))
+env = jinja2.Environment(autoescape=True, loader=loader)
+env.filters['bool'] = bool
 
-j2_template = jinja2.Template(template)
-
-print(j2_template.render(**vars))
+print(env.get_template(os.path.basename('$j2_template_file')).render(**vars))
 "
 }
 
--- a/devsetup/scripts/standalone.sh
+++ b/devsetup/scripts/standalone.sh
@@ -64,12 +64,8 @@ fi
 
 source ${SCRIPTPATH}/common.sh
 
-
 # Clock synchronization is important for both Ceph and OpenStack services, so both ceph deploy and tripleo deploy commands will make use of chrony to ensure the clock is properly in sync.
-# We'll use the NTP_SERVER environmental variable to define the NTP server to use.
-# If we are running alls these commands in a system inside the Red Hat network we should use the clock.corp.redhat.com server:
-# export NTP_SERVER=clock.corp.redhat.com
-# And when running it from our own systems outside of the Red Hat network we can use any available server:
+# We'll use the NTP_SERVER environmental variable to define the NTP server to use, e.g.:
 # export NTP_SERVER=pool.ntp.org
 
 if [[ ! -f $REPO_SETUP_CMDS ]]; then
@@ -126,7 +122,7 @@ __EOF__
 
 export HOST_PRIMARY_RESOLV_CONF_ENTRY=${HOST_PRIMARY_RESOLV_CONF_ENTRY}
 export INTERFACE_MTU=${INTERFACE_MTU:-1500}
-export NTP_SERVER=${NTP_SERVER:-"clock.corp.redhat.com"}
+export NTP_SERVER=${NTP_SERVER:-"pool.ntp.org"}
 export EDPM_COMPUTE_CEPH_ENABLED=${EDPM_COMPUTE_CEPH_ENABLED:-true}
 export EDPM_COMPUTE_CEPH_NOVA=${EDPM_COMPUTE_CEPH_NOVA:-true}
 export CEPH_ARGS="${CEPH_ARGS:--e \$HOME/deployed_ceph.yaml -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm-rbd-only.yaml}"
--- a/devsetup/scripts/tripleo.sh
+++ b/devsetup/scripts/tripleo.sh
@@ -23,6 +23,7 @@ SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 IP_ADRESS_SUFFIX=${IP_ADRESS_SUFFIX:-100}
 IP=${IP:-"${EDPM_COMPUTE_NETWORK_IP%.*}.${IP_ADRESS_SUFFIX}"}
 OS_NET_CONFIG_IFACE=${OS_NET_CONFIG_IFACE:-"nic1"}
+CLOUD_DOMAIN=${CLOUD_DOMAIN:-localdomain}
 GATEWAY=${GATEWAY:-"${EDPM_COMPUTE_NETWORK_IP}"}
 OUTPUT_DIR=${OUTPUT_DIR:-"${SCRIPTPATH}/../../out/edpm/"}
 SSH_KEY_FILE=${SSH_KEY_FILE:-"${OUTPUT_DIR}/ansibleee-ssh-key-id_rsa"}
@@ -30,6 +31,7 @@ SSH_OPT="-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i $SSH_KEY
 REPO_SETUP_CMDS=${REPO_SETUP_CMDS:-"${MY_TMP_DIR}/standalone_repos"}
 CMDS_FILE=${CMDS_FILE:-"${MY_TMP_DIR}/standalone_cmds"}
 SKIP_TRIPLEO_REPOS=${SKIP_TRIPLEO_REPOS:="false"}
+TRIPLEO_NETWORKING=${TRIPLEO_NETWORKING:-true}
 MANILA_ENABLED=${MANILA_ENABLED:-true}
 OCTAVIA_ENABLED=${OCTAVIA_ENABLED:-false}
 TELEMETRY_ENABLED=${TELEMETRY_ENABLED:-true}
@@ -39,7 +41,23 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
     exit 1
 fi
 
+sudo dnf install -y python-jinja2
 source ${SCRIPTPATH}/common.sh
+
+# Clock synchronization is important for both Ceph and OpenStack services, so both ceph deploy and tripleo deploy commands will make use of chrony to ensure the clock is properly in sync.
+# We'll use the NTP_SERVER environmental variable to define the NTP server to use, e.g.:
+# export NTP_SERVER=pool.ntp.org
+
+if [ $EDPM_COMPUTE_CELLS -eq 2 ] || [ $EDPM_COMPUTE_CELLS -gt 3 ] || [ $EDPM_COMPUTE_CELLS -eq 0 ] ; then
+    echo "Only a main cell0 plus a 2 additional compute cells supported yet!"
+    exit 1
+fi
+
+# TODO(bogdando): multi-cell with ceph HCI
+if [ $EDPM_COMPUTE_CELLS -eq 3 ] && [[ "$EDPM_COMPUTE_CEPH_ENABLED" == "true" ]]; then
+    echo "Ceph storage is not supported for multi-cell setup yet!"
+    exit 1
+fi
 
 if [[ -e /run/systemd/resolve/resolv.conf ]]; then
     HOST_PRIMARY_RESOLV_CONF_ENTRY=$(cat /run/systemd/resolve/resolv.conf | grep ^nameserver | grep -v "${EDPM_COMPUTE_NETWORK_IP%.*}" | head -n1 | cut -d' ' -f2)
@@ -47,34 +65,63 @@ else
     HOST_PRIMARY_RESOLV_CONF_ENTRY=${HOST_PRIMARY_RESOLV_CONF_ENTRY:-$GATEWAY}
 fi
 
-if [[ ! -f $CMDS_FILE ]]; then
-    cat <<EOF > $CMDS_FILE
+set +x
+cat <<EOF > $MY_TMP_DIR/.standalone_env_file
+export RH_REGISTRY_USER="$REGISTRY_USER"
+export RH_REGISTRY_PWD="$RH_REGISTRY_PWD"
+EOF
+chmod 0600 $MY_TMP_DIR/.standalone_env_file
+
+cat <<EOF > $CMDS_FILE
+. \$HOME/.standalone_env_file
+
 set -ex
 sudo dnf install -y podman python3-tripleoclient util-linux lvm2
 
 sudo hostnamectl set-hostname undercloud.localdomain
 sudo hostnamectl set-hostname undercloud.localdomain --transient
 
+cat >\$HOME/nova_noceph.yaml <<__EOF__
+parameter_defaults:
+    NovaEnableRbdBackend: false
+__EOF__
+
 export HOST_PRIMARY_RESOLV_CONF_ENTRY=${HOST_PRIMARY_RESOLV_CONF_ENTRY}
 export INTERFACE_MTU=${INTERFACE_MTU:-1500}
-export NTP_SERVER=${NTP_SERVER:-"clock.corp.redhat.com"}
+export NTP_SERVER=${NTP_SERVER:-"pool.ntp.org"}
 export IP=${IP}
 export GATEWAY=${GATEWAY}
-export EDPM_COMPUTE_CEPH_ENABLED=${EDPM_COMPUTE_CEPH_ENABLED:-false}
-export CEPH_ARGS="${CEPH_ARGS:--e \$HOME/deployed_ceph.yaml -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm.yaml}"
+export EDPM_COMPUTE_CEPH_ENABLED=${COMPUTE_CEPH_ENABLED:-false}
+export EDPM_COMPUTE_CEPH_NOVA=${COMPUTE_CEPH_NOVA:-false}
+export CEPH_ARGS="${CEPH_ARGS:--e \$HOME/deployed_ceph.yaml -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm-rbd-only.yaml}"
+[[ "\$EDPM_COMPUTE_CEPH_NOVA" == "false" ]] && export CEPH_ARGS="\${CEPH_ARGS} -e \$HOME/nova_noceph.yaml"
+export EDPM_COMPUTE_CELLS=${COMPUTE_CELLS:-1}
 export MANILA_ENABLED=${MANILA_ENABLED:-true}
 export OCTAVIA_ENABLED=${OCTAVIA_ENABLED}
 export TELEMETRY_ENABLED=${TELEMETRY_ENABLED:-true}
 
+set +x
 if [[ -f \$HOME/containers-prepare-parameters.yaml ]]; then
     echo "Using existing containers-prepare-parameters.yaml - contents:"
     cat \$HOME/containers-prepare-parameters.yaml
 else
+    login_args=" "
+    [ "\$RH_REGISTRY_USER" ] && [ -n "\$RH_REGISTRY_PWD" ] && login_args="--enable-registry-login"
     openstack tripleo container image prepare default \
-        --output-env-file \$HOME/containers-prepare-parameters.yaml
+        --output-env-file \$HOME/containers-prepare-parameters.yaml \${login_args}
     # Use wallaby el9 container images
     sed -i 's|quay.io/tripleowallaby$|quay.io/tripleowallabycentos9|' \$HOME/containers-prepare-parameters.yaml
 fi
+
+if [ "\$RH_REGISTRY_USER" ] && [ -n "\$RH_REGISTRY_PWD" ]; then
+    grep -q ContainerImageRegistryCredentials \$HOME/containers-prepare-parameters.yaml || \
+    cat >> \$HOME/containers-prepare-parameters.yaml <<__EOF__
+ContainerImageRegistryCredentials:
+    registry.redhat.io:
+        \${RH_REGISTRY_USER}: \$RH_REGISTRY_PWD
+__EOF__
+fi
+set -x
 
 # Use os-net-config to add VLAN interfaces which connect edpm-compute-0 to the isolated networks configured by install_yamls.
 sudo mkdir -p /etc/os-net-config
@@ -84,6 +131,7 @@ network:
     config: disabled
 __EOF__
 
+cell=\$(( IP_ADRESS_SUFFIX - 100 ))
 sudo systemctl enable network
 sudo cp /tmp/net_config.yaml /etc/os-net-config/config.yaml
 sudo os-net-config -c /etc/os-net-config/config.yaml
@@ -97,7 +145,6 @@ popd
 # is not true, the script will return 0 and cause an error in CI
 exit 0
 EOF
-fi
 
 while [[ $(ssh -o BatchMode=yes -o ConnectTimeout=5 $SSH_OPT root@$IP echo ok) != "ok" ]]; do
     sleep 5
@@ -105,6 +152,8 @@ while [[ $(ssh -o BatchMode=yes -o ConnectTimeout=5 $SSH_OPT root@$IP echo ok) !
 done
 
 # Render Jinja2 files
+# FIXME: .99 undercloud IP cannot be changed https://review.rdoproject.org/cgit/rdo-jobs/tree/playbooks/data_plane_adoption/setup_tripleo_os_net_config.yaml#n6
+# TODO(bogdnado): introduce CLOUD_USER var to remove all hardcoded 'zuul' values from these scripts
 PRIMARY_RESOLV_CONF_ENTRY=${HOST_PRIMARY_RESOLV_CONF_ENTRY}
 J2_VARS_FILE=$(mktemp --suffix=".yaml" --tmpdir="${MY_TMP_DIR}")
 cat << EOF > ${J2_VARS_FILE}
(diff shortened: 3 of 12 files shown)


devsetup/tripleo/overcloud_services _cell.j2
```diff
@@ -0,0 +1,53 @@
+---
```
marios, line 1: whitespace in the file name

marios (commented):
please lets discuss this during tomorrow data plane call before proceeding down this path

i want us all to agree on the direction.

i think you are proposing that we will have a 3 compute hci ceph job, and then a 1 cntrol/2 compute for cells (i.e. this job) but we have not had that conversation yet. 

we need to be on the same page about the ci direction since we (ci team) are currently working here

cc @cescgina @frenzyfriday @jistr

devsetup/tripleo/config-download-multistack.yaml
```diff
@@ -0,0 +1,149 @@
...
+      dns_domain: localdomain
+      mtu: 1500
+      name: ctlplane
+      tags:
+        - 192.168.122.0/24
+    subnets:
+      ctlplane-subnet:
+        cidr: 192.168.122.0/24
+        dns_nameservers: 192.168.122.1
+        gateway_ip: 192.168.122.1
```
bogdando, line 146: those need to be configurable, as in CI we use ``standalone_gateway: 192.168.122.10``

bogdando:
got the central overcloud stack deployed locally with this, some adjustments will be needed to have the gateway_ip paramterized

devsetup/tripleo/config-download-multistack.yaml
```diff
@@ -0,0 +1,164 @@
...
+      name: ctlplane
+      tags:
+        - 192.168.122.0/24
+    subnets:
+      ctlplane-subnet:
+        cidr: 192.168.122.0/24
+        # FIXME(bogdando): those need to be the hypervisor virbr0 IP as a DNS resolver,
+        # and the undercloud ctlplane IP as a gateway. But multi-node job in CI uses .10 for all...
+        dns_nameservers: [192.168.122.10, 192.168.122.1]
+        gateway_ip: 192.168.122.10
```
bogdando, line 145: TODO: have this one j2 templated as well

devsetup/scripts/tripleo.sh
```diff
@@ -35,16 +36,45 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
...
+
 if [[ -e /run/systemd/resolve/resolv.conf ]]; then
     HOST_PRIMARY_RESOLV_CONF_ENTRY=$(cat /run/systemd/resolve/resolv.conf | grep ^nameserver | grep -v "${EDPM_COMPUTE_NETWORK_IP%.*}" | head -n1 | cut -d' ' -f2)
 else
     HOST_PRIMARY_RESOLV_CONF_ENTRY=${HOST_PRIMARY_RESOLV_CONF_ENTRY:-$GATEWAY}
 fi
 
+cat <<EOF > $MY_TMP_DIR/.standalone_env_file
+export RH_REGISTRY_USER="$REGISTRY_USER"
+export RH_REGISTRY_PWD="$RH_REGISTRY_PWD"
```
marios, line 70: if we run this in ci we may leak the user and pass for the registry...

devsetup/tripleo/net_config.j2
```diff
@@ -26,37 +26,37 @@ network_config:
     mtu: {{ interface_mtu }}
     vlan_id: 44
     addresses:
-    - ip_netmask: 172.21.0.{{ ip_address_suffix }}/24
+    - ip_netmask: 172.21.0.{{ ip_address_suffix }}/16
```
marios, line 29: as previously discussed can you please avoid making changes here that affect the 3control/3compute multinode. This file is used for the overcloud node config in that job.

devsetup/tripleo/tripleo_install.sh
```diff
@@ -15,69 +15,158 @@
...
+    # For the time being, use the same network data (with network leafs) for all stacks we deploy
+    for cell in $(seq 0 $(( EDPM_COMPUTE_CELLS - 1))); do
+        [ $cell -eq 1 ] && continue
+        cp network_data${cell}.yaml network_data${cell}_cells_specific_unused.yaml
+        cp network_data1.yaml network_data${cell}.yaml
+    done
+    sed -i "s/overcloud/multistack/g" overcloud_services.yaml
+fi
+
+openstack overcloud network provision --output network_provision_out0.yaml ./network_data0.yaml
```
marios, line 33: why these filenames out0.yaml please restore?

devsetup/tripleo/tripleo_install.sh
```diff
@@ -15,69 +15,158 @@
...
+                    -e tripleo_cellv2_cell_name=cell${cell} \
+                    -e tripleo_cellv2_containercli=podman
+
+        echo "add cell $cell compute hosts to aggregates"
+        openstack aggregate create cell${cell} --zone cell${cell}
+        for i in $(openstack hypervisor list -f value -c 'Hypervisor Hostname'| grep cell${cell}); do
+            openstack aggregate add host cell${cell} $i
+        done
+    done
+fi
```
marios, line 172: needs newline

marios (changes requested):

devsetup/tripleo/net_config.j2
marios: (actually this is used for the undercloud node, not overcloud, but same point applies

devsetup/scripts/tripleo.sh
bogdando: this is intented for local setups, but I will add ``set +x``

devsetup/tripleo/net_config.j2
bogdando: Could we do this for multi-node as well? This change is not breaking, while allows not duplicating more things around

devsetup/tripleo/tripleo_install.sh
bogdando: this made it simpler to iterate for multi-stack, where is stack 0 is overcloud.
I will retain the name via adding an alias to the old names for a single-stack multi-node compat

devsetup/scripts/tripleo.sh
marios: even with +x i don't think this is acceptable since you are actually writing the user/pass to file here

this script is being executed by the existing multinode ci jobs i.e. not just for your new multi cell addition.

Isn't the whole point of this work that we will be adding a new multi cell job in ci? So what do you mean about local setups.

devsetup/scripts/tripleo.sh
```diff
@@ -56,24 +87,46 @@ sudo dnf install -y https://kojihub.stream.centos.org/kojifiles/packages/podman/
...
     # Use wallaby el9 container images
     sed -i 's|quay.io/tripleowallaby$|quay.io/tripleowallabycentos9|' \$HOME/containers-prepare-parameters.yaml
 fi
 
+if [ "\$RH_REGISTRY_USER" ] && [ -n "\$RH_REGISTRY_PWD" ]; then
+    grep -q ContainerImageRegistryCredentials /home/zuul/containers-prepare-parameters.yaml || \
+    cat >> /home/zuul/containers-prepare-parameters.yaml <<__EOF__
+  ContainerImageRegistryCredentials:
+    registry.redhat.io:
+      \${RH_REGISTRY_USER}: \$RH_REGISTRY_PWD
```
marios, line 125: same concern here about leaking the registry user/pass

marios (dismissed):

devsetup/tripleo/net_config.j2
marios: possibly but lets please wait until that is merged we are very close now and startign to merge the series

we can discuss next week?

devsetup/scripts/tripleo.sh
bogdando: > even with +x i don't think this is acceptable since you are actually writing the user/pass to file here
> this script is being executed by the existing multinode ci jobs i.e. not just for your new multi cell addition.
> Isn't the whole point of this work that we will be adding a new multi cell job in ci? So what do you mean about  local setups.

As a developer, I need a way to deploy it locally, using this same code.
It won't fly without registry creds. There is no reproducer for it in CI, or any docs on how to use it without auto gold nodes in zuul (not the best choice neither)

bogdando: We can make a patch to ci_framework to always unset ``RH_REGISTRY_PWD`` as adependency of this change. Then it should address your concen

bogdando: Done! Does it make sense now https://review.rdoproject.org/r/c/rdo-jobs/+/53192/34..35/playbooks/data_plane_adoption/install_yamls_deploy_tripleo.yaml ?

devsetup/tripleo/net_config.j2
bogdando: I failled back to a flat subnet, there is yet routing issues for /16 external network to address for that switch in ci framework

bogdando:
Local source trilpeo deployment has completed w/o errors for this revision.

devsetup/scripts/tripleo.sh
bogdando: PTAL @marios

marios: ack thanks @bogdando. I think https://review.rdoproject.org/r/c/rdo-jobs/+/53192/34..35/playbooks/data_plane_adoption/install_yamls_deploy_tripleo.yaml should make sure we aren't setting this in the ci runs.

It will be confusing though because someone will rightly ask "why are you setting the username and password of the registry to '' " ?

Thanks for acknowledging the concern and sanity checking that we aren't going to run this in ci.

marios (commented):
can you please ensure this is tested with the periodic multinode jobs before we merge

i.e. in your testproject run (at least one of)
```
periodic-data-plane-adoption-multinode-osp-17-to-extracted-crc-no-ceph
periodic-data-plane-adoption-multinode-osp-17-to-extracted-crc-with-ceph
```

bogdando:
> can you please ensure this is tested with the periodic multinode jobs before we merge
> 
> i.e. in your testproject run (at least one of)
> 
> ```
> periodic-data-plane-adoption-multinode-osp-17-to-extracted-crc-no-ceph
> periodic-data-plane-adoption-multinode-osp-17-to-extracted-crc-with-ceph
> ```

https://review.rdoproject.org/r/c/testproject/+/53212/8..9/.zuul.yaml

devsetup/scripts/tripleo.sh
```diff
@@ -36,16 +37,46 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
...
 
+sudo dnf install -y python-jinja2
 source ${SCRIPTPATH}/common.sh
 
+# Clock synchronization is important for both Ceph and OpenStack services, so both ceph deploy and tripleo deploy commands will make use of chrony to ensure the clock is properly in sync.
+# We'll use the NTP_SERVER environmental variable to define the NTP server to use.
+# If we are running alls these commands in a system inside the Red Hat network we should use the clock.corp.redhat.com server:
+# export NTP_SERVER=clock.corp.redhat.com
+# And when running it from our own systems outside of the Red Hat network we can use any available server:
+# export NTP_SERVER=pool.ntp.org
```
marios, line 48: do we really need this comment, there is an existing variable and indeed what you are documenting here is how it is used

marios (commented):
just another pass but its a huge diff 
main concern is the existing multinode jobs but i believe you are already running those in your test runs.

devsetup/scripts/tripleo.sh
bogdando: I have copied this note from standalone tripleo shell script, to align them better. It makes sense to clarify that important moment here as well

bogdando:
> just another pass but its a huge diff main concern is the existing multinode jobs but i believe you are already running those in your test runs.

there is no huge diff if you follow the "Compare" button updates to show only changes for a revision. Only reabasing brings in irrelevant changes.

bogdando:
This is already deployable locally for a few weeks as of now, yet I am still struggling to make networking configs working in testproject. This is still WIP, just for your info.

bogdando:
Without tripleo-networking applied, I am still having a pingtest problem with external network GW in testproject.
Adding an option to omit ping tests completely (for the srouce tripleo cloud), when going with a BYI network on deployed servers

devsetup/tripleo/net_config.j2
```diff
@@ -14,7 +14,10 @@ network_config:
   - ip_netmask: {{ ctlplane_ip }}/{{ ctlplane_cidr }}
   - ip_netmask: {{ ctlplane_ip }}/32
   - ip_netmask: {{ ctlplane_vip }}/32
-  routes: []
+  routes:
```
cescgina, line 17: @bogdando this bit is what is causing the undercloud node to become unreachable

cescgina: this causes a restart of the `eth0` interface, that has the public ip that zuul assigns to the node

bogdando:
New test project for multinode jobs (not deploying multicell yet) https://review.rdoproject.org/r/c/testproject/+/54199

devsetup/scripts/tripleo.sh
```diff
@@ -39,43 +41,94 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
     exit 1
 fi
 
+sudo dnf install -y python-jinja2
 source ${SCRIPTPATH}/common.sh
 
+# Clock synchronization is important for both Ceph and OpenStack services, so both ceph deploy and tripleo deploy commands will make use of chrony to ensure the clock is properly in sync.
+# We'll use the NTP_SERVER environmental variable to define the NTP server to use.
+# If we are running alls these commands in a system inside the Red Hat network we should use the clock.corp.redhat.com server:
+# export NTP_SERVER=clock.corp.redhat.com
```
cescgina, line 50: I understand this is already published elsewhere, but I don't think we should be posting internal urls in public repos

devsetup/scripts/tripleo.sh
```diff
@@ -39,43 +41,94 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
...
+# And when running it from our own systems outside of the Red Hat network we can use any available server:
+# export NTP_SERVER=pool.ntp.org
+
+if [ $EDPM_COMPUTE_CELLS -eq 2 ] || [ $EDPM_COMPUTE_CELLS -gt 3 ] || [ $EDPM_COMPUTE_CELLS -eq 0 ] ; then
```
cescgina, line 54: just want to make sure I'm parsing the if correctly, this is trying to check that the number of compute cells is 1 or 3, right?

devsetup/scripts/tripleo.sh
```diff
@@ -39,43 +41,94 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
...
+    echo "Only a main cell0 plus a 2 additional compute cells supported yet!"
+    exit 1
+fi
+
+# TODO(bogdando): multi-cell with ceph HCI
+if [ $EDPM_COMPUTE_CELLS -eq 3 ] && [[ "$EDPM_COMPUTE_CEPH_ENABLED" =~ "rue" ]]; then
```
cescgina, line 60: wouldn't it be better to stop the execution at this point if this configuration is unsupported atm?

devsetup/scripts/tripleo.sh
```diff
@@ -39,43 +41,94 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
...
+__EOF__
+
 export HOST_PRIMARY_RESOLV_CONF_ENTRY=${HOST_PRIMARY_RESOLV_CONF_ENTRY}
 export INTERFACE_MTU=${INTERFACE_MTU:-1500}
 export NTP_SERVER=${NTP_SERVER:-"clock.corp.redhat.com"}
 export IP=${IP}
 export GATEWAY=${GATEWAY}
-export EDPM_COMPUTE_CEPH_ENABLED=${EDPM_COMPUTE_CEPH_ENABLED:-false}
-export CEPH_ARGS="${CEPH_ARGS:--e \$HOME/deployed_ceph.yaml -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm.yaml}"
+export EDPM_COMPUTE_CEPH_ENABLED=${COMPUTE_CEPH_ENABLED:-False}
```
cescgina, line 100: `False` should be lowercase, as that is what we're checking later on https://github.com/openstack-k8s-operators/install_yamls/pull/826/files#diff-6d137e1f40b93f9957103127166db87dbd340c674dcf1d8694def950ebd56821R101

devsetup/scripts/tripleo.sh
```diff
@@ -39,43 +41,94 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
...
+export EDPM_COMPUTE_CEPH_NOVA=${COMPUTE_CEPH_NOVA:-False}
```
cescgina, line 101: same as above

devsetup/scripts/tripleo.sh
```diff
@@ -39,43 +41,94 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
...
     openstack tripleo container image prepare default \
-        --output-env-file \$HOME/containers-prepare-parameters.yaml
+        --output-env-file \$HOME/containers-prepare-parameters.yaml \${login_args}
     # Use wallaby el9 container images
     sed -i 's|quay.io/tripleowallaby$|quay.io/tripleowallabycentos9|' \$HOME/containers-prepare-parameters.yaml
 fi
 
+if [ "\$RH_REGISTRY_USER" ] && [ -n "\$RH_REGISTRY_PWD" ]; then
+    grep -q ContainerImageRegistryCredentials /home/zuul/containers-prepare-parameters.yaml || \
+    cat >> /home/zuul/containers-prepare-parameters.yaml <<__EOF__
```
cescgina, line 124: should use `$HOME` instead of `/home/zuul`, as this block is mostly intended for dev use cases, where user will likely be different

devsetup/scripts/tripleo.sh
```diff
@@ -117,39 +172,92 @@ ip_address_suffix: ${IP_ADRESS_SUFFIX}
 interface_mtu: ${INTERFACE_MTU:-1500}
 ntp_server: ${NTP_SERVER}
 gateway_ip: ${GATEWAY}
+manage_default_route: ${TRIPLEO_NETWORKING}
 dns_server: ${PRIMARY_RESOLV_CONF_ENTRY}
-user_home: ${HOME}
+user_home: /home/zuul
```
cescgina, line 177: we should not hardcode the user here

devsetup/tripleo/net_config.j2
```diff
@@ -14,7 +14,14 @@ network_config:
   - ip_netmask: {{ ctlplane_ip }}/{{ ctlplane_cidr }}
   - ip_netmask: {{ ctlplane_ip }}/32
   - ip_netmask: {{ ctlplane_vip }}/32
+  {% if manage_default_route|bool %}
```
cescgina, line 17: we should include a default false here in case this is not defined (we use this template to configure networking in ci without calling the tripleo.sh script)

bogdando:
Needs https://review.rdoproject.org/r/c/rdo-jobs/+/54221

devsetup/scripts/tripleo.sh
bogdando: correct

devsetup/scripts/tripleo.sh
bogdando: perhaps, however there are lots of assumptions have already been made for using zuul user in this and other scripts

devsetup/scripts/tripleo.sh
bogdando: the issue on my setup was that on hypervisor I had no zuul user, but in VMs, it needed to be zuul. So that default HOME env var was incorrectly interpolated by j2 render.

To properly fix this we'd need to introduce a new env var for these scripts and Make file, somthing like CLOUD_USER, and get rid of all assumptions made here for zuul user, which goes a bit beyond of this patch scope

jistr:
I think we should probably discuss this and https://github.com/openstack-k8s-operators/data-plane-adoption/pull/517 on one of the adoption calls or better yet, schedule a separate call to do a high level walk through and what should be the plan for going forward with multi-cell?

devsetup/tripleo/net_config.j2
marios: i have a similar concern about the default_route
I arrived here following the TRIPLEO_NETWORKING
The intention is that this will only affect the multicell scenario/job but the default given is TRUE 

```
EDPM_CONFIGURE_NETWORKING ?= true
```

We could either make the default be false, and as @cescgina pointed out default it here, or the inverse, which is what you proposed with https://review.rdoproject.org/r/c/rdo-jobs/+/54221 where you explicitly set it false for the vanila multinode case

```
            EDPM_CONFIGURE_NETWORKING: false  

```

In my opinion it is more appropriate for this to default to false and then you can explicitly set this in the multi-cell job where is will be needed.

In any case though and before we merge we must run the multinode jobs so we can sanity check they are ok with these changes. I know you've run them in the past but we should run them again once this is ready for merge  as there are many ongoing changes here.

devsetup/scripts/standalone.sh
```diff
@@ -126,7 +122,7 @@ __EOF__
 
 export HOST_PRIMARY_RESOLV_CONF_ENTRY=${HOST_PRIMARY_RESOLV_CONF_ENTRY}
 export INTERFACE_MTU=${INTERFACE_MTU:-1500}
-export NTP_SERVER=${NTP_SERVER:-"clock.corp.redhat.com"}
+export NTP_SERVER=${NTP_SERVER:-"pool.ntp.org"}
```
marios, line 125: fair enough but this change has enough going on without also dealing with this default switch here

devsetup/scripts/tripleo.sh
```diff
@@ -39,43 +41,88 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
...
 fi
 
+sudo dnf install -y python-jinja2
 source ${SCRIPTPATH}/common.sh
 
+# Clock synchronization is important for both Ceph and OpenStack services, so both ceph deploy and tripleo deploy commands will make use of chrony to ensure the clock is properly in sync.
+# We'll use the NTP_SERVER environmental variable to define the NTP server to use, e.g.:
+# export NTP_SERVER=pool.ntp.org
+
+if [ $EDPM_COMPUTE_CELLS -eq 2 ] || [ $EDPM_COMPUTE_CELLS -gt 3 ] || [ $EDPM_COMPUTE_CELLS -eq 0 ] ; then
```
marios, line 51: can we instead explicitly search for what IS allowed ? i.e. the "main cell0 plus additional 2 compute cells"

marios (commented):
@bogdando I think it may help to coordinate everyone with a call to discuss this and https://review.rdoproject.org/r/c/rdo-jobs/+/53192 and https://github.com/openstack-k8s-operators/ci-framework/pull/1984 and the others in the series

the ci reviews call may be one place, or if you prefer to setup another call?
maybe we can sync about that in today's data plane adoption general/weekly.

Before we start merging and after any further comments are addressed/changes made here or in the rest of the patches we need to make sure we run the multinode jobs (otherwise we'll just be chasing that to fix things)

in upstream we need to run at least one of the jobs (ceph/no ceph)


[periodic-adoption-standalone-to-crc-no-ceph](https://review.rdoproject.org/zuul/job/periodic-adoption-standalone-to-crc-no-ceph)
[periodic-adoption-standalone-to-crc-ceph](https://review.rdoproject.org/zuul/job/periodic-adoption-standalone-to-crc-ceph)

downstream please run the networker job 
[periodic-internal-adoption-multinode-networker-to-crc](https://sf.hosted.upshift.rdu2.redhat.com/zuul/t/tripleo-ci-internal/job/periodic-internal-adoption-multinode-networker-to-crc)

bogdando:
the [failure](https://logserver.rdoproject.org/99/54199/4/check/periodic-adoption-multinode-to-crc-no-ceph/b2673be/controller/ci-framework-data/logs/ci_script_010_run_tripleo.log) in testproject for existing multinode jobs looks real
``" Cannot find any hosts on "overcloud" in network "ctlplane": tripleoclient.exceptions.DeploymentError: Cannot find any hosts on "overcloud" in network "ctlplane"``

devsetup/scripts/standalone.sh
bogdando: could you elaborate? in rod jobs we provide an overriding interface https://github.com/29418/rdo-jobs/blob/6c8849c108a11ff1167ad9e726aeb5f039b9edfb/playbooks/data_plane_adoption/ci_framework_deploy_standalone_vm.yaml#L12 so I don't get the question

bogdando:
> @bogdando I think it may help to coordinate everyone with a call to discuss this and https://review.rdoproject.org/r/c/rdo-jobs/+/53192 and [openstack-k8s-operators/ci-framework#1984](https://github.com/openstack-k8s-operators/ci-framework/pull/1984) and the others in the series

for simplicity, I would defer discussing next steps, which is https://review.rdoproject.org/r/c/rdo-jobs/+/53192 and https://github.com/openstack-k8s-operators/ci-framework/pull/1984, and https://github.com/openstack-k8s-operators/data-plane-adoption/pull/517 . This patch scope if a fasts and quick way to stand up a development setup of multi-stack tripleo. This is needed for futher work until we get to a workable ci-framework "osp-plugin", with multi-stack, solution

devsetup/scripts/tripleo.sh
bogdando: do you mean state in the echo message? what are we going to "search" here?

bogdando:
> Before we start merging and after any further comments are addressed/changes made here or in the rest of the patches we need to make sure we run the multinode jobs

it should be good as well to make sure it passes test project https://review.rdoproject.org/r/c/testproject/+/54199

devsetup/scripts/standalone.sh
marios: you are changing the default. The change itself is probably right since we generally avoid exposing the interal urls in this space. It does mean though that all jobs previously overriding the default will now have to switch it. I am questioning if that is a necessary change as part of the series 
i.e. it is 'right' but is it needed here, on top of all the other changes being made.

devsetup/scripts/tripleo.sh
```diff
@@ -39,43 +41,88 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
...
+cat <<EOF > $CMDS_FILE
+. \$HOME/.standalone_env_file
+
 set -ex
 sudo dnf install -y podman python3-tripleoclient util-linux lvm2
 
 sudo hostnamectl set-hostname undercloud.localdomain
 sudo hostnamectl set-hostname undercloud.localdomain --transient
 
+cat >\$HOME/nova_noceph.yaml <<__EOF__
```
marios, line 84: why don't we just add this file into the rest of the env files instead of creating like that
i mean, what is special about this one?

devsetup/scripts/tripleo.sh
```diff
@@ -39,43 +41,88 @@ if [[ ! -f $SSH_KEY_FILE ]]; then
...
 export OCTAVIA_ENABLED=${OCTAVIA_ENABLED}
 export TELEMETRY_ENABLED=${TELEMETRY_ENABLED:-true}
 
+set +x
 if [[ -f \$HOME/containers-prepare-parameters.yaml ]]; then
     echo "Using existing containers-prepare-parameters.yaml - contents:"
     cat \$HOME/containers-prepare-parameters.yaml
 else
+    login_args=" "
+    [ "\$RH_REGISTRY_USER" ] && [ -n "\$RH_REGISTRY_PWD" ] && login_args="--enable-registry-login"
```
marios, line 109: what about the case where we have the user/pwd set to ""  i.e. like in https://review.rdoproject.org/r/c/rdo-jobs/+/54221/5/playbooks/data_plane_adoption/install_yamls_deploy_tripleo.yaml
will it still work if you are using --enable-registry-login?

devsetup/scripts/tripleo.sh
bogdando: it is a recommended approach in general to provide separate files for different configuration areas, rather than mangling large files.

devsetup/scripts/tripleo.sh
bogdando: I guess https://review.rdoproject.org/r/c/testproject/+/54199 will show this out

devsetup/scripts/tripleo.sh
marios: like if EDPM_COMPUTE_CELLS != 3 (or whatever the allowed is i think 3?) then error

marios (commented):
waiting to hear that testing is good and will revisit

devsetup/scripts/tripleo.sh
bogdando: This already explicitely checks for allowed values of a 1 and 3 :)

bogdando:
the regression test is passed the multi-node check https://review.rdoproject.org/r/c/testproject/+/54199 ! Now please help testing it downstream, for networker

bogdando:
recheck

devsetup/scripts/standalone.sh
marios: i just sanity checked that we are already setting this to d/stream servers for our dstream jobs so we should be OK on the default switch.
marking resolved.

devsetup/scripts/tripleo.sh
marios: i meant more like, we are already carrying a bunch of environment files. Fine if you don't want to put this into an existing one but you can just add a new file under the tripleo directory. 
With this we have some environments/parameter_defaults coming from files, and others like this one being created by heredoc. I think it just further adds to the complexity of this 'tool'.

devsetup/scripts/tripleo.sh
marios: result looks OK 

https://logserver.rdoproject.org/99/54199/4/check/periodic-adoption-multinode-to-crc-no-ceph/8770982/undercloud/zuul/containers-prepare-parameters.yaml

i guess the next bit for adding creds into the file is not executed since we set user/pass to ""

marios (approved):
networker log looks OK with tripleo deploy complete https://sf.hosted.upshift.rdu2.redhat.com/logs/78/678/02dce54df035fc84769dd16020946b926ebd7a51/check-gitlab-cee/periodic-internal-adoption-multinode-networker-to-crc/26a0f31/controller/ci-framework-data/logs/ci_script_010_run_tripleo.log

Overcloud Deployed successfully

bogdando:
can we merge this now please?

bogdando:
@jistr PTAL

devsetup/scripts/common.sh
```diff
@@ -33,18 +33,18 @@ function jinja2_render {
...
 with open('$j2_vars_file', 'r') as f:
     vars = yaml.safe_load(f.read())
 
-with open('$j2_template_file', 'r') as f:
-    template = f.read()
-
-j2_template = jinja2.Template(template)
+loader = jinja2.FileSystemLoader(os.path.dirname('$j2_template_file'))
+env = jinja2.Environment(autoescape=True, loader=loader)
+env.filters['bool'] = bool
```
olliewalsh, line 45: I would avoid using different logic to the ansible bool filter. It will return True for "yes", False for "no", while `bool("no")` will return True. https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/filter/core.py#L89

devsetup/scripts/tripleo.sh
olliewalsh: `if [ $EDPM_COMPUTE_CELLS -ne 1 -a $EDPM_COMPUTE_CELLS -ne 3 ]; then ...`

devsetup/scripts/tripleo.sh
```diff
@@ -84,6 +131,7 @@ network:
     config: disabled
 __EOF__
 
+cell=\$(( IP_ADRESS_SUFFIX - 100 ))
```
olliewalsh, line 134: where is this used?

devsetup/scripts/common.sh
olliewalsh: Also expect filtering through bool will be a no-op. `if <expr>` and `if bool(<expr>)` are equivalent AFAIK

fao89:
/hold
just noticed @olliewalsh comments

bogdando:
@olliewalsh I will address your comments in follow up, sorry.

devsetup/scripts/common.sh
bogdando: This is not a perfect solution, I agree. However, I tested it locally and in CI for multinode jobs, and it works. I will try to improve this in follow ups

devsetup/tripleo/tripleo_install.sh
```diff
@@ -15,11 +15,32 @@
 # under the License.
 set -ex
```
marios, line 17: @bogdando please check slack when you get some time

this breaks the ceph job (your test ran multinode-no-ceph) so please have a look and if we cannot find a good solution today then we can revert and work it out before going again

PR #828, opened 2024-05-13
Allows to filter what we install from download_tools

cjeanner:
This patch introduces tags in the role, so that we can target precisely
what we want to install.

It also remove mentions of `become_user: root` - that's the default
value, no need to mention it.

### DIFF
--- a/devsetup/roles/download_tools/tasks/main.yaml
+++ b/devsetup/roles/download_tools/tasks/main.yaml
@@ -1,7 +1,8 @@
 ---
 - name: Install build dependencies
   become: true
-  become_user: root
+  tags:
+    - dependencies
   ansible.builtin.package:
     name:
       - jq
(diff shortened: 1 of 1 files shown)


devsetup/roles/download_tools/tasks/main.yaml
```diff
@@ -97,14 +122,17 @@
...
   ansible.builtin.file:
     src: "{{ lookup('env', 'HOME') }}/bin/yq_linux_amd64"
     dest: "{{ lookup('env', 'HOME') }}/bin/yq"
     state: link
 
 - name: Set proper golang on the system
   become: true
-  become_user: root
+  tags:
+    - goland
```
abays, line 135: `s/goland/golang`

cjeanner: aren't we in the Land of Go? damned GPS :]

PR #830, opened 2024-05-14
edpm: Delete compute service if it exists

cubeek:
When creating the new edpm nodes the patch checks if one with the same name already exists. If it does, it is replaced by the new one.

Related-JIRA: [OSPRH-639](https://issues.redhat.com//browse/OSPRH-639)

### DIFF
--- a/devsetup/scripts/edpm-node-cleanup.sh
+++ b/devsetup/scripts/edpm-node-cleanup.sh
@@ -40,6 +40,12 @@ if [ "x" != "x$chassis_uuid" ]; then
     run_ovn_ctl_command SB chassis-del $chassis_uuid
 fi
 
+# We don't know the domain of the FQDN so we need to search for the name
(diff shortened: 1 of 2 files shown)


bogdando (commented):
lgtm

devsetup/scripts/edpm-node-cleanup.sh
```diff
@@ -40,6 +40,12 @@ if [ "x" != "x$chassis_uuid" ]; then
     run_ovn_ctl_command SB chassis-del $chassis_uuid
 fi
 
+# We don't know the domain of the FQDN so we need to search for the name
+compute_service_uuid=$(run_openstack_command compute service list -c ID -c Host --service nova-compute -f value | awk "/${EDPM_COMPUTE_NAME}/{ print \$1 }")
+if [ "x" != "x$compute_service_uuid" ]; then
+    run_openstack_command compute service delete $compute_service_uuid
```
gibizer, line 46: Can we assume at this point that the nova-compute service is stopped on these computes? Otherwise the service will recreate itself in the DB.

Also this service delete will fail if there is already VMs exists on the compute.

gibizer (commented):
I know I'm late but I have a question inline.

PR #832, opened 2024-05-15
Add workaround for ceph deployment in multinode tripleo

cescgina:
The ceph deployment for multinode tripleo job seems to fail when
creating an image, apparently because we currently only have one
compute. To have a functional cluster with 1 compute, we set an initial
ceph config with osd_pool_default_size=1.

This change also fixes the deployment so the ComputeHCI is actually
defined. To do that we needed to change the compute node hostname to
computehci and define the storage management network for the compute
node. We plan to have 3 computes soon, but
meanwhile we add this workaround to run the rest of the adoption tests.

### DIFF
--- a/devsetup/tripleo/ceph.sh
+++ b/devsetup/tripleo/ceph.sh
@@ -45,8 +45,8 @@ openstack overcloud roles generate Controller ComputeHCI > roles.yaml
 
 # disable external gateway for controller nodes
 sed -i "s/default_route_networks: \['External'\]/default_route_networks: \['ControlPlane'\]/" roles.yaml
-sed -i "/External:/d" -i roles.yaml
-sed -i "/subnet: external_subnet/d" -i roles.yaml
+sed -i "/External:/d" roles.yaml
+sed -i "/subnet: external_subnet/d" roles.yaml
 
 # generate ceph_spec file
 openstack overcloud ceph spec config-download.yaml \
@@ -61,4 +61,5 @@ openstack overcloud ceph deploy \
     --ceph-spec ceph_spec.yaml \
     --network-data network_data.yaml \
     --cephadm-default-container \
-    --output deployed_ceph.yaml
+    --output deployed_ceph.yaml \
+    --config initial-ceph.conf # TODO: jgilaber only required have ceph working with only 1 compute node should remove it once we move to have 3 computes
--- a/devsetup/tripleo/initial-ceph.conf
+++ b/devsetup/tripleo/initial-ceph.conf
@@ -1,0 +1,2 @@
+[global]
+osd_pool_default_size=1
--- a/devsetup/scripts/tripleo.sh
+++ b/devsetup/scripts/tripleo.sh
@@ -138,6 +138,7 @@ scp $SSH_OPT tripleo/ansible_config.cfg zuul@$IP:$HOME/ansible_config.cfg
 if [[ "$EDPM_COMPUTE_CEPH_ENABLED" == "true" ]]; then
     scp $SSH_OPT tripleo/ceph.sh root@$IP:/tmp/ceph.sh
     scp $SSH_OPT tripleo/generate_ceph_inventory.py root@$IP:/tmp/generate_ceph_inventory.py
+    scp $SSH_OPT tripleo/initial-ceph.conf root@$IP:$HOME/initial-ceph.conf
 fi
 
 if [[ -f $HOME/containers-prepare-parameters.yaml ]]; then
--- a/devsetup/tripleo/config-download.yaml
+++ b/devsetup/tripleo/config-download.yaml
@@ -122,6 +122,10 @@ parameter_defaults:
         ip_address: 172.18.0.106
         ip_address_uri: 172.18.0.106
         ip_subnet: 172.18.0.0/24
+      storage_mgmt:
+        ip_address: 172.20.0.106
+        ip_address_uri: 172.20.0.106
+        ip_subnet: 172.20.0.0/24
       tenant:
         ip_address: 172.19.0.106
         ip_address_uri: 172.19.0.106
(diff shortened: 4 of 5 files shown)


devsetup/tripleo/ceph.sh
```diff
@@ -62,3 +62,12 @@ openstack overcloud ceph deploy \
     --network-data network_data.yaml \
     --cephadm-default-container \
     --output deployed_ceph.yaml
+
+# TODO: jgilaber workaround to have ceph working with only 1 compute node
+# should remove it once we move to have 3 computes
+ssh -o StrictHostKeyChecking=no root@192.168.122.103 cephadm shell -- ceph osd pool create images --size 1
```
fultonj, line 68: -1 

Please remove lines 65-73 and then update line 60 so it passes `--single-host-defaults`. It should look like this:

openstack overcloud ceph deploy \
  --single-host-defaults \
  --tld localdomain \

This will configure ceph ceph as described below since you're using a single-node cluster:

  https://docs.ceph.com/en/latest/cephadm/install/#single-host

As per the above `osd_pool_default_size = 2` will be set and every pool created (when `openstack overcloud deploy` runs will inherit that pool size).

You can tweak the initial ceph configuration further by passing an initial ceph conf as described here:

https://docs.openstack.org/project-deploy-guide/tripleo-docs/latest/features/deployed_ceph.html#ceph-configuration-options

In general, tripleo be passed the desired ceph configuration so that you don't need to SSH in and run these commands.

fultonj (commented):
There's an easier and cleaner way to do this as recommended below.

devsetup/tripleo/ceph.sh
fmount, line 68: -2: This is a workaround that shouldn't be tied to `install_yamls`.
I'll let @fultonj comment but as per [1] you can pass a `ceph.conf` that will be assimilated during the bootstrap of the cluster, and it will set the option for you.
As described in [2], you can put in an `initial-ceph.conf` something like:

```
[global]
osd_pool_default_size = 1
```
and, as described for the network [3], you can pass it to the `ceph deploy` command (`--config ceph.conf`):

```
openstack overcloud ceph deploy \
  --config initial-ceph.conf ....
```

Let's try this approach that seems more generic, useful for the long term, and it's not a workaround.
@fultonj can you check my comment and see if is accurate?

@katarimanojk FYI

[1] https://github.com/openstack-archive/tripleo-ansible/blob/stable/wallaby/tripleo_ansible/roles/tripleo_cephadm/tasks/bootstrap.yaml#L37C15-L37C46
[2] https://docs.ceph.com/en/latest/install/manual-deployment/
[3] https://docs.openstack.org/project-deploy-guide/tripleo-docs/wallaby/features/deployed_ceph.html#example-directly-setting-network-and-ms-bind-options

fmount (changes requested):

devsetup/tripleo/ceph.sh
cescgina: thanks @fultonj  for the link, that is much cleaner. I'll try first with `--single-host-defaults` and if that's not enough I'll add a ceph.conf

cescgina: indeed, trying now with @fultonj suggestion above

fultonj: I'm hoping `--single-host-defaults` is sufficient but if it's not we can use `--config initial-ceph.conf`

cescgina: it seems to be enough @fultonj thanks, the problem I'm facing now is that the compute node is not being deployed, something must be wrong in the parameters I'm passing to `openstack overcloud deploy`

cescgina:
recheck

cescgina:
@fmount @fultonj with the current version all the adoption tests pass, testing in https://review.rdoproject.org/r/c/testproject/+/53209, last run was green https://review.rdoproject.org/zuul/build/fbdcc60f121f4fef9f2d3a9881bc7808 please take a look when free

devsetup/tripleo/initial-ceph.conf
```diff
@@ -0,0 +1,2 @@
+[global]
+osd_pool_default_size=1
```
fmount, line 2: This represents a workaround for a 3+1 scenario.
I'm ok with the rest of the patch but I'm wondering if we want to merge the workaround, or hold this patch until a 3+3 topology (that doesn't need this) is in place.

devsetup/tripleo/ceph.sh
```diff
@@ -61,4 +61,6 @@ openstack overcloud ceph deploy \
     --ceph-spec ceph_spec.yaml \
     --network-data network_data.yaml \
     --cephadm-default-container \
-    --output deployed_ceph.yaml
+    --output deployed_ceph.yaml \
+    --config initial-ceph.conf \
+    --single-host-defaults # TODO: jgilaber only required have ceph working with only 1 compute node should remove it once we move to have 3 computes
```
fmount, line 66: `--single-host-defaults`  represents a workaround for a 3+1 scenario.
I'm ok with the rest of the patch but I'm wondering if we want to merge the workaround, or hold this patch until a 3+3 topology (that doesn't need this) is in place.

cescgina: @fmount our plan (from CI side) was to merge this workaround clearly marked as such, so we can set a periodic job already to make sure changes anywhere do not break it (or we can tell early if they do). And then soon come back and remove the workaround to setup the 3+3 scenario

fmount: Ack, so +1 from me: the only thing I'm not sure is if `--single-host-defaults` is required in this contex (not sure you have the chance to test it)

cescgina: sure, can verify that later today, once the current run finishes

cescgina: can confirm that `--single-host-defaults` is not needed, build https://review.rdoproject.org/zuul/build/65131a41dae04ce3bd8bfab367d594d2 passed without it, as can be seen in https://logserver.rdoproject.org/09/53209/9/check/data-plane-adoption-osp17-multinode-to-extracted-crc-with-ceph/65131a4/controller/ci-framework-data/logs/ci_script_010_run_tripleo.log, the ceph deploy command used is `openstack overcloud ceph deploy --tld localdomain --ceph-spec ceph_spec.yaml --network-data network_data.yaml --cephadm-default-container --output deployed_ceph.yaml --config initial-ceph.conf`

fmount: +1

PR #839, opened 2024-05-28
Enable chassis as gw for standalone adoption jobs

cescgina:
Due to recent changes we need to have enable-chassis-as-gw
set in order for chassis to be used as gw. In standalone
since single node is involved if this is not set the chassis
will not host router gw ports and will have issues.

### DIFF
--- a/devsetup/standalone/role.j2
+++ b/devsetup/standalone/role.j2
@@ -28,6 +28,8 @@
     {{ net.name }}:
       subnet: {{ net.name.lower() }}_subnet
 {%- endfor %}
+  RoleParametersDefault:
+    OVNCMSOptions: "enable-chassis-as-gw"
   ServicesDefault:
     - OS::TripleO::Services::Aide
     - OS::TripleO::Services::AodhApi


lewisdenny:
/lgtm

Some more context: 
The patch that introduces this requirement: https://review.opendev.org/c/openstack/neutron/+/908325/4

karelyatin (approved):
/lgtm

marios (approved):
lgtm

cescgina:
> So this will add the option to pre-adoption cloud, what about after adoption? as that requires edpm_enable_chassis_gw=true to be set

@karelyatin I guess we should add that to the dataplane ansibleVars then? https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/dataplane_adoption/tasks/main.yaml#L282

karelyatin:
> > So this will add the option to pre-adoption cloud, what about after adoption? as that requires edpm_enable_chassis_gw=true to be set
> 
> @karelyatin I guess we should add that to the dataplane ansibleVars then? https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/dataplane_adoption/tasks/main.yaml#L282

@cescgina yes seems so

PR #849, opened 2024-06-10
Ensure there is only one metallb controller manager running

jcapiitao:
Applying OKD specific patches makes the MetalLB controller manager to recreate a replica. This creates a time window in which there are two replicas, one running and the other terminating. If we run the command 'oc wait pod' during this timing, then the command wait for both to be ready and fail after timeout.
To avoid this situation we add a task to wait to have only one replica available. This is a workaround solution.


fao89 (approved):

fao89:
recheck

jcapiitao:
recheck

jcapiitao:
recheck

jcapiitao:
recheck

fao89 (approved):

PR #845, opened 2024-06-06
remove creation of custom nova dataplane service

SeanMooney:
The default nova dataplane service now supports an optional
nova-extra-config configmap

This change removes the creation of the custom service
but keeps the creation fo the nova-extra-config configmap

Depends-On: https://github.com/openstack-k8s-operators/dataplane-operator/pull/920
Related: [OSPRH-6407](https://issues.redhat.com//browse/OSPRH-6407)

PR #850, opened 2024-06-11
Revert "Use DataPlaneService.DataSources"

SeanMooney:
Reverts openstack-k8s-operators/install_yamls#846

this creates a depency cycle between install_yamls and the dataplane operator.

reverting this change is one way to possible break this other wise we will need to force merge

https://github.com/openstack-k8s-operators/dataplane-operator/pull/920 and https://github.com/openstack-k8s-operators/install_yamls/pull/845

### DIFF
--- a/scripts/gen-nova-custom-dataplane-service.sh
+++ b/scripts/gen-nova-custom-dataplane-service.sh
@@ -27,10 +27,9 @@ cat <<EOF >>kustomization.yaml
       path: /metadata/name
       value: nova-custom
     - op: add
-      path: /spec/dataSources
+      path: /spec/configMaps
       value:
-        - configMapRef:
-            name: nova-extra-config
+        - nova-extra-config
 EOF
 
 # Create the nova-extra-config CM based on the provided config file


SeanMooney:
settign hold for now as i think force merging 

https://github.com/openstack-k8s-operators/dataplane-operator/pull/920 and https://github.com/openstack-k8s-operators/install_yamls/pull/845

will likely be the better option but I'm waiting on ci results on 

https://github.com/openstack-k8s-operators/install_yamls/pull/845 to confirm that they two work correctly

scripts/gen-nova-custom-dataplane-service.sh
```diff
@@ -27,10 +27,9 @@ cat <<EOF >>kustomization.yaml
       path: /metadata/name
       value: nova-custom
     - op: add
-      path: /spec/dataSources
+      path: /spec/configMaps
       value:
-        - configMapRef:
-            name: nova-extra-config
```
SeanMooney, line 33: so the reason this is problematic is this si removing the the ssh-key and cell secreate as it s a direct replacement of the list.

so what ti might have to do instead of the revert is a patch to hanel that proerply which I'm going to work on as a third option.
then when we change ci results for all 3 we can decide how to proceed

SeanMooney: im hoping we can proceed with https://github.com/openstack-k8s-operators/install_yamls/pull/851 instead and i can abandon this if that works

SeanMooney:
https://github.com/openstack-k8s-operators/install_yamls/pull/851 passed ci so ill close this in favor of that

PR #852, opened 2024-06-12
Add env var to select nodes for 'make nncp'

abays:
Allow for the selection of a subset of OCP nodes to be targeted by NNCPs. Certain adoption testing scenarios need something like this.  It does not affect default behavior.


fao89:
recheck

fao89 (approved):

PR #853, opened 2024-06-14
Add retries to NNCP deployment

lewisdenny:
Sometimes applying the NNCP CR fails due to once of the kubernetes-nmstate probes failing, usually the api-server.

There are two ideas considered to fix the issue:
1. Extend the probe timeout
2. Delete and reapply the NNCP CR

Results:
1. Sadly the probe timeout is hard coded [1] [2]
2. This is the path I choose, deleting the NNCP/NNCE resource and reapplying it recovers the issue and gives the OCP environment time to settle.

[1] https://github.com/nmstate/kubernetes-nmstate/issues/831
[2] https://github.com/nmstate/kubernetes-nmstate/blob/272b29459ca99a9de62af24f509afe9f3deb058f/pkg/probe/probes.go#L59-L71JIRA: https://issues.redhat.com/browse/OSPRH-7605

### DIFF
--- a/scripts/retry_make_nncp.sh
+++ b/scripts/retry_make_nncp.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+n=0
+retries="${1:-5}"  # Number of retries with a default value of 5
+
+while true; do
+    make nncp && break
+    make nncp_cleanup
+    n=$((n+1))
+    if (( n >= retries )); then
+        echo "Failed to run 'make nncp' target. Aborting"
+        exit 1
+    fi
+    sleep 10
+done
--- a/Makefile
+++ b/Makefile
@@ -435,6 +435,7 @@ NNCP_CTLPLANE_IPV6_ADDRESS_SUFFIX   ?=10
 NNCP_GATEWAY_IPV6                   ?=fd00:aaaa::1
 NNCP_DNS_SERVER_IPV6                ?=fd00:aaaa::1
 NNCP_ADDITIONAL_HOST_ROUTES         ?=
+NNCP_RETRIES ?= 5
 
 # MetalLB
 ifeq ($(NETWORK_ISOLATION_USE_DEFAULT_NETWORK), true)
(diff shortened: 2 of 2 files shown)


lewisdenny:
recheck
no pod logs captured to see whats happening in the ansibleee job :/

scripts/retry_make_nncp.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+n=0
+retries="${1:-5}"  # Number of retries with a default value of 5
+
+while true; do
+    make nncp && break
+    make nncp_cleanup
```
SeanMooney, line 8: i feel like this is not a stable pattern

when we delete and recreate the resource  even if its the same content its going to be a separate object at the k8s level so without an exponential backoof i.e. waiting longer each time we call make nncp i think its possible that we ould be in a loop just because we kept restarting before it could make progress.

should we increase the NNCP_TIMEOUT slightly on each retry?

lewisdenny:
recheck

unrelated failure in post job collect logs failed:
```
kex_exchange_identification: read: Connection reset by peer

rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
rsync error: unexplained error (code 255) at io.c(226) [Receiver=3.1.3]
```

scripts/retry_make_nncp.sh
lewisdenny: >i feel like this is not a stable pattern
It's not ideal but it's a worst case recovery, we already give the NNCP CR 240 seconds to come up and I haven't seen any jobs failing due to timeout. 

>i think its possible that we ould be in a loop just because we kept restarting before it could make progress.
I haven't seen any job fail due to NNCP CR taking over 240s to apply. This loop is limited to 5 iterations so max it would be stuck for is 20mins and this would only be if there is a real issue with the NNCP CR that needs to be fixed, aka a real bug.

>should we increase the NNCP_TIMEOUT slightly on each retry?
No I don't believe we should, I believe it's high enough already unless you have seen jobs failing to apply the NNCP in under 240 seconds?

bshephar (approved):
If it helps bring some additional stability to CI jobs, then lgtm.

ciecierski (approved):
I had problem with nncp on my local environment and this pr fixed it so I'm giving LGTM. However I would be also interested why we need this oc apply retires in the first place.

fao89 (approved):

PR #857, opened 2024-06-19
Fix: devsetup's crc_scrub make target

Akrog:
Existing `crc_scrub` target will fail if the CRC deployment has already been removed.

It also fails to remove the `crc` command.

This patch changes the target so that the scrub (removal of the crc downloaded bits) works as expected, which is convenient to change  the CRC version.

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -176,9 +176,10 @@ crc_cleanup: ## Destroys the CRC env, but does NOT clear ( --clear-cache ) the c
 	sudo update-ca-trust
 
 .PHONY: crc_scrub
-crc_scrub: crc_cleanup
+crc_scrub:
+	make crc_cleanup || true
 	rm -rf ~/.crc
-	sudo rm $(which crc)
+	sudo rm $$(which crc)
 
 .PHONY: crc_attach_default_interface
 crc_attach_default_interface:


ciecierski (approved):
LGTM

devsetup/Makefile
```diff
@@ -176,9 +176,10 @@ crc_cleanup: ## Destroys the CRC env, but does NOT clear ( --clear-cache ) the c
 	sudo update-ca-trust
 
 .PHONY: crc_scrub
-crc_scrub: crc_cleanup
+crc_scrub:
+	make crc_cleanup || true
 	rm -rf ~/.crc
-	sudo rm $(which crc)
+	sudo rm `which crc`
```
abays, line 182: Isn't ` `` ` deprecated in favor of `$(...)`?  At least I recall tools like `shellcheck` complaining about the former.

Akrog: Yes it is, but `$` has meaning in a `Makefile` so it doesn't work.
This was the easiest way to circumvent that, since escaping it with '\$' didn't seem to work for me.

abays: I wonder if that's why we use `$$` elsewhere in certain places?  For example [1].

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/884e173955b8ba6cbf182e1f564990d703a963a7/Makefile#L688

abays:
recheck

abays (approved):
/lgtm

PR #860, opened 2024-06-20
[devsetup] Assign IP by MAC not by hostname

Akrog:
Our current script `devsetup/scripts/interfaces_setup.sh` that is called from the make target `crc_attach_default_interface` is creating the static IP entry in the DHCP libvirt network using the CRC VM hostname.

This hostname used to be `crc` but now I'm seeing it being created as `crc-vlf7c-master-0`, which means that the static IP we want for the default interface will not be assigned.

This patch changes the way the DHCP will identify the VM and assign the static IP. When a MAC_ADDRESS is provided to the script (that's what the make target does) it will not use the hostname but the MAC address.

For cleanup the script `devsetup/scripts/interfaces-setup-cleanup.sh` will first try to remove the dhcp entry by MAC and if that fails it will try using the host.


fmount (approved):
/lgtm

PR #862, opened 2024-06-25
test dataplane prep targets

fao89:


fao89:
recheck

fao89:
recheck

fao89:
recheck

scripts/clone-operator-repo.sh
```diff
@@ -65,8 +65,15 @@ function git_clone_checkout {
 
     pushd ${OPERATOR_BASE_DIR}
 
-    echo "Cloning repo: git clone ${git_opts} ${branch} ${repo} ${operator}-operator${ALT_CHECKOUT:+-$ALT_CHECKOUT}"
-    git clone ${git_opts} ${branch} ${repo} ${operator}-operator${ALT_CHECKOUT:+-$ALT_CHECKOUT}
+    if [ ! -d ${operator}-operator${ALT_CHECKOUT:+-$ALT_CHECKOUT} ]; then
```
slagle, line 68: is this change necessary as the checkout should always be deleted by edpm_deploy_cleanup. I don't think this will be an issue any longer with your other Makefile changes.

fao89: you are right!

slagle (commented):
LGTM

ASBishop:
I encountered the problem, found this patch and tried it, and it worked perfectly. I recommend you remove the DNM and commit message, as it lgtm!

PR #864, opened 2024-06-27
bump kuttl to 0.17.0

stuggi:


gibizer:
would be nice to run some test with depends on to see if this does not break any operator tests.

gibizer:
lets see if depends-on works in https://github.com/openstack-k8s-operators/nova-operator/pull/805

PR #865, opened 2024-06-27
WIP - Clone openstack-operator repo during prepare calls

viroel:
When running install_yamls with CHECKOUT_FROM_OPENSTACK_REF enabled, the clone-operator-repo script make sure that we clone openstack-operator repo, since we need it to check for operators references. But when running with this flag disabled, some make targets fail to deploy since the openstack-operator is referenced but not previously cloned.


viroel:
oh nevermind, seems to be fixed by https://github.com/openstack-k8s-operators/install_yamls/commit/f1125d8352aa90ef2a231126235d4de63994930c

PR #873, opened 2024-07-16
Optionally disable Barbican and Manila

xek:
Fix the Makefile and standalone.sh to forward
the appropriate env variables for disabling
Barbican and Manila.

Jira: https://issues.redhat.com/browse/OSPRH-6857


devsetup/Makefile
```diff
@@ -60,6 +60,9 @@ EDPM_COMPUTE_ADDITIONAL_HOST_ROUTES ?= '[]'
 TLS_ENABLED ?= false
 DNS_DOMAIN ?= localdomain
 
+BARBICAN_SERVICE_ENABLED ?= false
+MANILA_SERVICE_ENABLED ?= false
+
```
gouthampacha, line 65: right now; both these services are enabled on the standalone deployment; your change will break that: https://github.com/openstack-k8s-operators/install_yamls/blob/ff5075d356345f9e7e945ac53a164aebae055c52/devsetup/standalone/openstack.sh#L23-L24

Please default to True and disable these locally if you wish.

xek: Thanks, I made the change

devsetup/Makefile
```diff
@@ -60,6 +60,9 @@ EDPM_COMPUTE_ADDITIONAL_HOST_ROUTES ?= '[]'
 TLS_ENABLED ?= false
 DNS_DOMAIN ?= localdomain
 
+BARBICAN_SERVICE_ENABLED ?= false
```
marios, line 63: here we have default to false

in standalone.sh we have default to true (line 51/52):

```
BARBICAN_ENABLED=${BARBICAN_ENABLED:-true}
MANILA_ENABLED=${MANILA_ENABLED:-true}
```


can we be consistent on this

devsetup/scripts/standalone.sh
```diff
@@ -138,6 +138,8 @@ export COMPUTE_DRIVER=${COMPUTE_DRIVER:-"libvirt"}
 export IP=${IP}
 export GATEWAY=${GATEWAY}
 export STANDALONE_VM=${STANDALONE_VM}
+export BARBICAN_ENABLED=${BARBICAN_ENABLED}
```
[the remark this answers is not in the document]
xek: yep, I'll just add the export keyword on line 51-52

devsetup/Makefile
xek: Right, that's an error, it should default to true, as to not change the previous behavior

devsetup/scripts/standalone.sh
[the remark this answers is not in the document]
xek: I added the keywords for all other duplicated variable forwarding/initialization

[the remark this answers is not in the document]
xek: Sorry, when I did that, I didn't notice these exports are placed in a different script, the $CMDS_FILE, so all of these should stay in $CMDS_FILE as is. So it looks to me no change is required on this comment?

marios (approved):

PR #874, opened 2024-07-18
Adoption: remove podman pin

jistr:
Resolves: [OSPRH-8539](https://issues.redhat.com//browse/OSPRH-8539)


bogdando:
let's check this in testproject for adoption jobs to not fail with this change.
It might be that we still want the pinning in upstream CI as we are using RDO

jistr:
Marking as ready to review. I tested this on periodic job, with testproject both upstream and downstream. All green. (More details in the Jira card.)

fao89 (approved):

PR #875, opened 2024-07-21
Set finalizers to null before delete

holser:
In some cases project cannot be removed due to transition dependency. To force a removal we need to set finalizers to null

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -608,6 +608,7 @@ namespace_cleanup: ## deletes the namespace specified via NAMESPACE env var, als
 	$(eval $(call vars,$@))
 	make keystone_cleanup
 	make mariadb_cleanup
+	oc patch project ${NAMESPACE} --type=merge -p '{"metadata": {"finalizers":null}}'
 	oc delete project ${NAMESPACE}
 	${CLEANUP_DIR_CMD} ${OUT}/${NAMESPACE}
 


Makefile
```diff
@@ -608,6 +608,7 @@ namespace_cleanup: ## deletes the namespace specified via NAMESPACE env var, als
 	$(eval $(call vars,$@))
 	make keystone_cleanup
 	make mariadb_cleanup
+	oc patch project ${NAMESPACE} --type=merge -p '{"metadata": {"finalizers":null}}'
```
holser, line 611: very similar code is [data-plane-adoption cleanup](https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/pcp_cleanup/tasks/main.yaml#L10)

ciecierski:
recheck

jistr (approved):

Makefile
ciecierski: similar, but why are here mariadb_cleanup and keystone_cleanup?

ciecierski: @stuggi can you advise if we can remove these two make targets from namespace_cleanup?

karelyatin: me not sure if deleting the project like this will not leave stale resources(pods/services etc) in that namespace? shouldn't we ensure everything deleted before deleting the namespace?

stuggi: this is true, it is not guaranteed that the delete will work. if there are any resources in the namespace which have a finalizer set, they will block the deletion of the namespace. So any other cleanup should be done before, but this target could be used afterwards as a project cleanup.

holser:
I am closing this one as setting finalizers don't solve the issues we have

PR #889, opened 2024-08-01
Add config-download for networker multinode adoption CI

marios:
As part of [1] adds the bits needed for the multinode OSP17 deployment with dedicated networker nodes.

[1] https://issues.redhat.com/browse/OSPRH-8810

### DIFF
--- a/devsetup/tripleo/tripleo_install.sh
+++ b/devsetup/tripleo/tripleo_install.sh
@@ -21,20 +21,35 @@ source stackrc
 openstack overcloud network provision --output network_provision_out.yaml ./network_data.yaml
 openstack overcloud network vip provision --stack overcloud --output vips_provision_out.yaml ./vips_data.yaml
 
+# check if hostnamemap contains networkers
+networker_nodes="FALSE"
+config_download="config-download.yaml"
+if (grep -q networker hostnamemap.yaml); then
+    networker_nodes="TRUE"
+    config_download="config-download-networker.yaml"
+fi
+
 # update the config-download with proper overcloud hostnames
 control0=$(grep "overcloud-controller-0" hostnamemap.yaml  | awk '{print $2}')
 control1=$(grep "overcloud-controller-1" hostnamemap.yaml  | awk '{print $2}')
 control2=$(grep "overcloud-controller-2" hostnamemap.yaml  | awk '{print $2}')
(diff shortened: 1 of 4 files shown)


devsetup/tripleo/overcloud_services.yaml
```diff
@@ -40,7 +40,8 @@ parameter_defaults:
   DockerPuppetDebug: true
   ContainerCli: podman
   ControllerCount: 3
-  ComputeCount: 3
+  ComputeCount: 2
```
marios, line 43: just for testing 
will find a better place to override this

devsetup/tripleo/tripleo_install.sh
```diff
@@ -94,6 +109,6 @@ openstack overcloud deploy --stack overcloud \
     -e /home/zuul/containers-prepare-parameters.yaml -e /usr/share/openstack-tripleo-heat-templates/environments/podman.yaml \
     -e /usr/share/openstack-tripleo-heat-templates/environments/low-memory-usage.yaml \
     -e /usr/share/openstack-tripleo-heat-templates/environments/debug.yaml --validation-warnings-fatal ${ENV_ARGS} ${CEPH_OVERCLOUD_ARGS} \
-    -e /home/zuul/overcloud_services.yaml -e /home/zuul/config-download.yaml \
+    -e /home/zuul/overcloud_services.yaml -e /home/zuul/$config_download \
```
cescgina, line 112: if I understand correctly, now we have a `ComputeCount: 3` in `overcloud_services` and a `ComputeCount: 2` in the new `config-download-networker`. Do we know how the precendence is handled? Is it like ansible when only the last occurrence counts?

marios: yes exactly
with the tripleo-heat-templates environments last one wins and I am relying on that config-download-networker.yaml to come after services.yaml

it allows us to keep it simpler here with the networker vars in one file

cescgina: ack, in that case maybe I would left a comment noting that we rely on that order to get the right amount of computes, for future reference, other than that, the PR lgtm

marios:
can someone add the lgtm label @raukadah or @viroel please if when you next have some time thank you!

PR #890, opened 2024-08-12
Allow control of tlsEnabled for OpenStackDataplaneNodeSet

stuggi:
For testing tls enable as day2 lets have a parameter where one can control if tls should be set on the nodeset

Related: [OSPRH-6624](https://issues.redhat.com//browse/OSPRH-6624)


abays (approved):
/lgtm

PR #893, opened 2024-08-12
Remove the hardcoded dataplane secret

fao89:
https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L372


fmount (commented):
/lgtm

PR #898, opened 2024-08-19
Keep Octavia disabled by default on 17.1 deployments

cescgina:
Enabling octavia by default in the 17.1 envs seems to crash adoption
jobs. The problem seems to be that octavia is not yet adopted, so when
running tempest the load-balancer is announced, but is not really
functional because the tripleo-cleanup service kills it, leading to
tempest crashing.

Depends-On: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/567


cescgina:
recheck

PR #900, opened 2024-08-19
Fix IP_ADRESS_SUFFIX for multinode

bogdando:
When defaulting it to the exported IP_ADRESS_SUFFIX env, iteration won't work for consequent make targets invoked with different IP_ADRESS_SUFFIX'es.

Fix that by evaluating IP_ADRESS_SUFFIX based on
100+EDPM_COMPUTE_SUFFIX.


fao89 (approved):

bogdando:
oops, I've got confused by the approved label

PR #901, opened 2024-08-19
Add FIRSTBOOT_EXTRA for EDPM node gen script

bogdando:
This arg may be used for additional host configs, like repo setup commands which should be executed for all EDPM nodes.

For standalone and tripleo devsetup targets, there is REPO_SETUP_CMDS which can only be executed on a given node, where either of deploy scripts gets executed, which doesn't work for multinode cases.

### DIFF
--- a/devsetup/scripts/gen-edpm-node.sh
+++ b/devsetup/scripts/gen-edpm-node.sh
@@ -23,6 +23,7 @@ MY_TMP_DIR="$(mktemp -d)"
 trap 'rm -rf -- "$MY_TMP_DIR"' EXIT
 
 EDPM_SERVER_ROLE=${EDPM_SERVER_ROLE:-"compute"}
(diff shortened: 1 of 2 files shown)


devsetup/scripts/gen-edpm-node.sh
```diff
@@ -271,8 +272,14 @@ echo GATEWAY=$GATEWAY >> $NETSCRIPT
 echo DNS1=$DNS >> $NETSCRIPT
 sed -i s/dhcp/none/g $NETSCRIPT
 sed -i /PERSISTENT_DHCLIENT/d $NETSCRIPT
+
+# Additional commands
+
 EOF
 
+touch $FIRSTBOOT_EXTRA
+cat "$FIRSTBOOT_EXTRA" >> ${OUTPUT_DIR}/${EDPM_COMPUTE_NAME}-firstboot.sh
```
fao89, line 281: I don't get it, it is an empty file, when/where do we add content to `/tmp/edpm-firstboot-extra`?

fao89 (approved):

PR #902, opened 2024-08-19
Add ${REDHAT_OPERATORS} var

slagle:
Set REDHAT_OPERATORS=true for installing openstack-operator (and deps)
from redhat-marketplace.

REDHAT_OPERATORS=true make openstack


fao89:
recheck

fao89 (approved):

PR #909, opened 2024-09-12
Minor adjustments for devsetup tripleo multinode

bogdando:
* Use the original ansible to_bool() filter (py3) in jinja2_render.
* Simplify a condition for a supported cells number.
* Copy nova_noceph.yaml from the local t-h-t files instead of heredocs.
* Hide RH_REGISTRY_ credentials, when dumping the containers conf file.
* Fix a typo in the supported cells number message.

### DIFF
--- a/devsetup/scripts/tripleo.sh
+++ b/devsetup/scripts/tripleo.sh
@@ -48,8 +48,8 @@ source ${SCRIPTPATH}/common.sh
 # We'll use the NTP_SERVER environmental variable to define the NTP server to use, e.g.:
 # export NTP_SERVER=pool.ntp.org
 
-if [ $EDPM_COMPUTE_CELLS -eq 2 ] || [ $EDPM_COMPUTE_CELLS -gt 3 ] || [ $EDPM_COMPUTE_CELLS -eq 0 ] ; then
-    echo "Only a main cell0 plus a 2 additional compute cells supported yet!"
+if [ $EDPM_COMPUTE_CELLS -ne 1 -a $EDPM_COMPUTE_CELLS -ne 3 ]; then
+    echo "Only a main cell1 plus a 2 additional compute cells supported yet!"
     exit 1
 fi
 
@@ -81,11 +81,6 @@ sudo dnf install -y podman python3-tripleoclient util-linux lvm2
 sudo hostnamectl set-hostname undercloud.localdomain
 sudo hostnamectl set-hostname undercloud.localdomain --transient
 
-cat >\$HOME/nova_noceph.yaml <<__EOF__
-parameter_defaults:
-    NovaEnableRbdBackend: false
-__EOF__
-
 export HOST_PRIMARY_RESOLV_CONF_ENTRY=${HOST_PRIMARY_RESOLV_CONF_ENTRY}
 export INTERFACE_MTU=${INTERFACE_MTU:-1500}
 export NTP_SERVER=${NTP_SERVER:-"pool.ntp.org"}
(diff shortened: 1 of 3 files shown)


devsetup/scripts/tripleo.sh
```diff
@@ -101,14 +96,13 @@ export OCTAVIA_ENABLED=${OCTAVIA_ENABLED}
 export TELEMETRY_ENABLED=${TELEMETRY_ENABLED:-true}
 
 set +x
-if [[ -f \$HOME/containers-prepare-parameters.yaml ]]; then
-    echo "Using existing containers-prepare-parameters.yaml - contents:"
-    cat \$HOME/containers-prepare-parameters.yaml
-else
+if [ ! -f \$HOME/containers-prepare-parameters.yaml ]; then
```
marios, line 99: fair enough. 
We may even just remove the conditional altogether, though keeping it will still allow someone to override if they needed to.

for the record/for other reviewers, this was added when we used to deploy tripleo wallaby in upstream jobs but wanted to override that for the downstream job (where we specified a containers-prep file like https://code.engineering.redhat.com/gerrit/c/openstack/tripleo-ci-internal-jobs/+/445396/28/playbooks/data_plane_adoption/containers-prepare-parameters.yaml

marios: if we are going to keep this conditional if then can we please re-add the else with the echo statement

```
else
    echo "Using existing containers-prepare-parameters.yaml - contents:"
    cat \$HOME/containers-prepare-parameters.yaml
```

This isn't being used in any jobs any more but if we are leaving the ability for this in local run/deployment the debug info will be useful if someone DOES want to use a custom containers prep file.

devsetup/scripts/tripleo.sh
```diff
@@ -101,14 +96,13 @@ export OCTAVIA_ENABLED=${OCTAVIA_ENABLED}
...
     login_args=" "
     [ "\$RH_REGISTRY_USER" ] && [ -n "\$RH_REGISTRY_PWD" ] && login_args="--enable-registry-login"
+    set -x
     openstack tripleo container image prepare default \
         --output-env-file \$HOME/containers-prepare-parameters.yaml \${login_args}
+    set +x
     # Use wallaby el9 container images
     sed -i 's|quay.io/tripleowallaby$|quay.io/tripleowallabycentos9|' \$HOME/containers-prepare-parameters.yaml
```
marios, line 107: Since you're here and in the mood for improvements, you can also remove this sed since there is no more wallaby tripleo deploy here/

devsetup/scripts/tripleo.sh
```diff
@@ -101,14 +96,13 @@ export OCTAVIA_ENABLED=${OCTAVIA_ENABLED}
...
 
 set +x
-if [[ -f \$HOME/containers-prepare-parameters.yaml ]]; then
-    echo "Using existing containers-prepare-parameters.yaml - contents:"
-    cat \$HOME/containers-prepare-parameters.yaml
-else
+if [ ! -f \$HOME/containers-prepare-parameters.yaml ]; then
     login_args=" "
     [ "\$RH_REGISTRY_USER" ] && [ -n "\$RH_REGISTRY_PWD" ] && login_args="--enable-registry-login"
+    set -x
```
marios, line 102: will this expose the login args that are placed in the containers-prep file?
i understand you had to add these for local runs but one of the main goals of this work is to add a multicell job. In that case (the ci job) will you just use the existing registry login (that comes from the config repo playbooks and used in the other adoption jobs)?

devsetup/scripts/tripleo.sh
```diff
@@ -248,6 +243,9 @@ if [[ "$EDPM_COMPUTE_CEPH_ENABLED" == "true" ]]; then
 fi
 
 if [[ -f $HOME/containers-prepare-parameters.yaml ]]; then
+    echo "Using existing containers-prepare-parameters.yaml - contents:"
+    # requires 'make download_tools'
+    yq '.parameter_defaults.ContainerImageRegistryCredentials="{ ...snip... }"' $HOME/containers-prepare-parameters.yaml
```
marios, line 248: ok i see this should address both my comments above, one about the echo and the other about the credentials. As long as we keep setting the user/pass to "" for the ci jobs case (the actual login happens with zuul secrets in config repo)

marios (commented):
ok but please lets sanity check the multinodes as usual

devsetup/scripts/tripleo.sh
bogdando: no, as ``login_args="--enable-registry-login"`` and contains no creds.

bogdando: In CI, there would be no RH_REGISTRY creds, so it would work by usual for CI meanings.

bogdando: I got now that you mean, removed ``-x``

bogdando:
> ok but please lets sanity check the multinodes as usual

https://review.rdoproject.org/r/c/testproject/+/54199

jistr:
LGTM

PR #912, opened 2024-09-19
DNM: Total cleanup of all operator resources

jpodivin:
Certain resources may be restored after cleanup step. By removing all `olm` labeled resources from the cluster we can ensure that their number is minimized. 

Operator resources[0] are unfortunately still persisting :/ 

[0] https://docs.redhat.com/en/documentation/openshift_container_platform/4.12/html/operatorhub_apis/operator-operators-coreos-com-v1#operator-operators-coreos-com-v1


jpodivin:
I can't get rid of all of them, and it doesn't seem they do any damage. It may be something to revisit if we find out they do.

PR #915, opened 2024-09-24
Pass NEUTRON_KUTTL_DIR to kuttl neutron env

booxter:
The variable is now used by one of the tests that was recently added to main with:

https://github.com/openstack-k8s-operators/neutron-operator/pull/406

Without setting the variable, the test fails to execute the script and fails.

Related: https://issues.redhat.com//browse/OSPRH-4296

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1748,7 +1748,7 @@ cinder_kuttl: kuttl_common_prep cinder cinder_deploy_prep ## runs kuttl tests fo
 
 .PHONY: neutron_kuttl_run
 neutron_kuttl_run: ## runs kuttl tests for the neutron operator, assumes that everything needed for running the test was deployed beforehand.
-	kubectl-kuttl test --config ${NEUTRON_KUTTL_CONF} ${NEUTRON_KUTTL_DIR} --namespace ${NAMESPACE}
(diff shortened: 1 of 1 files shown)


booxter:
@karelyatin FYI

booxter:
recheck

fao89 (commented):
/approve

booxter:
@karelyatin please confirm this is the right thing to do for kuttl

karelyatin:
@booxter thx for the fix. Just the referenced patch and Jira not related to the usage, the var was injected previously. Also we do similar in ovn. So we need to have similar fix for OVN_KUTTL_DIR too.
But feels like it would be better if we can avoid the usage of this var in our tests, /me will check if that's doable as that would be better than this.

karelyatin:
> @booxter thx for the fix. Just the referenced patch and Jira not related to the usage, the var was injected previously. Also we do similar in ovn. So we need to have similar fix for OVN_KUTTL_DIR too. But feels like it would be better if we can avoid the usage of this var in our tests, /me will check if that's doable as that would be better than this.

Proposed https://github.com/openstack-k8s-operators/ovn-operator/pull/396 and https://github.com/openstack-k8s-operators/neutron-operator/pull/465 , can you check if ok to go that way?

booxter:
@karelyatin thanks for proper fixes, these are definitely better.

PR #917, opened 2024-09-25
Ensure EDPM nodes have a unique NVMe NQN

ASBishop:
When creating EDPM compute nodes, remove stale files in /etc/nvme that result in nodes having the same NVMe NQN.

There are provisions to generate a unique NQN when the nvme-cli package is installed or when os-brick accesses the NVMe subsystem, but these won't overwrite existing files.

### DIFF
--- a/devsetup/scripts/gen-edpm-node-bgp.sh
+++ b/devsetup/scripts/gen-edpm-node-bgp.sh
@@ -218,6 +218,9 @@ sed -i /PERSISTENT_DHCLIENT/d /etc/sysconfig/network-scripts/ifcfg-eth0
 systemctl restart NetworkManager
 ip r add default via 192.168.122.1
 
+# Remove stale artifacts that are auto-generated during EDPM deployment
+rm -f /etc/nvme/hostid /etc/nvme/hostnqn
+
 EOF
 done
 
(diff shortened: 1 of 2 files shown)


devsetup/scripts/gen-edpm-node-bgp.sh
```diff
@@ -218,6 +218,9 @@ sed -i /PERSISTENT_DHCLIENT/d /etc/sysconfig/network-scripts/ifcfg-eth0
 systemctl restart NetworkManager
 ip r add default via 192.168.122.1
 
+# Remove stale artifacts that are auto-generated during EDPM deployment
+rm -f /etc/nvme/hostid /etc/nvme/hostnqn
```
Akrog, line 222: nit: This is ok, but we could also create the right values instead of making os-brick do it:

```
/usr/sbin/dmidecode -s system-uuid > /etc/nvme/hostid
/usr/sbin/nvme gen-hostnqn > /etc/nvme/hostnqn
```

Akrog (approved):
/lgtm

devsetup/scripts/gen-edpm-node-bgp.sh
fmount: @ASBishop you can follow up with the above.

Akrog: There's no real need to change it, because os-brick will recreate those files.
It's just my preference. I like an explicit recreation instead of just removing them, but everything will work either way.

ASBishop: It's unlikely that os-brick will need to recreate the files because the edpm deployment installs the nvme-cli package, and it will recreate them first.

PR #919, opened 2024-10-01
Deploy Rook operator

fmount:
This patch allows to plug `Rook` to the existing `openstack-k8s-operators` development workflow. With this patch is possible to deploy `Rook` operator, and it can be used to attach an external `Ceph` cluster (already consumed by `OpenStack`) to `OpenShift`.
The same code can be used to deploy an internal `Ceph` cluster, but is out of scope for the current workflow.

This process is extensively documented [here](https://gist.github.com/fmount/5cf340799026bb03abc8928f20705d76) and it can be proposed as a doc patch as part of a follow up.


abays (approved):
/lgtm

PR #920, opened 2024-10-03
Use stringData when creating root password secret

rabi:
When using `data`,  `NodeRootPassword` value should be base64 encoded. If using a normal string for $EDPM_ROOT_PASSWORD secret creation fails.

### DIFF
--- a/scripts/gen-edpm-bmset-password-secret.sh
+++ b/scripts/gen-edpm-bmset-password-secret.sh
@@ -24,7 +24,7 @@ metadata:
   name: baremetalset-password-secret
   namespace: ${NAMESPACE}
 type: Opaque
-data:
+stringData:
   NodeRootPassword: ${EDPM_ROOT_PASSWORD}
 EOF
 }


scripts/gen-edpm-bmset-password-secret.sh
```diff
@@ -24,7 +24,7 @@ metadata:
   name: baremetalset-password-secret
   namespace: ${NAMESPACE}
 type: Opaque
-data:
+stringData:
```
fmount, line 27: Ok I see everything starts with `BM_ROOT_PASSWORD` passed as input, that is propagated here.
I don't have big concerns to prefer `data` over `stringData` (as we're delegating k8s to encode the secret for us), so I'm ok with the patch if it ease the development workflow.

fmount (approved):
/lgtm

scripts/gen-edpm-bmset-password-secret.sh
fmount: @abays FYI ^

abays: Does our CI automation currently use this?  I just wonder if they're already passing base64-encoded values through the env var (`BM_ROOT_PASSWORD`), such that if it was injected into `stringData`, k8s would try to  base64-encode something that was already encoded by the user.

fmount: Good question, in that case we need to update how CI pass this info as well. @rabi I'm wondering if we want to clarify this aspect before merging.

rabi: I had checked it before proposing. I didn't find it being used in any of the `openstack-k8s-operator` repos. Unless there is some other places that we should look for.

fmount: ack, feel free to remove the hold when you're comfortable merging the patch then.

abays: I really don't know if something in CI is using this already or not.  It just seems possible to me that some existing job automation might be setting `BM_ROOT_PASSWORD` as an encoded string already.  Maybe we should ask someone on the CI side to comment?  @raukadah, maybe you know something about this?

rabi: I don't see it being used in upstream/downstream CI.

PR #921, opened 2024-10-04
[DNM] Test edpm-update fix

karelyatin:


karelyatin:
confirmed the fix works, the change is pulled downstream

PR #923, opened 2024-10-07
Fix hostnamemap for multi-stack tripleo deploy

bogdando:
Instead of mangling its content inside of the shell script,
statically provide it for overcloud_services j2 template.

Fix missing insertion of *HostnameFormat values which breaks
tripleo deployment.


bogdando:
testproject https://review.rdoproject.org/r/c/testproject/+/54199/6/.zuul.yaml

bogdando:
@marios PTAL

cescgina:
897 merged and it shows no conflicts, so I think we can go ahead with this one

PR #926, opened 2024-10-10
Add rook_deploy target

fmount:
As a follow up of #919 , this one adds the ability to deploy an `internal` `Ceph` cluster using `rook` (both `mons` and `OSDs` are deployed within `OCP`).
Additionally, a `make rook_crc_disk` allows to build a disk and attach it as a device to the existing OCP node (it currently supports crc as the main use case).

### DIFF
--- a/scripts/gen-rook-kustomize.sh
+++ b/scripts/gen-rook-kustomize.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+#
+# Copyright 2024 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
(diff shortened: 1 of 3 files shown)


fao89 (approved):

scripts/gen-rook-kustomize.sh
```diff
@@ -0,0 +1,48 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+set -ex
+
+# expect that the common.sh is in the same dir as the calling script
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+. ${SCRIPTPATH}/common.sh --source-only
+
+if [ -z "$CEPH_IMAGE" ]; then
```
abays, line 22: Is this actually `IMAGE` instead here?  Seems like it might be based on [1].

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/8d26cd388b9423cf2cb79d854499b5e0ef80397f/Makefile#L2169

fmount:
recheck

scripts/gen-rook-kustomize.sh
fmount: oh right, thank you!

abays (approved):
/lgtm

scripts/gen-rook-kustomize.sh
fmount: Done

PR #927, opened 2024-10-10
Add missing services for Cell* roles for tripleo

bogdando:
Align included services with the t-h-t's CellController role.

### DIFF
--- a/devsetup/tripleo/overcloud_roles.yaml
+++ b/devsetup/tripleo/overcloud_roles.yaml
@@ -225,14 +225,17 @@
     - OS::TripleO::Services::BootParams
     - OS::TripleO::Services::CACerts
     - OS::TripleO::Services::Clustercheck
+    - OS::TripleO::Services::Collectd
     - OS::TripleO::Services::ContainerImagePrepare
     - OS::TripleO::Services::ContainersLogrotateCrond
+    - OS::TripleO::Services::Frr
     - OS::TripleO::Services::HAproxy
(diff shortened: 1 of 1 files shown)


bogdando:
@marios Note: Cell roles are not used in multi-node nor standalone CI, so there is no need for a regression check in testproject

bogdando:
testing for a WIP multi-cell job https://review.rdoproject.org/r/c/testproject/+/53212

devsetup/tripleo/overcloud_roles.yaml
```diff
@@ -225,14 +225,17 @@
     - OS::TripleO::Services::BootParams
     - OS::TripleO::Services::CACerts
     - OS::TripleO::Services::Clustercheck
+    - OS::TripleO::Services::Collectd
```
marios, line 228: are these only relevant for the mutli cell job ? If yes then perhaps we need to consider a mutli_cell_overcloud_roles.yaml for that specific case? 

Otherwise can you please ensure to run the ceph/noceph "standard" multinode job at least once when testing this to sanity check we don't need to fix something else there to enable these

bogdando: Those are default for tripleo for cellcontroller role

marios (approved):

cescgina (commented):
/lgtm

fao89 (commented):
/approve

PR #931, opened 2024-10-16
Add gen-edpm-node-bootc.sh for creating bootc based nodes

slagle:
- **Refactor common EDPM node code into gen-edpm-node-common.sh**
- **Add gen-edpm-node-bootc.sh for creating bootc based nodes**


devsetup/scripts/gen-edpm-node-bootc.sh
```diff
@@ -0,0 +1,101 @@
...
+      nameservers:
+        addresses:
+          - ${DATAPLANE_DNS_SERVER}
+EOF
+
+if [ ! -f ${EDPM_BOOTC_OUTPUT_DIR}/cidata.iso ]; then
+    genisoimage -output ${EDPM_BOOTC_OUTPUT_DIR}/cidata.iso -V CIDATA -r -J ${EDPM_BOOTC_OUTPUT_DIR}/user-data ${EDPM_BOOTC_OUTPUT_DIR}/meta-data ${EDPM_BOOTC_OUTPUT_DIR}/network-config
+fi
+
+sudo podman run --rm -it --privileged \
```
rabi, line 81: Any reason we can't just use the image built by edpm-image-builder? We can change the scripts to pass user-data/network-config with 'virt-install --cloud-init`(?)  [1]. 

[1] https://github.com/virt-manager/virt-manager/blob/main/man/virt-install.rst#--cloud-init

slagle: initially, i wanted to make use of the config.toml support as much as possible and that is something used during the build process. given how limited that interface is, I think we probably could just use the already built image and rely on cloud-init entirely instead.

fao89 (approved):

PR #933, opened 2024-10-18
Add NMstate OLM CR name parameters

pablintino:
Mainly used in CI as we use the default values to perform kustomizations or apply additional changes.
If the Subscription and OperatorGroup CR names are extracted to default vars in the Makefile we will be able to know beforehand their names.


abays (approved):
/lgtm

PR #934, opened 2024-10-18
Allow skipping install NMstate

pablintino:
In case it's already installed the user may want to skip installing NMstate.


lewisdenny (approved):
/lgtm

No change to the default behavior but if INSTALL_NMSTATE set to false, nmstate won't be installed, only nncp will be applied.

abays (approved):
/lgtm

PR #944, opened 2024-11-12
Allow enabling monitoring for CRC

umago:
CRC disables cluster monitoring by default but, some operators like Lightspeed requires monitoring. This patch is adding a new configuration environment variable for the crc-setup.sh script to allow enabling it.

More: https://crc.dev/crc/getting_started/getting_started/administrative-tasks/#_starting_monitoring


fmount (approved):
/lgtm

PR #945, opened 2024-11-14
[TripleO Deploy] Enable ssh from all network

karelyatin:
Currently we handle it in rdo-jobs with playbook[1], but that will not work with upcoming 17.1.4 changes related to nftables switch. With this we shouldn't need to handle it in job config.

[1] playbooks/data_plane_adoption/allow_zuul_connect_tripleo.yaml


fmount (approved):
/lgtm

PR #947, opened 2024-11-18
Add watcher target to install_yaml

raukadah:
This pr adds watcher, watcher_cleanup, watcher_deploy and watcher_deploy_cleanup target to install and deploy watcher service.


raukadah:
No longer needed

PR #949, opened 2024-11-20
Tunnel related improvements

Akrog:
In this PR I'm adding some fixes and features helpful for a [PoC I'm working on where we connect a remote edpm node using tunnel](https://github.com/Akrog/rhoso-rhelai/tree/main/remote-edpm).

Changes are as follow:

- Fix `openstack_deploy_prep` target that is setting `CTLPLANE_IP_ADDRESS_PREFIX` to an IPv6 prefix.
- Support changing dataplane kustomization scenario directory location using instead of always assuming it's within the cloned openstack-operator samples directory
- Remove the DHCP static IP assignment when VM is down
- Support post gen edpm kustomize scripts on the `edpm_deploy_prep` target.

More information on these changes is available on the respective patches commit message.

### DIFF
--- a/scripts/gen-edpm-kustomize.sh
+++ b/scripts/gen-edpm-kustomize.sh
@@ -103,6 +103,30 @@ cat <<EOF >>kustomization.yaml
 EOF
 fi
 
+if [ -n "$EDPM_REGISTRY_URL" ]; then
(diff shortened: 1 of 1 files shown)


Akrog:
recheck

PR #951, opened 2024-11-22
Allow crc delete to fail

slagle:
crc delete will fail if it's already been deleted. Handle that
gracefully and just ignore the failure.


stuggi (approved):
/lgtm

PR #955, opened 2024-11-28
Bump ipa image in adoption script and make it a parameter

fmount:
Current failures for no-ceph jobs:

```
+ podman run -d --name freeipa-server-container --sysctl net.ipv6.conf.lo.disable_ipv6=0 --security-opt seccomp=unconfined --ip 10.88.0.2 -e IPA_SERVER_IP=10.88.0.2 -e PASSWORD=fce95318204114530f31f885c9df588f -h ipa.ooo.test --read-only --tmpfs /run --tmpfs /tmp -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v /tmp/ipa-data:/data:Z quay.io/freeipa/freeipa-server:fedora-39 no-exit -U -r OOO.TEST --setup-dns --no-reverse --no-ntp --no-dnssec-validation --auto-forwarders
    Trying to pull quay.io/freeipa/freeipa-server:fedora-39...
    time="2024-11-28T04:16:35-05:00" level=warning msg="Failed, retrying in 1s ... (1/3). Error: initializing source docker://quay.io/freeipa/freeipa-server:fedora-39: reading manifest fedora-39 in quay.io/freeipa/freeipa-server: unknown: Tag fedora-39 was deleted or has expired. To pull, revive via time machine"
    time="2024-11-28T04:16:37-05:00" level=warning msg="Failed, retrying in 1s ... (2/3). Error: initializing source docker://quay.io/freeipa/freeipa-server:fedora-39: reading manifest fedora-39 in quay.io/freeipa/freeipa-server: unknown: Tag fedora-39 was deleted or has expired. To pull, revive via time machine"
    time="2024-11-28T04:16:38-05:00" level=warning msg="Failed, retrying in 1s ... (3/3). Error: initializing source docker://quay.io/freeipa/freeipa-server:fedora-39: reading manifest fedora-39 in quay.io/freeipa/freeipa-server: unknown: Tag fedora-39 was deleted or has expired. To pull, revive via time machine"
    Error: initializing source docker://quay.io/freeipa/freeipa-server:fedora-39: reading manifest fedora-39 in quay.io/freeipa/freeipa-server: unknown: Tag fedora-39 was deleted or has expired. To pull, revive via time machine
    + rv=125
    + rm -rf -- /tmp/tmp.2XcUu1oo97
    ```

PR #956, opened 2024-11-28
Fix tag for freeipa-server for TLS-e adoption

bogdando:
The tag fedora-39 has expired, use fedora-40 instead


fmount:
FYI https://github.com/openstack-k8s-operators/install_yamls/pull/955

PR #958, opened 2024-11-29
README revision

jpodivin:
Adjust REAME to explicitly state that we have NETWORK_ISOLATION set to true by default, and that it's better to use openstack_deploy_wait target. 

This should save newcomers some head scratches.


karelyatin (commented):
/lgtm

jpodivin:
This is definitely outdated now, and I don't have time, nor up to date info, to update it.

PR #959, opened 2024-11-29
Fix missiing ifcfg device and bootproto in edpm gen

bogdando:
Whenever there is no device/bootproto setting present, edpm VM no longer takes expected IP address.

Add missing settings to make it working properly.


bogdando:
prolly not needed

PR #960, opened 2024-12-02
Add openstack_ga, openstack_crds_cleanup and openstack_deploy_update target

stuggi:
* The openstack_ga target allows to deploy the oldest available index of the openstack-operator using the commit from the stable branch configured using the OPENSTACK_STABLE_BRANCH environment variable, which defaults to 18.0-fr1.

* The openstack_crds_cleanup target allows to remove all openstack related CRDs from an environment. All deployments need to be deleted before

* The openstack_deploy_update target batches the OpenstackVersion target version to be the avaliable version to trigger an update.

With this an operator update from oldest available openstack-operator index of the 18.0-fr1 branch to the latest from main can be tested using this procedure:

```
rm -rf  out/operator/openstack-operator/
make openstack_ga
make openstack_deploy (wait for the deployment to finish) make openstack_cleanup
make openstack
make openstack_deploy_update
```

Related: https://github.com/openstack-k8s-operators/openstack-operator/pull/1217

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,9 +89,12 @@ GIT_CLONE_OPTS      ?=
 GALERA_REPLICAS         ?=
 
 # OpenStack Operator
-OPENSTACK_IMG                ?= quay.io/openstack-k8s-operators/openstack-operator-index:${OPENSTACK_K8S_TAG}
+OPENSTACK_REGISTRY           ?= quay.io
+OPENSTACK_REGISTRY_NAMESPACE ?= openstack-k8s-operators
+OPENSTACK_IMG                ?= ${OPENSTACK_REGISTRY}/${OPENSTACK_REGISTRY_NAMESPACE}/openstack-operator-index:${OPENSTACK_K8S_TAG}
 OPENSTACK_REPO               ?= https://github.com/openstack-k8s-operators/openstack-operator.git
 OPENSTACK_BRANCH             ?= ${OPENSTACK_K8S_BRANCH}
+OPENSTACK_STABLE_BRANCH      ?= 18.0-fr1
 OPENSTACK_COMMIT_HASH        ?=
 
 ifeq ($(NETWORK_ISOLATION), true)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -727,6 +730,18 @@ openstack: openstack_prep operator_namespace ## installs the operator, also runs
 	$(eval $(call vars,$@,openstack))
 	oc apply -f ${OPERATOR_DIR}
 
+.PHONY: openstack_ga
+openstack_ga: export BRANCH=${OPENSTACK_STABLE_BRANCH}
+openstack_ga: export REPO_DIR=${OPERATOR_BASE_DIR}/openstack-operator
+openstack_ga: export REGISTRY=${OPENSTACK_REGISTRY}
+openstack_ga: export NAMESPACE=${OPENSTACK_REGISTRY_NAMESPACE}
+openstack_ga: openstack_repo operator_namespace ## installs the operator for a feature branch using the first available index image matching a commit.
```
dprince, line 738: I'm not convinced we need this target. Why not just use OPENSTACK_IMG as the way to drive this. This allows us to have realistic targets when upgrading FROM and TO something. Realistic in the sense that we could use an index that contains actual bundles we used for a historical downstream release for example quay.io/dprince/openstack-operator-index:1.0.2 is what we released for 18.0.2. We obviously want to upgrade TO the latest HEAD of the upstream maintenance branches but still this could be driven by simply using OPENSTACK_IMG right?

dprince (changes requested):

Makefile
stuggi: yes with what we discussed yesterday its no longer needed. I'll update or submt a new pr later today

dprince: And for CI we could simply set the OPENSTACK_IMG being use as the baseline FROM target for upgrades in the .ci-operator.yaml file of the openstack-operator repo. This way we wouldn't have to change anything in openstack/release when modifying our baseline FROM target.

stuggi:
closing in favor of https://github.com/openstack-k8s-operators/install_yamls/pull/964

PR #961, opened 2024-12-03
[Adoption] Allow to deploy TripleO Manila with Ganesha

fmount:
This patch adds the `MANILA_NFS_ENABLED` environment variable and the associated THT.


fmount:
recheck

devsetup/standalone/openstack.sh
```diff
@@ -21,7 +21,8 @@ EDPM_COMPUTE_DHCP_AGENT_ENABLED=${EDPM_COMPUTE_DHCP_AGENT_ENABLED:-true}
 COMPUTE_DRIVER=${COMPUTE_DRIVER:-"libvirt"}
 INTERFACE_MTU=${INTERFACE_MTU:-1500}
 BARBICAN_ENABLED=${BARBICAN_ENABLED:-true}
-MANILA_ENABLED=${MANILA_ENABLED:-true}
+MANILA_CEPHFS_ENABLED=${MANILA_CEPHFS_ENABLED:-true}
+MANILA_NFS_ENABLED=${MANILA_NFS_ENABLED:-false}
```
gouthampacha, line 25: MANILA_CEPHNFS_ENABLED

probably? just to call out that this is actually Ceph-via-NFS

fmount: yeah, makes sense, thank you! I will update  it if we keep this patch up!

PR #964, opened 2024-12-04
Add openstack_crds_cleanup and openstack_patch_version target

stuggi:
* The openstack_crds_cleanup target allows to remove all openstack related CRDs from an environment. All deployments need to be deleted before.

* The openstack_patch_version target patches the OpenstackVersion target version to be the avaliable version to trigger an update.

With this an operator update (reinstall) and a deployment update can be tested using this procedure:

~~~
make openstack_crds_cleanup
OPENSTACK_IMG=quay.io/openstack-k8s-operators/openstack-operator-index:123 make openstack_wait
TIMEOUT=600s make openstack_wait_deploy
make openstack_cleanup
make openstack_wait
make openstack_patch_version
~~~

TODO:
add targets to perform edpm deploy/update (ovn controller and final service update)

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1217

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -798,6 +798,12 @@ openstack_update_run:
 	$(eval $(call vars,$@,openstack))
 	bash scripts/openstack-update.sh
 
+OV := $(shell oc get openstackversion -n $(NAMESPACE) -o name)
+.PHONY: openstack_patch_version
+openstack_patch_version: ## patches the openstackversion target version to the available version, if there is an update available
+	$(eval $(call vars,$@,openstack))
+	oc wait -n ${NAMESPACE} ${OV} --for=condition=MinorUpdateAvailable --timeout=${TIMEOUT} && \
+	oc patch -n ${NAMESPACE} ${OV} --type merge --patch '{"spec": {"targetVersion": "$(shell oc get -n $(NAMESPACE) ${OV} -o yaml | yq .status.availableVersion)"}}'
 
 .PHONY: edpm_deploy_generate_keys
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -798,6 +798,12 @@ openstack_update_run:
 	$(eval $(call vars,$@,openstack))
 	bash scripts/openstack-update.sh
 
+OV := $(shell oc get openstackversion -n $(NAMESPACE) -o name)
+.PHONY: openstack_patch_version
+openstack_patch_version: ## patches the openstackversion target version to the available version, if there is an update available
+	$(eval $(call vars,$@,openstack))
+	oc wait -n ${NAMESPACE} ${OV} --for=condition=MinorUpdateAvailable --timeout=${TIMEOUT} && \
```
dprince, line 805: I'm not sure we strictly need this new condition to implement upgrade testing. If we add it we'll have it on main, but not on the FR1 maintenance branch so we won't be able to use the job there.

dprince: It is enough to just check that the new catalog/index is deployed and that OpenStackVersion is in a ReadyCondition True state. Then if targetVersion != availableVersion you'd know there is an update (and this is shown clearly on the CLI too)

stuggi: as mentioned in the other PR, it just makes it a lot easier and it we can backport it. I'd say its a simple change and could be useful for any automation (also done by customers)

stuggi:
recheck

dprince (approved):

PR #965, opened 2024-12-05
Restoring expansion in interface scripts

jpodivin:
It's generally better to see scripts expanded. Both scripts were committed with the option commented out, I assume this was an oversight.

### DIFF
--- a/devsetup/scripts/crc-setup.sh
+++ b/devsetup/scripts/crc-setup.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# set -x
+set -x
 
 if [ "$EUID" -eq 0 ]; then
     echo "Please do not run as root."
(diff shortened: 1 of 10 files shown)


fao89 (approved):

devsetup/scripts/crc-setup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# set -x
+set -x
```
bogdando, line 2: [creds](https://github.com/openstack-k8s-operators/install_yamls/pull/965/files#diff-5fe68a3a8ef2686186052e6114a3d72939895df06382cc44ce9bb24ce4feeca4R70) and KUBEADMIN_PWD should not be echoed

PR #969, opened 2024-12-09
Make NNCP optional

hjensas:
In case the the NNCP was already applied the user may wan to skip applying the network configuration policy.


stuggi (approved):
/lgtm

PR #970, opened 2024-12-11
Add openstack_cleanup_all make target

weinimo:
Goal of this new make target is to clean up everything that may be necessary in order to re-run `make openstack` successfully if it failed before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -747,6 +747,10 @@ openstack_cleanup: operator_namespace## deletes the operator, but does not clean
 	oc delete catalogsource --all=true
 	test -d ${OPERATOR_BASE_DIR}/baremetal-operator && make crc_bmo_cleanup || true
 
+.PHONY: openstack_cleanup_all
+openstack_cleanup_all: openstack_cleanup  nncp_cleanup namespace_cleanup ## deletes the operator and its resources
+	@echo "Completed full cleanup."
+
 .PHONY: openstack_repo
 openstack_repo: export REPO=${OPENSTACK_REPO}
 openstack_repo: export BRANCH=${OPENSTACK_BRANCH}


Makefile
```diff
@@ -747,6 +747,10 @@ openstack_cleanup: operator_namespace## deletes the operator, but does not clean
 	oc delete catalogsource --all=true
 	test -d ${OPERATOR_BASE_DIR}/baremetal-operator && make crc_bmo_cleanup || true
 
+.PHONY: openstack_cleanup_all
+openstack_cleanup_all: openstack_cleanup  nncp_cleanup namespace_cleanup ## deletes the operator and its resources
```
stuggi, line 751: this won't work. openstack_cleanup will uninstall the operators. if they are down, you can not cleanup the openstack namespace.

stuggi (commented):
what is the goal for this target? because `make openstack` only installs the operators and its dependencies. it does not deploy the ctlplane. should it also cleanup the deployment, like `make openstack_deploy_cleanup`

Because of that the usual workflow is to run 
```
make openstack_deploy_cleanup
<wait for the deployemts gone>
make openstack_cleanup
make crc_storage_cleanup
...
```
there you could add the `nncp_cleanup` at the end and the `namespace_cleanup`. if you do not wait for the ctlplane resources cleanup up before uninstall the operators, you will have remaining objects with a finalizer. as a result you won't be able to delete the namespace

if you uninstall the operators, the CRDs are still on the cluster. if you also want to remove those, you'd have to run `openstack_crds_cleanup`.

metallb resource are in the `metallb-system` namespace. `metallb_config_cleanup` would remove those, and if you also want to uninstall that operator `metallb_cleanup`.

similar to cert-manager `certmanager_cleanup`

Makefile
weinimo: So I just need to switch the order?

PR #972, opened 2024-12-12
Fix tripleo install for multinode adoption testing

bogdando:
Escape the HOME env var as we want to refer to the target host' user home instead of the host where we generate the script.

Fix ContainerImageRegistryCredentials for tripleo. Quote the registry username as it may not work otherwise.


devsetup/scripts/tripleo.sh
```diff
@@ -121,7 +121,7 @@ if [ "\$RH_REGISTRY_USER" ] && [ -n "\$RH_REGISTRY_PWD" ]; then
     cat >> \$HOME/containers-prepare-parameters.yaml <<__EOF__
 ContainerImageRegistryCredentials:
```
bogdando, line 122: indentation...

fmount (approved):

bogdando:
recheck

bogdando:
please merge

fao89 (approved):

PR #973, opened 2024-12-12
Bump kuttl to 0.20.0

tosky:
It allows again for more granularity in JUnitXML reports. The other changes are mainly dependency bumps, plus an way to impersonate a different users. So no breaking changes are expected from this bump.


fmount (approved):
/lgtm

PR #975, opened 2024-12-13
Install metallb per default

stuggi:
With [1] there is a new CRD to manage pod secondary network interface IP announcements in a BGP environment. The controller for this relies on metallb FRRConfiguraton CRD. Therefore the operator needs to be deployed.

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/322

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -714,12 +714,12 @@ endif
 
 ##@ OPENSTACK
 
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

PR #984, opened 2024-12-18
openstack_wait: call 'make openstack' dynamically

dprince:
In some cases CI might end up calling 'make openstack' multiple times when using openstack_wait. This updates the Makefile dependency here to be dynamic and only call 'make openstack' if a subscription isn't found for the operator.


abays (approved):
/lgtm

PR #985, opened 2024-12-19
Fix post deployment hook for /etc/hosts in multi-stack

bogdando:
Add /etc/hosts entries for all cells' nodes on internal_api network as well. Enable rabbitmq connections from superconductor and scheduler via transport URLs defined for other cells'.

NOTE: tripleo_stack_name only used in start/end markers by the tripleo_hosts_entries role. Use a non-conflicting "stack" name for markers to append to hosts entries created for the overcloud stack, and stacks created for additional cells.

Required-by: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/746

### DIFF
--- a/devsetup/tripleo/tripleo_install.sh
+++ b/devsetup/tripleo/tripleo_install.sh
@@ -273,10 +273,13 @@ if [ $EDPM_COMPUTE_CELLS -gt 1 ] ; then
         done
     done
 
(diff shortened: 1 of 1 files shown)


slagle (approved):

bogdando:
Happy New Year, folks!

Can we have this merged, please? @fao89 @jistr @stuggi @olliewalsh

stuggi:
> Happy New Year, folks!
> 
> Can we have this merged, please? @fao89 @jistr @stuggi @olliewalsh

its already in merge pool

PR #995, opened 2025-01-08
Add internalapi routes for scenario A adoption

averdagu:
Doing data-plane adoption while testing scenario A (different subnets between wallaby and next-gen) will result in connectivity problem since pods will try to reach old CP using default route (since both CP are on different subnets) and won't be able to reach it.

Adding this route in internalapi to support data-plane adoption scenario A gates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,10 +51,9 @@ NETWORK_STORAGE_ADDRESS_PREFIX 	?= 172.18.0
 NETWORK_TENANT_ADDRESS_PREFIX 	?= 172.19.0
 NETWORK_STORAGEMGMT_ADDRESS_PREFIX 	?= 172.20.0
 NETWORK_DESIGNATE_ADDRESS_PREFIX 	?= 172.28.0
+# Additional routes that will be forwarded to nncp and net-attach-def
 INTERNALAPI_HOST_ROUTES ?=
-STORAGE_HOST_ROUTES ?=
 TENANT_HOST_ROUTES ?=
-STORAGEMGMT_HOST_ROUTES ?=
 
 # network isolation
 NETWORK_ISOLATION   ?= true
@@ -2395,6 +2394,8 @@ netattach: export DESIGNATE_PREFIX=${NETWORK_DESIGNATE_ADDRESS_PREFIX}
 netattach: export VLAN_START=${NETWORK_VLAN_START}
 netattach: export VLAN_STEP=${NETWORK_VLAN_STEP}
 netattach: export CTLPLANE_IP_ADDRESS_PREFIX=${NNCP_CTLPLANE_IP_ADDRESS_PREFIX}
+netattach: export NETATT_INTERNALAPI_HOST_ROUTES=${INTERNALAPI_HOST_ROUTES}
+netattach: export NETATT_TENANT_HOST_ROUTES=${TENANT_HOST_ROUTES}
 netattach: namespace ## Creates network-attachment-definitions for the networks the workers are attached via nncp
 	$(eval $(call vars,$@,netattach))
 	bash scripts/gen-netatt.sh
--- a/scripts/gen-netatt.sh
+++ b/scripts/gen-netatt.sh
@@ -120,8 +120,19 @@ if [ -n "$IPV4_ENABLED" ]; then
     cat >> ${DEPLOY_DIR}/internalapi.yaml <<EOF_CAT
         "range": "${INTERNALAPI_PREFIX}.0/24",
         "range_start": "${INTERNALAPI_PREFIX}.30",
-        "range_end": "${INTERNALAPI_PREFIX}.70"
+        "range_end": "${INTERNALAPI_PREFIX}.70",
 EOF_CAT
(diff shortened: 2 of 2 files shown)


scripts/gen-netatt.sh
```diff
@@ -122,6 +122,18 @@ if [ -n "$IPV4_ENABLED" ]; then
         "range_start": "${INTERNALAPI_PREFIX}.30",
         "range_end": "${INTERNALAPI_PREFIX}.70"
 EOF_CAT
+    # In the data-plane adoption scenario A (where different IP subnet ranges
+    # between next-gen and wallaby are used) the net-attach-def needs additional routes.
```
karelyatin, line 126: we should allow generic routes to be specified set with parameters, by default non will be set and by exporting env vars those could be set.

averdagu: Good idea, implementing it now

karelyatin: wanted to have this generic for all the networks not just internalapi
Also can't we reuse other existing vars *_HOST_ROUTES  instead of creating similar var?

Makefile
```diff
@@ -2308,6 +2308,7 @@ nncp_with_retries: ## Deploy NNCP with retries
 nncp: export INTERFACE=${NNCP_INTERFACE}
 nncp: export BRIDGE_NAME=${NNCP_BRIDGE}
 nncp: export INTERNALAPI_PREFIX=${NETWORK_INTERNALAPI_ADDRESS_PREFIX}
+nncp: export INTERNALAPI_ROUTE=${NETWORK_INTERNALAPI_ROUTE}
```
karelyatin, line 2311: looks unused

scripts/gen-netatt.sh
```diff
@@ -122,6 +122,17 @@ if [ -n "$IPV4_ENABLED" ]; then
...
+    if [ -n "$INTERNALAPI_HOST_ROUTE" ]; then
```
karelyatin, line 127: is this tested? as looks syntax error here with generated yaml as not joined with comma

karelyatin, line 127: INTERNALAPI_HOST_ROUTES

scripts/gen-netatt.sh
```diff
@@ -190,6 +201,17 @@ if [ -n "$IPV4_ENABLED" ]; then
         "range_start": "${TENANT_PREFIX}.30",
         "range_end": "${TENANT_PREFIX}.70"
 EOF_CAT
+    # In the data-plane adoption scenario A (where different IP subnet ranges
+    # between next-gen and wallaby are used) the net-attach-def needs additional routes.
+    if [ -n "$TENANT_HOST_ROUTE" ]; then
```
karelyatin, line 206: INTERNALAPI_HOST_ROUTES

scripts/gen-netatt.sh
```diff
@@ -190,6 +201,17 @@ if [ -n "$IPV4_ENABLED" ]; then
...
+    cat >> ${DEPLOY_DIR}/tenant.yaml <<EOF_CAT
+        "routes": [
+          {
+             "dst": "${TENANT_HOST_ROUTES}"
+          }
+        ]
+EOF_CAT
+    fi
```
karelyatin, line 214: what about storage and storage_mgmt networks?

karelyatin (changes requested):

scripts/gen-netatt.sh
averdagu: As per discussion a while back on nova channel, it is not possible to modify this two networks during dataplane adoption

scripts/gen-netatt.sh
averdagu: changed to NETATT_TENANT_HOST_ROUTES

scripts/gen-netatt.sh
averdagu: Yep, I changed the whole variable to `NETATT_INTERNALAPI_HOST_ROUTES`

averdagu: I thought it was, but I think i removed the depends-on on this change, after pulling all comments will trigger it again

Makefile
averdagu: True, deleted

scripts/gen-netatt.sh
karelyatin: The route in net-attach is generic, not specific to adoption, just needed for adoption case. so having it generic for all involved network is good.

karelyatin: CI fails as malformed json is generated, i think that could be handled by always adding the empty routes section, and when overrides are provided add to the routes section ?
Also i think let's be generic, i.e the patch should be Allow passing routes to all the network attachments including storage/storagemgmt etc. Similar to what being done for nncp. And remove the comments specific to adoption scenarios.
Dataplane scenario is just an use case of this support.

averdagu: Found out why json was malformated, handled already.
Also I added the ability to pass routes to all networks (storage/storagemgmt) as on the nncp section.
Still missing dataplane and designate. don't know if I should also add it

PR #996, opened 2025-01-15
Add WatcherPassword to osp-secret

amoralej:
It is needed for the watcher-operator-validation job to pass which is executed in watcher-operator gate.


abays (approved):
/lgtm

cescgina:
recheck failures don't seem related to the PR

PR #997, opened 2025-01-15
bootc cloud-init updates

slagle:
cloud-init cloud-config updates:
- Set passwords for root and cloud-admin users
- gateway4 key is deprecated, use routes instead

Also changes the script to run genisoimage every time instead of
checking if it already exists. It's not an expensive operation, and
since the cloud-init files are already overwritten every time, the iso
should be rebuilt every time as well to avoid confusion.


bshephar (approved):
Looks good, I've been doing this with the `config.toml` interface for the `bootc-image-builder`. But it's nice to consolidate around `cloud-init` given that's the mechanism leveraged by metal3 during the provisioning process.

PR #998, opened 2025-01-20
Ensure the openstacks crd exists in kuttl tests

cescgina:
Instead of just checking for any openstack.org crd, check explicitely
for openstacks.operator.openstack.org, and wait until that is available.
In some CI jobs, we run kuttl test for other operators before the
openstack-operators one. That means that the current check interprets
the fact that there are some openstack crds as being ready. Then when it
tries to get openstacks.operator.openstack.org, it fails because it's
not present yet.

Also remove the '|| true' at the end of the command, since that ignores
any error that might occurr.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2037,8 +2037,8 @@ openstack_kuttl_run: ## runs kuttl tests for the openstack operator, assumes tha
 openstack_kuttl: export NAMESPACE = ${OPENSTACK_KUTTL_NAMESPACE}
 openstack_kuttl: input deploy_cleanup openstack openstack_deploy_prep ## runs kuttl tests for the openstack operator. Installs openstack operator and cleans up previous deployments before running the tests, cleans up after running the tests.
 	# Wait until OLM installs openstack CRDs
-	timeout $(TIMEOUT) bash -c "while ! (oc get crd | grep openstack.org); do sleep 1; done"
-	bash -c '(oc get crd openstacks.operator.openstack.org && make openstack_init) || true'
+	timeout $(TIMEOUT) bash -c "while ! (oc get crd openstacks.operator.openstack.org); do sleep 1; done"
+	make openstack_init
 	$(eval $(call vars,$@,ansibleee))
 	make wait
 	$(eval $(call vars,$@,infra))


Makefile
```diff
@@ -2032,8 +2032,8 @@ openstack_kuttl_run: ## runs kuttl tests for the openstack operator, assumes tha
 openstack_kuttl: export NAMESPACE = ${OPENSTACK_KUTTL_NAMESPACE}
 openstack_kuttl: input deploy_cleanup openstack openstack_deploy_prep ## runs kuttl tests for the openstack operator. Installs openstack operator and cleans up previous deployments before running the tests, cleans up after running the tests.
 	# Wait until OLM installs openstack CRDs
-	timeout $(TIMEOUT) bash -c "while ! (oc get crd | grep openstack.org); do sleep 1; done"
-	bash -c '(oc get crd openstacks.operator.openstack.org && make openstack_init) || true'
```
abays, line 2036: I agree with removing this `|| true`.  We have seen it mask errors in CI.

abays: I think I remember now why the `|| true` was added in the first place.  The idea was to make this `openstack_kuttl` target backwards-compatible with older versions of OpenStack operator that did not have the new `OpenStack` initialization CRD.  i.e. the `oc get crd openstacks.operator.openstack.org` would return exit code 1 and thus be considered an error, so without the `|| true` the command would fail with the older OpenStack operator version and prevent the target from running.

The question becomes: do we need this target to work with older version of the OpenStack operator anymore?  The `main` branch of the OpenStack operator repo now has the `OpenStack` CRD.  Should the `main` branch of `install_yamls` be forced to assume it will always be working with this new install paradigm?

dprince: I think the main branch can be considered for FR2 right. We already have 18.0-fr1

Makefile
```diff
@@ -2032,8 +2032,8 @@ openstack_kuttl_run: ## runs kuttl tests for the openstack operator, assumes tha
...
+	timeout $(TIMEOUT) bash -c "while ! (oc get crd openstacks.operator.openstack.org); do sleep 1; done"
```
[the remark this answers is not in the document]
atyronesmith: $(TIMEOUT) is in the Makefile context which is correct.

Makefile
cescgina: thanks, @abays  that is simpler, done

bshephar (approved):

abays (approved):
/lgtm

PR #1000, opened 2025-01-22
Cleanup swift resources after glance kuttl tests

fmount:
When executing `glance` `kuttl` tests, we observed lingering `mutatingwebhookconfiguration` resources during the workflow execution. 
This can cause problems when the workflow involves both the `openstack` control plane deployment (using the new paradigm) and the usual `kuttl` deployment based on `OLM`.
This patch cleanup `swift` resources after `kuttl` is executed, with the effect of removing lingering swift `webhook` resources.


abays (approved):
/lgtm

PR #1003, opened 2025-01-28
Fix cephadm install when deploying an external ceph cluster

fmount:
Newer `cephadm` versions are not distributed as a single python script anymore.
This patch fixes the way we get this binary so we can rely on the official `Ceph` repositories (like we do in `devstack`) [1]

[1] https://github.com/openstack/devstack-plugin-ceph/blob/master/devstack/lib/cephadm#L166

### DIFF
--- a/devsetup/ceph/deploy.sh
+++ b/devsetup/ceph/deploy.sh
@@ -10,7 +10,10 @@ REQUIREMENTS=("jq" "lvm" "python3")
 
 # DEFAULT OPTIONS
 FSID="4b5c8c0a-ff60-454b-a1b4-9747aa737d19"
-CONTAINER_IMAGE=${CONTAINER_IMAGE:-'quay.io/ceph/ceph:v19'}
+CONTAINER_IMAGE=${CONTAINER_IMAGE:-'quay.io/ceph/ceph:v18'}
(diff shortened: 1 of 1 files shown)


fmount:
Tested locally and I can connect to an external Reef based Ceph cluster:

```
$ oc get CephCluster
NAME            DATADIRHOSTPATH   MONCOUNT   AGE   PHASE       MESSAGE                          HEALTH      EXTERNAL   FSID
rook-external                                5s    Connected   Cluster connected successfully   HEALTH_OK   true       4b5c8c0a-ff60-454b-a1b4-9747aa737d19

$ oc get sc
NAME                                     PROVISIONER                        RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
ceph-rbd                                 rook-ceph.rbd.csi.ceph.com         Delete          Immediate              true                   71s
crc-csi-hostpath-provisioner (default)   kubevirt.io.hostpath-provisioner   Retain          WaitForFirstConsumer   false                  55d
local-storage                            kubernetes.io/no-provisioner       Delete          WaitForFirstConsumer   true                   26h

```

fmount:
recheck

PR #1004, opened 2025-01-28
    Make baremetal targets more resilient to re-running

slagle:
This refactors some of the baremetal targets and scripts so that the
targets can be re-used without completely removing everything (resources
and instances) that were created on previous runs. This enables:

- creating different sets of edpm compute nodes, some preprovisioned and
  some baremetal, without the baremetal target completely deleting all
  edpm instances that existed previously.
- adding additional baremetal nodes to create different
  NodeSets/Deployments without deleting all previous resources.
- Setting specific baremetal node prefixes and suffixes to control the
  naming so that specific new nodes can be added without overwriting
  others.
- Removes over eager rm's, such as for out/edpm, which completely wipes
  out anything edpm related already deployed.

### DIFF
--- a/devsetup/Makefile
+++ b/devsetup/Makefile
@@ -127,7 +127,8 @@ BMAAS_ROUTE_LIBVIRT_NETWORKS ?= ${BMAAS_NETWORK_NAME},crc,default
 DATAPLANE_PLAYBOOK ?= osp.edpm.download_cache
 DATAPLANE_CUSTOM_SERVICE_RUNNER_IMG ?=quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
 BM_NETWORK_NAME            ?=default
-BM_INSTANCE_NAME_PREFIX    ?=edpm-compute
+BM_INSTANCE_NAME_PREFIX    ?=edpm-compute-baremetal
+BM_INSTANCE_NAME_SUFFIX    ?=0
 BM_INSTANCE_MEMORY         ?=8192
 BM_NODE_COUNT              ?=1
 BM_ROOT_PASSWORD_SECRET    ?=
@@ -379,6 +380,7 @@ edpm_baremetal_compute: export OPERATOR_NAME=openstack
 edpm_baremetal_compute: export NAMESPACE=${BMH_NAMESPACE}
 edpm_baremetal_compute: export BMAAS_NETWORK_NAME=${BM_NETWORK_NAME}
 edpm_baremetal_compute: export BMAAS_INSTANCE_NAME_PREFIX=${BM_INSTANCE_NAME_PREFIX}
+edpm_baremetal_compute: export BMAAS_INSTANCE_NAME_SUFFIX=${BM_INSTANCE_NAME_SUFFIX}
 edpm_baremetal_compute: export BMAAS_INSTANCE_MEMORY=${BM_INSTANCE_MEMORY}
 edpm_baremetal_compute: export BMAAS_NODE_COUNT=${BM_NODE_COUNT}
 edpm_baremetal_compute: export NODE_COUNT=${BM_NODE_COUNT}
@@ -392,13 +394,12 @@ edpm_baremetal_compute: ## Create virtual baremetal for the dataplane
 edpm_baremetal_compute_cleanup: export BMAAS_NETWORK_NAME=${BM_NETWORK_NAME}
 edpm_baremetal_compute_cleanup: export NAMESPACE=${BMH_NAMESPACE}
 edpm_baremetal_compute_cleanup: export BMAAS_INSTANCE_NAME_PREFIX=${BM_INSTANCE_NAME_PREFIX}
+edpm_baremetal_compute_cleanup: export BMAAS_INSTANCE_NAME_SUFFIX=${BM_INSTANCE_NAME_SUFFIX}
 edpm_baremetal_compute_cleanup: export BMAAS_NODE_COUNT=${BM_NODE_COUNT}
 edpm_baremetal_compute_cleanup: export NODE_COUNT=${BM_NODE_COUNT}
 edpm_baremetal_compute_cleanup:  ## Cleanup dataplane with BMAAS
 	$(eval $(call vars))
 	scripts/edpm-compute-baremetal.sh --cleanup
-	pushd .. && rm -Rf out/edpm || true && popd
-	make bmaas_sushy_emulator_cleanup || true
 	make bmaas_virtual_bms_cleanup || true
 
 .PHONY: edpm_compute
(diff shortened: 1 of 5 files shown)


steveb (commented):
/lgtm

devsetup/Makefile
```diff
@@ -377,25 +378,39 @@ edpm_baremetal_compute: export OPERATOR_NAME=openstack
...
 edpm_baremetal_compute_cleanup: export BMAAS_INSTANCE_NAME_PREFIX=${BM_INSTANCE_NAME_PREFIX}
+edpm_baremetal_compute_cleanup: export BMAAS_INSTANCE_NAME_SUFFIX=${BM_INSTANCE_NAME_SUFFIX}
 edpm_baremetal_compute_cleanup: export BMAAS_NODE_COUNT=${BM_NODE_COUNT}
 edpm_baremetal_compute_cleanup: export NODE_COUNT=${BM_NODE_COUNT}
 edpm_baremetal_compute_cleanup:  ## Cleanup dataplane with BMAAS
 	$(eval $(call vars))
-	scripts/edpm-compute-baremetal.sh --cleanup
-	pushd .. && rm -Rf out/edpm || true && popd
-	make bmaas_sushy_emulator_cleanup || true
+	if [ "${NODE_COUNT}" = "1" ]; then \
```
rabi, line 406: As we're changing the name of the vms and bmhs, cleanup won't work with things created with earlier version of the script. They have to be manually deleted.

devsetup/Makefile
```diff
@@ -377,25 +378,39 @@ edpm_baremetal_compute: export OPERATOR_NAME=openstack
...
+edpm_baremetal_compute: export BMAAS_INSTANCE_NAME_SUFFIX=${BM_INSTANCE_NAME_SUFFIX}
 edpm_baremetal_compute: export BMAAS_NODE_COUNT=${BM_NODE_COUNT}
 edpm_baremetal_compute: export NODE_COUNT=${BM_NODE_COUNT}
 edpm_baremetal_compute: export DEPLOY_DIR=../out/bmaas
 edpm_baremetal_compute: ## Create virtual baremetal for the dataplane
 	$(eval $(call vars))
 	make bmaas_virtual_bms
 	make bmaas_sushy_emulator_wait
-	bash scripts/edpm-compute-baremetal.sh --create
+	if [ "${NODE_COUNT}" = "1" ]; then \
```
rabi, line 389: Though you can now re-run to create additional BMHs, as we cleanup the existing OpenstackDataPlaneNodeSet/OpenStackDataPlaneDeployments with `edpm_deploy_cleanup`[1] that would deprovision the all existing BMHs as well, before provisioning them again.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L922

devsetup/Makefile
```diff
@@ -662,27 +677,42 @@ bmaas_metallb_cleanup:
...
 bmaas_virtual_bms: export DISK_SIZE = ${BMAAS_INSTANCE_DISK_SIZE}
 bmaas_virtual_bms: export OS_VARIANT = ${BMAAS_INSTANCE_OS_VARIANT}
 bmaas_virtual_bms: export VIRT_TYPE = ${BMAAS_INSTANCE_VIRT_TYPE}
 bmaas_virtual_bms: export NET_MODEL = ${BMAAS_INSTANCE_NET_MODEL}
 bmaas_virtual_bms: ## Create libvirt VM for BMaaS
-	scripts/bmaas/vbm-setup.sh --create
+	if [ ${NODE_COUNT} -eq 1 ]; then \
+		scripts/bmaas/vbm-setup.sh --create ; \
+	else \
+		for INDEX in $(shell seq 0 $$((${NODE_COUNT} -1))) ; do \
```
rabi, line 691: Calling these scripts as many times as the number of nodes is possibly little inefficient, though does not matter probably as no one would be deploying large number of nodes with install_yamls.

bshephar: Why do we need to iterate in the Makefile though? The `vbm-setup.sh` takes a `NODE_COUNT` env var and iterates within the script itself:
https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/bmaas/vbm-setup.sh#L166-L170

https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/bmaas/vbm-setup.sh#L145-L147


```
if [ ${NODE_COUNT} -eq 1 ]; then \
 		scripts/bmaas/vbm-setup.sh --create ; \
 	else
 	        scripts/bmaas/vbm-setup.sh --create --num-nodes $NODE_COUNT
fi
```

Then we can just handle the `NODE_NAME_SUFFIX` in the script? 

Just avoids re-running the script like Rabi is mentioning/

bshephar: I guess the whole check isn't really needed in that scenario actually, because `NODE_COUNT` would just be handled in the script anyway.

devsetup/Makefile
slagle: I think that would still be possible with using the env vars to set the names. However, I don't think we need to maintain backwards compatibility here anyway.

devsetup/Makefile
slagle: true. I would actually drop the cleanup deps from edpm_baremetal_deploy and edpm_deploy. I usually use the existing resources already written out, customize them as needed, and use them with oc directly. I guess it depends how far we want to take it in terms of making the targets work without cleanup or making them re-runnable.

devsetup/Makefile
slagle: This PR changes vbm-setup.sh to drop --num-nodes, so it manages just a single VM. That change is already included in this PR. Are you asking me to drop that change from the PR? The only inefficient part of calling the script twice would be setting up the libvirt logging, not really something I'd optimize in a dev repo, but we can go back to the way it was if that is wanted.

gen-edpm-node.sh manages 1 vm at a time. That is used for networker and bootc vms already, with the Makefile controlling the looping. So, I was changing the baremetal scripts to work in a similar way.

gen-edpm-node-bgp.sh is hardcoded to manage 3 at a time, so that's a different way as well.

Whichever way it's done, I'd like to see these things consistent.

slagle: @bshephar @rabi can you take another look at this one?

I made the requested changes so that the looping is still done in the script(s) now. I think this is an overall improvement given what it enables.

slagle: @hjensas @steveb as well.

devsetup/Makefile
```diff
@@ -379,6 +380,7 @@ edpm_baremetal_compute: export OPERATOR_NAME=openstack
 edpm_baremetal_compute: export NAMESPACE=${BMH_NAMESPACE}
 edpm_baremetal_compute: export BMAAS_NETWORK_NAME=${BM_NETWORK_NAME}
 edpm_baremetal_compute: export BMAAS_INSTANCE_NAME_PREFIX=${BM_INSTANCE_NAME_PREFIX}
+edpm_baremetal_compute: export BMAAS_INSTANCE_NAME_SUFFIX=${BM_INSTANCE_NAME_SUFFIX}
```
rabi, line 383: This sounds like the initial index used for new instances added. I assume if one has to add new instances they have to check the count of existing instances and set this accordingly which kind of not inituitive to me.

rabi (commented):
LGTM as an improvement though as earlier mentioned `edpm_deploy_baremetal` deprovisioning existing nodes (i.e calling edpm_deploy_cleanup target) would not help with the intended goal of reusability.

devsetup/Makefile
slagle: yes, that's how it works and I added some docs for that that. It's the same way the edpm_compute targets work. I think for developers, it's ok.

bshephar (approved):

PR #1010, opened 2025-02-04
Revert FIPS trunk.rdoproject.org workaround

steveb:
trunk.rdoproject.org now supports TLSv1.3 so this workaround is not longer required. We are not running upstream FIPS jobs anyway so the risk of regression here is low.

Jira: [RHOSZUUL-1517](https://issues.redhat.com//browse/RHOSZUUL-1517)


stuggi:
recheck

bshephar (approved):

hjensas (approved):

bshephar (approved):

PR #1011, opened 2025-02-06
Wait for webhook service after resource initialization

viroel:
Adds a `wait` for openstack-operator webhook service to be
created after resources initialization process. Some CI jobs
are failing to deploy OpenStack right after openstack_init target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -747,6 +747,7 @@ openstack_init: openstack_wait
 	bash -c 'test -f ${OPERATOR_BASE_DIR}/openstack-operator/config/samples/operator_v1beta1_openstack.yaml || make openstack_repo'
 	oc apply -f ${OPERATOR_BASE_DIR}/openstack-operator/config/samples/operator_v1beta1_openstack.yaml
 	oc wait openstack/openstack -n ${OPERATOR_NAMESPACE} --for condition=Ready --timeout=${TIMEOUT}
+	timeout ${TIMEOUT} bash -c "while ! (oc get services -n ${OPERATOR_NAMESPACE} | grep -E '^(openstack|openstack-baremetal|infra)-operator-webhook-service' | wc -l | grep -q -e 3); do sleep 5; done"
 
 .PHONY: openstack_cleanup
 openstack_cleanup: operator_namespace## deletes the operator, but does not cleanup the service resources


Makefile
```diff
@@ -747,6 +747,7 @@ openstack_init: openstack_wait
 	bash -c 'test -f ${OPERATOR_BASE_DIR}/openstack-operator/config/samples/operator_v1beta1_openstack.yaml || make openstack_repo'
 	oc apply -f ${OPERATOR_BASE_DIR}/openstack-operator/config/samples/operator_v1beta1_openstack.yaml
 	oc wait openstack/openstack -n ${OPERATOR_NAMESPACE} --for condition=Ready --timeout=${TIMEOUT}
+	timeout ${TIMEOUT} bash -c "while ! (oc get mutatingwebhookconfiguration | grep -E '^(openstack|infra)-operator-mutating-webhook-configuration' | wc -l | grep -q -e 2); do sleep 1; done"
```
abays, line 750: This can happen for the OpenStack operator webhooks too, so I think we would want to cover that case here as well.  Soon OpenStack Baremetal operator webhooks will be present as well, and this would need to be adjusted.  We are working on a more comprehensive solution, but fixes like these make sense in the meantime.

abays: Sorry, I misread your code.  I see that OpenStack operator is already included.  The problem, though, isn't the `MutatingWebhookConfiguration` resources themselves.  It's that the associated K8S `Service` hasn't been created yet.  That is what really needs to be examined here.

viroel: Right, it will work and be more simple too, let me update, thanks!

bshephar (approved):

viroel:
Updated to check for all 3 available  webhook services

Makefile
```diff
@@ -747,6 +747,7 @@ openstack_init: openstack_wait
 	bash -c 'test -f ${OPERATOR_BASE_DIR}/openstack-operator/config/samples/operator_v1beta1_openstack.yaml || make openstack_repo'
 	oc apply -f ${OPERATOR_BASE_DIR}/openstack-operator/config/samples/operator_v1beta1_openstack.yaml
 	oc wait openstack/openstack -n ${OPERATOR_NAMESPACE} --for condition=Ready --timeout=${TIMEOUT}
+	timeout ${TIMEOUT} bash -c "while ! (oc get services | grep -E '^(openstack|openstack-baremetal|infra)-operator-webhook-service' | wc -l | grep -q -e 3); do sleep 5; done"
```
abays, line 750:
```suggestion
	timeout ${TIMEOUT} bash -c "while ! (oc get services -n ${OPERATOR_NAMESPACE} | grep -E '^(openstack|openstack-baremetal|infra)-operator-webhook-service' | wc -l | grep -q -e 3); do sleep 5; done"
```

abays (approved):
/lgtm

abays (approved):
/lgtm

viroel:
recheck

last failure: `received unexpected HTTP status: 502 Bad Gateway` when pulling containers

viroel:
One job still failed with this validation [1]
Looking at my local env, it seems that the service endpoints is taking a few seconds to update with controller-manager pod IP address.
Looking at my timestamps:
```
kind: Service
metadata:
  creationTimestamp: "2025-02-06T20:24:47Z"
  name: openstack-operator-webhook-service
```
```
kind: Endpoints
metadata:
  annotations:
    endpoints.kubernetes.io/last-change-trigger-time: "2025-02-06T20:24:58Z"
  creationTimestamp: "2025-02-06T20:24:47Z"
  name: openstack-operator-webhook-service
```
so in my env, it took 11 seconds to update with pod's information.
so the validation would need to be more complex, checking if all 3 service endpoints have at least one pod ip configured.

[1] https://logserver.rdoproject.org/61/61/265b94350067b4775360052f8a99082ecbac34d4/github-check/watcher-operator-kuttl/cd93786/job-output.txt

viroel:
I am going to move with this one, and propose another patch and test it with more time. This one already help us a lot.

PR #1012, opened 2025-02-06
Allow passing extra args to kubectl-kuttl

karelyatin:
Can now pass arbitary args while running test, for example to run a specific test can run:-
make ovn_kuttl_run KUTTL_ARGS="--test ovn-config"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1221,7 +1221,7 @@ barbican_deploy: input barbican_deploy_prep ## installs the service instance usi
 
 .PHONY: barbican_deploy_validate
 barbican_deploy_validate: input namespace ## checks that barbican was properly deployed. Set BARBICAN_KUTTL_DIR to use assert file from custom repo.
-	kubectl-kuttl assert -n ${NAMESPACE} ${BARBICAN_KUTTL_DIR}/../common/assert_sample_deployment.yaml --timeout 180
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2030,7 +2030,7 @@ openstack_kuttl_run: ## runs kuttl tests for the openstack operator, assumes tha
 	    oc delete osctlplane --all --namespace ${NAMESPACE}; \
 		make crc_storage_cleanup_with_retries; \
 		make crc_storage_with_retries; \
-		kubectl-kuttl test --config ${OPENSTACK_KUTTL_CONF} ${OPENSTACK_KUTTL_DIR} --test $${test_dir}; \
+		kubectl-kuttl test --config ${OPENSTACK_KUTTL_CONF} ${OPENSTACK_KUTTL_DIR} --test $${test_dir}; \ $(KUTTL_ARGS)
```
bshephar, line 2033: `${KUTTL_ARGS}` is coming after a semi-colon on this line.

bshephar (changes requested):
Just that one line I pointed out, I think you'll need to move `${KUTTL_ARGS}` to before the semi-colon

Makefile
karelyatin: good catch, i just did sed at all places :)

karelyatin: Not adding at all to this target as already using --test arg in it.

karelyatin: @bshephar can you check again

bshephar (approved):

PR #1015, opened 2025-02-10
Add infra dependency to mariadb kuttl

fmount:
`infra-operator` now provides shared resources for most of the operators, including `Topology`, that is referenced by the services' operators CRs to inject a given `topologySpreadConstraints` or `Affinity` to their resulting `Pods`.
This patch simply adds the `infra` dependency to `mariadb`, to make sure that a `Topology CRD` is deployed and available for `kuttl` tests.


stuggi (approved):
/lgtm

fmount:
recheck

PR #1016, opened 2025-02-18
[crc] Set system network-mode

stuggi:
with [1] the default network-mode changed to user, which make crc deploments to fail with v2.47.0 and our config. Lets configure network-mode system in our deployment script per default.

[1] https://github.com/crc-org/crc/commit/22ae876990db2d981bb5c95b0a83c43d662a5b1a


omersch381:
It worked for me

slagle (approved):

PR #1020, opened 2025-03-04
Minor ipv6-nat64 fixes

steveb:
- sudo where appropriate
- use full context for chcon calls
- reduce the default of the size of SNO disk from 150G to 31G which aligns with CRC

### DIFF
--- a/devsetup/scripts/ipv6-nat64/sno.sh
+++ b/devsetup/scripts/ipv6-nat64/sno.sh
@@ -75,7 +75,7 @@ ARCH="${ARCH:-x86_64}"
 MEMORY="${MEMORY:-32768}"
 VCPUS="${VCPUS:-12}"
 OS_VARIANT="${OS_VARIANT:-fedora-coreos-stable}"
-DISK_SIZE="${DISK_SIZE:-150}"
+DISK_SIZE="${DISK_SIZE:-31}"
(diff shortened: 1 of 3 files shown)


devsetup/scripts/ipv6-nat64/sno.sh
```diff
@@ -75,7 +75,7 @@ ARCH="${ARCH:-x86_64}"
 MEMORY="${MEMORY:-32768}"
 VCPUS="${VCPUS:-12}"
 OS_VARIANT="${OS_VARIANT:-fedora-coreos-stable}"
-DISK_SIZE="${DISK_SIZE:-150}"
+DISK_SIZE="${DISK_SIZE:-31}"
```
bshephar, line 78: Is this possibly going to cause issues for some CI jobs since we use local-storage? Or do we think 31GB should be sufficient for all of our jobs?

bshephar (commented):
Mostly seems fine to me. I just wonder how much space is consumed by all of the local-storage PV's for a fully deployed environment on CRC?

Don't want to have some subset of CI jobs failing because they're filling up the 31GB of storage

devsetup/scripts/ipv6-nat64/sno.sh
steveb: I just want to align this with CRC. Jobs will be overriding this value for their requirements where necessary

PR #1029, opened 2025-03-20
Support a CRC equivalent installation of SNO

steveb:
This change adapts the IPv6 lab Single-node-Openshift script to support being installed in an IPv4-only environment which is as similar as possible to CRC. This allows development to occur against a SNO environment when required without bringing up a whole IPv6 lab.

Not counting download time SNO installs in 37 minutes vs 7 for CRC, so we can't simply switch to SNO in all scenarios, but making it easier to develop with is desirable in some circumstances.

CRC actually supports microshift OKD variants too. This change also adds a CRC_PRESET variable to allow switching CRC variants but this is left undocumented because microshift is missing components such as OLM which the development workflow requires.

### DIFF
--- a/devsetup/scripts/ipv6-nat64/sno.sh
+++ b/devsetup/scripts/ipv6-nat64/sno.sh
@@ -51,23 +51,42 @@ WORK_DIR="${WORK_DIR:-$HOME/.sno-workdir}"
 # OCP installer
 PULL_SECRET="${PULL_SECRET:-${HOME}/pull-secret.txt}"
 SSH_PUB_KEY="${SSH_PUB_KEY:-${HOME}/.ssh/id_rsa.pub}"
-OCP_VERSION="${OCP_VERSION:-latest-4.14}"
+OCP_VERSION="${OCP_VERSION:-latest-4.18}"
 OCP_MIRROR_URL="${OCP_MIRROR_URL:-https://mirror.openshift.com/pub/openshift-v4/clients/ocp}"
 OCP_ADMIN_PASSWD=${OCP_ADMIN_PASSWD:-12345678}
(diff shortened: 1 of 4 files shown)


devsetup/scripts/ipv6-nat64/sno.sh
```diff
@@ -351,11 +372,12 @@ function create {
     get_openshift_installer
     get_rhcos_live_iso
     create_install_iso
-    create_dnsmasq_config
+    if [ "${IP_VERSION}" = "v6" ]; then
```
hjensas, line 375: We need to set up something for DNS still?

When I tried this, I get this error:
```
INFO Waiting up to 20m0s (until 11:46AM CET) for the Kubernetes API at https://api.sno.lab.example.com:6443... 
ERROR Attempted to gather ClusterOperator status after wait failure: listing ClusterOperator objects: Get "https://api.sno.lab.example.com:6443/apis/config.openshift.io/v1/clusteroperators": dial tcp: lookup api.sno.lab.example.com on 127.0.0.53:53: no such host 
INFO Use the following commands to gather logs from the cluster 
INFO openshift-install gather bootstrap --help    
ERROR Bootstrap failed to complete: Get "https://api.sno.lab.example.com:6443/version": dial tcp: lookup api.sno.lab.example.com on 127.0.0.53:53: no such host 
ERROR Failed waiting for Kubernetes API. This error usually happens when there is a problem on the bootstrap host that prevents creating a temporary control plane. 
```

I think we just set for the IPv4 case?
```
address=/apps.sno.lab.example.com/${SNO_HOST_IP}
host-record=api.sno.lab.example.com,${SNO_HOST_IP}
host-record=api-int.sno.lab.example.com,${SNO_HOST_IP}
```

PR #1031, opened 2025-03-21
Workaround random issues in nmstate-handler setup

karelyatin:
In Downstream we randomly hitting issues where nmstate-handler setup fails due to missing scc rule in 'nmstate-handler' cluster role.

The scc rule creation is conditional[1], and if openshift-apiserver is not running or is being re created the condition evaluates to false and then nmstate-handler setup just stucks.

In order to workaround this adding a wait condition for api server pod.

[1] https://github.com/nmstate/kubernetes-nmstate/pull/1113

Related-Issue: [OSPCIX-554](https://issues.redhat.com//browse/OSPCIX-554)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2320,6 +2320,7 @@ else
 	oc apply -f ${OPERATOR_DIR}
 	timeout ${TIMEOUT} bash -c "while ! (oc get deployments/nmstate-operator -n ${NAMESPACE}); do sleep 10; done"
 	oc wait deployments/nmstate-operator -n ${NAMESPACE} --for condition=Available --timeout=${TIMEOUT}
+	timeout ${TIMEOUT} bash -c "while ! (oc wait pod -n openshift-apiserver -l apiserver=true --for condition=Ready); do sleep 10; done"
 	oc apply -f ${DEPLOY_DIR}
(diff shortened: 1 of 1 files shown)


averdagu:
LGTM, this could fix OSPCIX-554 which is happening quite frequently

danpawlik:
In the near future, when we will be starting using crc-cloud, there would be an option: https://review.rdoproject.org/r/plugins/gitiles/config/+/refs/heads/master/roles/prepare-crc-cloud/defaults/main.yaml#24 `enable_ramdisk` that will move etcd storage to ramdisk - with that the OpenShift should be stable.

bshewale (approved):

Makefile
```diff
@@ -2320,6 +2320,7 @@ else
 	oc apply -f ${OPERATOR_DIR}
 	timeout ${TIMEOUT} bash -c "while ! (oc get deployments/nmstate-operator -n ${NAMESPACE}); do sleep 10; done"
 	oc wait deployments/nmstate-operator -n ${NAMESPACE} --for condition=Available --timeout=${TIMEOUT}
+	timeout ${TIMEOUT} bash -c "while ! (oc wait pod -n openshift-apiserver -l apiserver=true --for condition=Ready); do sleep 10; done"
```
danpawlik, line 2323: 500s might not be enough to establish cluster after etcd crash, but let's try.

danpawlik (approved):

Makefile
karelyatin: with etcd crash i think we will not be able to deploy other things too, i just used the same timeout like other places.

PR #1033, opened 2025-03-25
Increase timeout for barbican

danpawlik:
It is not enough time to finish barbican tests in 180s.


danpawlik:
Issue was in other place.

PR #1037, opened 2025-03-27
Add set_slower_etcd_profile rule to Makefile

danpawlik:
The 'set_slower_etcd_profile' rule adjusts the etcd profile to 'Slower' . That would be mostly used in CI jobs to mitigate instability (e.g. OpenShift API restarts because etcd crashed earlier) or sometimes caused by slow disks (noisy neighbors or old hardware), to which etcd is highly sensitive.


abays (approved):
/lgtm

frenzyfriday (approved):

danpawlik:
recheck

PR #1039, opened 2025-03-28
Use a job instead of oc debug for 'make crc_storage'

abays:
Recent analysis of Prow jobs has revealed that most of our failures are due to this error (or the equivalent in `crc_storage_cleanup`):

```
Removing debug pod ...
error: unable to upgrade connection: container container-00 not found in pod oko-12-bdbms-master-0-debug-7b4gk_openstack-kuttl-tests
make[3]: *** [Makefile:613: crc_storage] Error 1 
```

This PR attempts to use a Kubernetes `Job` to create and clean-up the CRC storage instead, which will hopefully be more reliable.

### DIFF
--- a/scripts/create-pv.sh
+++ b/scripts/create-pv.sh
@@ -14,7 +14,12 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 set -ex
+
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+. "${SCRIPTPATH}/storage_common.sh"
+
 PV_NUM=${PV_NUM:-12}
+TIMEOUT=${TIMEOUT:-300s}
 
 released=$(oc get pv -o json | jq -r '.items[] | select(.status.phase | test("Released")).metadata.name')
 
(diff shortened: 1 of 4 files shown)


scripts/create-pv.sh
```diff
@@ -22,11 +26,13 @@ for name in $released; do
...
 if [ -z "$NODE_NAMES" ]; then
     echo "Unable to determine node name with 'oc' command."
     exit 1
 fi
 for node in $NODE_NAMES; do
-    oc debug $node -T -- chroot /host /usr/bin/bash -c "for i in `seq -w -s ' ' $PV_NUM`; do echo \"creating dir /mnt/openstack/pv\$i on $node\"; mkdir -p /mnt/openstack/pv\$i; done"
+    . "${SCRIPTPATH}/storage_apply.sh" "${node}" "create"
 done
+
+oc wait job -n openstack -l install-yamls.crc.storage --for condition=Complete
```
olliewalsh, line 38: IIUC default timeout is 30s, maybe worth extending it?

abays: Good point

abays: Done

scripts/storage_apply.sh
```diff
@@ -0,0 +1,75 @@
...
+OPERATION=${2:-"create"}
+
+oc delete -n openstack job "crc-storage-${NODE}" --ignore-not-found
+
+cat << EOF | oc apply -f -
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: crc-storage-${NODE}
+  namespace: openstack
```
dprince, line 28: should we run this in the openstack-operators namespace? Or maybe even in its own "openstack-crc-storage" namespace to keep it isolated. I typically run crc_storage very early and won't always have this pre-created.

abays:
recheck

dprince (approved):

PR #1043, opened 2025-04-02
Fix cleanup statements to reflect db name

vakwetu:
The DB name can change depending on the configuration.  We need to update the cleanup statements accordingly.


fmount (approved):
/lgtm

PR #1061, opened 2025-06-03
Increase openstack-update timeout

vyzigold:
I've seen a job failure, because of timeout when waiting for the openstackversions MinorUpdateDataplane condition, but in the must-gather (which happens some time after the failure), the openstackversions CR looks fine. This to me suggests, that there just wasn't enough time to finish the update.

The failed job is the `component-cloudops-edpm-update-rhel9-rhoso18.0-crc-telemetry-logging` on Jun 3rd.


stuggi (approved):
/lgtm

PR #1070, opened 2025-07-09
Set upstream DNS server in coreDNS instead of resolve.conf

olliewalsh:
Adding a static DNS server via nncp causes an API outage on crc. Just configure the upstream DNS in coreDNS instead.

Jira: [OSPRH-15370](https://issues.redhat.com//browse/OSPRH-15370)


dprince:
this looks very  nice

stuggi:
looks good to me!

PR #1075, opened 2025-07-17
Disable upstream DNS when NNCP_DNS_SERVER is explicitly empty

olliewalsh:


holser (approved):
It works as expected.

PR #1081, opened 2025-08-14
Fix comment on update_services to allow parsing by CI framework

jistr:
CI framework uses the `.PHONY: update_services` line to find and parse the names of make targets. The part after `.PHONY:` is stripped of whitespace and considered to be the name of the make target [1]. That means we can't put comments on this line, it should be on the target definition line below.

(I also tweaked the update targets' comments slightly for better readability.)

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/cddba525607e728620c021f067fa8041fd9c0ed0/plugins/modules/generate_make_tasks.py#L117

Resolves: https://issues.redhat.com/browse/OSPRH-19070

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -854,13 +854,13 @@ openstack_update_run:
 	$(eval $(call vars,$@,openstack))
 	bash scripts/openstack-update.sh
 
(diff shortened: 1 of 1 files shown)


jistr:
recheck

slagle (approved):

PR #1089, opened 2025-09-03
[telemetry_kuttl] Add dependency on ovn operator

karelyatin:
The Needed-By patch is adding scrape config for OVN Northd Metric
services and for that it needs to access OVNNorthd instance.

Needed-By: https://github.com/openstack-k8s-operators/telemetry-operator/pull/747
Related-Issue: https://issues.redhat.com/browse/OSPRH-12570


fmount (approved):
/lgtm

PR #1090, opened 2025-09-11
Use xorrisofs instead of genisoimage

slagle:
genisoimage is replaced by xorrisofs in recent RHEL versions (9.4+).
Although genisoimage is setup as an alternative for xorrisofs, just go
ahead and call xorrisofs directly.

The xorrisofs command could fail with the following error if the
cidata.iso output file is not writable:
libburn : SORRY : Failed to open device (a pseudo-drive) : Permission denied
This can be the case when the output file is owned qemu:qemu, which
happens when attached to a domain. Delete the file before generating a
new one to avoid this error.


slagle:
@rabi PTAL

PR #1095, opened 2025-10-02
Fix openstack_crds Make target

abays:
Removes `openstack_deploy_prep` from being a prerequisite for `openstack_crds`, as the former requires
`openstack_deploy_prep_cleanup`, which assumes that the CRDs are already installed (which causes `openstack_crds` to fail and defeats its intended purpose).

Nonetheless, I'm a bit wary of removing the `openstack_deploy_prep` prerequisite because I wonder if it was added for a specific CI scenario that needed to use `openstack_crds` and for some reason needed the prerequisite.  If anyone is aware of such a context, please let me know.


stuggi:
it was added in 6ff941f11e19584c61818f8f7a20d84255825b91 @dprince maybe you can remember?

PR #1102, opened 2025-10-14
Update video models to VGA

mrunge:
cirrus is not supported and has been removed from more recent versions of libvirt.


karelyatin:
recheck fresh

slagle (approved):

PR #1133, opened 2026-01-29
Change IP_ADRESS_SUFFIX to IP_ADDRESS_SUFFIX

danpawlik:
The variable miss a character, which raise a warning on spell checking. It was done using sed command:

    find devsetup -type f -exec sed -i 's/IP_ADRESS_SUFFIX/IP_ADDRESS_SUFFIX/g' {} +;


bogdando (changes requested):
I would rather issued a warning msg, if using the old var, and added the new var to work the same way. Or we might become surprised with a blast radius of this change

bogdando (commented):
LGTM

slagle (approved):

slagle:
recheck

bogdando (approved):

PR #1137, opened 2026-02-18
Add test-operator KUTTL integration

kstrenkova:
Add test-operator to install_yamls following the standard operator pattern (manila, horizon, cinder, etc).

Provides:
- Operator deployment: test_operator_prep, test_operator, test_operator_cleanup
- CR deployment: tes_operatort_deploy_prep, test_operator_deploy, test_operator_deploy_cleanup
- KUTTL tests: test_operator_kuttl, test_operator_kuttl_run

Usage: make test_operator_kuttl

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -496,6 +496,24 @@ TELEMETRY_KUTTL_CONF      ?= ${TELEMETRY_KUTTL_BASEDIR}/kuttl-test.yaml
 TELEMETRY_KUTTL_NAMESPACE ?= telemetry-kuttl-tests
 TELEMETRY_KUTTL_DIR       ?= ${TELEMETRY_KUTTL_BASEDIR}/${TELEMETRY_KUTTL_RELPATH}
 
+# Test Operator
+TEST_IMG             ?= quay.io/openstack-k8s-operators/test-operator-index:${OPENSTACK_K8S_TAG}
+TEST_REPO            ?= https://github.com/openstack-k8s-operators/test-operator.git
+TEST_BRANCH          ?= ${OPENSTACK_K8S_BRANCH}
+TEST_COMMIT_HASH     ?=
+ANSIBLETEST          ?= config/samples/test_v1beta1_ansibletest.yaml
+ANSIBLETEST_CR       ?= ${OPERATOR_BASE_DIR}/test-operator/${ANSIBLETEST}
+HORIZONTEST          ?= config/samples/test_v1beta1_horizontest.yaml
+HORIZONTEST_CR       ?= ${OPERATOR_BASE_DIR}/test-operator/${HORIZONTEST}
+TEMPEST              ?= config/samples/test_v1beta1_tempest.yaml
+TEMPEST_CR           ?= ${OPERATOR_BASE_DIR}/test-operator/${TEMPEST}
+TOBIKO               ?= config/samples/test_v1beta1_tobiko.yaml
+TOBIKO_CR            ?= ${OPERATOR_BASE_DIR}/test-operator/${TOBIKO}
+TEST_CR              ?= ${TEMPEST_CR}
+TEST_KUTTL_CONF      ?= ${OPERATOR_BASE_DIR}/test-operator/kuttl-test.yaml
+TEST_KUTTL_DIR       ?= ${OPERATOR_BASE_DIR}/test-operator/test/kuttl/tests
+TEST_KUTTL_NAMESPACE ?= test-operator-kuttl-tests
+
 # BMO
 BMO_REPO                         ?= https://github.com/metal3-io/baremetal-operator
 BMO_BRANCH                       ?= release-0.9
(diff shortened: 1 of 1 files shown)


kstrenkova:
recheck

kstrenkova:
recheck

kstrenkova:
recheck

kstrenkova:
recheck

kstrenkova:
recheck

Makefile
```diff
@@ -2708,6 +2751,55 @@ telemetry_kuttl: kuttl_common_prep ovn heat heat_deploy certmanager telemetry te
...
+# Set the value of $TEST_KUTTL_NAMESPACE if you want to run test-operator
+# kuttl tests in a namespace different than the default (test-operator-kuttl-tests)
+test_operator_kuttl: kuttl_common_prep test_operator test_operator_deploy_prep
+	$(eval $(call vars,$@,test))
+	make wait
+	make test_operator_kuttl_run
+	make deploy_cleanup
+	make test_operator_cleanup
+	make kuttl_common_cleanup
+
```
stuggi, line 2787: I think we should call `bash scripts/restore-namespace.sh` at the end. iirc its to make sure we end up in the openstack namespace as the default project after cleanup the kuttl namespace

Makefile
```diff
@@ -2708,6 +2751,55 @@ telemetry_kuttl: kuttl_common_prep ovn heat heat_deploy certmanager telemetry te
...
+.PHONY: test_operator_cleanup
+test_operator_cleanup: ## deletes the operator, but does not cleanup the service resources
+	$(eval $(call vars,$@,test))
+	bash scripts/operator-cleanup.sh
+	${CLEANUP_DIR_CMD} ${OPERATOR_DIR}
+
+.PHONY: test_operator_deploy_prep
+test_operator_deploy_prep: export KIND=Tempest
+test_operator_deploy_prep: export REPO=${TEST_REPO}
+test_operator_deploy_prep: export BRANCH=${TEST_BRANCH}
```
stuggi, line 2750: should we support the HASH, custom TEST_CR?

The test_operator_deploy_prep clones the repo and creates directories, but doesn't set up a CR for deployment. The deploy_cleanup target then tries to oc kustomize ${DEPLOY_DIR} | oc delete, which would fail or no-op since nothing was placed in ${DEPLOY_DIR} ?

Makefile
```diff
@@ -2708,6 +2751,55 @@ telemetry_kuttl: kuttl_common_prep ovn heat heat_deploy certmanager telemetry te
...
+test_operator_deploy_prep: test_operator_deploy_cleanup ## prepares CR for test-operator
+	$(eval $(call vars,$@,test))
+	mkdir -p ${OPERATOR_BASE_DIR} ${OPERATOR_DIR} ${DEPLOY_DIR}
+	bash scripts/clone-operator-repo.sh
```
stuggi, line 2755: we do not need similar to like others have, like keystone to support a custom CR ?

```
	cp ${KEYSTONEAPI_CR} ${DEPLOY_DIR}
```

stuggi (commented):
the commit uses `test_operator_*`, but the commit message is talking about `make test_kuttl`. All other operators use cinder_kuttl, like no operator in it. Was it because single test is similar to the shell test command? just using `test` might result in other issued?

Makefile
kstrenkova: Ack, added.

Makefile
kstrenkova: Now that I am looking at the other operators, I can see this pattern. I added the hash and the TEST_CR, but in a way that adds CR for every resource type test-operator supports. I used the `KIND=.*` pattern that is in OVN, but I will have to test if this will work in my test-operator change.

Makefile
kstrenkova: Yes, you are right, I added it.

kstrenkova:
> the commit uses `test_operator_*`, but the commit message is talking about `make test_kuttl`. All other operators use cinder_kuttl, like no operator in it. Was it because single test is similar to the shell test command? just using `test` might result in other issued?

Yes, precisely. When I used `test` it didn't work, so I changed everything to `test_operator` so they match. But I left the uppercase values with `TEST` only because it automatically creates these values in some way and I was getting error that `TEST_IMG` is not defined, etc. I could change it to `test_operator` and then others to `test_*` if that would look better to you also.

Thank you for the review! I will test it next week, as I am done with my hours for this week. I will definitely need to double check how I am doing things, but if you see anything wrong off the bat, please feel free to comment :D

Makefile
kstrenkova: Update: I have changed it so that only one resource type is tested at time, by default Tempest.

Makefile
```diff
@@ -2708,6 +2761,65 @@ telemetry_kuttl: kuttl_common_prep ovn heat heat_deploy certmanager telemetry te
...
+test_operator_deploy: input test_operator_deploy_prep ## installs the test-operator CRs using kustomize. Runs prep step in advance. Set TEST_CR to choose test type, TEST_REPO and TEST_BRANCH to deploy from a custom repo.
+	$(eval $(call vars,$@,test))
+	make wait
+	bash scripts/operator-deploy-resources.sh
+
+.PHONY: test_operator_deploy_cleanup
+test_operator_deploy_cleanup: ## cleans up the service instance, Does not affect the operator.
+	$(eval $(call vars,$@,test))
+	oc kustomize ${DEPLOY_DIR} | oc delete --ignore-not-found=true -f -
+	${CLEANUP_DIR_CMD} ${DEPLOY_DIR}
```
abays: I think normally we also clean up the operator dir here.  For example [1].  Is there some reason that pattern wasn't reused here?

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/ed029a6ff410c12b4ee9e8454a81dc3d20d679b3/Makefile#L1532

Makefile
```diff
@@ -2708,6 +2761,65 @@ telemetry_kuttl: kuttl_common_prep ovn heat heat_deploy certmanager telemetry te
...
+	$(eval $(call vars,$@,test))
+	bash scripts/operator-cleanup.sh
+	${CLEANUP_DIR_CMD} ${OPERATOR_DIR}
+
+.PHONY: test_operator_deploy_prep
+test_operator_deploy_prep: export KIND=.*
+test_operator_deploy_prep: export REPO=${TEST_REPO}
+test_operator_deploy_prep: export BRANCH=${TEST_BRANCH}
+test_operator_deploy_prep: export HASH=${TEST_COMMIT_HASH}
+test_operator_deploy_prep: test_operator_deploy_cleanup ## prepares CR for test-operator. Set TEST_CR to deploy a specific test type (defaults to Tempest).
```
abays, line 2752: Is there any way/need to customize the actual service image used in the CR?  I understand that the underlying `KIND` varies here, so maybe it would be a bit tricky to support customizing the embedded container image in the spec.  I'm just curious.  If you don't feel that it's even needed right now, that's fine.

Makefile
kstrenkova: You are right, I changed it to match the pattern.

Makefile
kstrenkova: I am not quite sure if there is a need per say. I would say, it can be added as a follow-up, as currently it does seem a bit tricky and I want to add just the base first.

abays (approved):
/lgtm

PR #1144, opened 2026-04-13
Restore rabbitmq-cluster-operator in infra_kuttl target

lmiccini:
The rabbitmq-migration and rabbitmq-coexistence kuttl tests need the upstream rabbitmq-cluster-operator CRD (rabbitmq.com/v1beta1) to be installed. Re-add the rabbitmq prerequisite alongside infra_rabbitmq_deploy so both sets of tests can run.


stuggi (approved):
/lgtm

PR #1145, opened 2026-04-14
Install cert-manager in the ovn-kuttl tests

slawqo:
With patch [1] ovn-operator introduced dependency on the cert-manager and requires it to be available in the kuttl tests environment.

[1] https://github.com/openstack-k8s-operators/ovn-operator/pull/541

Related: #[OSPRH-1922](https://redhat.atlassian.net/browse/OSPRH-1922)


abays (approved):
/lgtm

PR #1146, opened 2026-05-08
Make repo and branch configurable for repo-setup

rabi:
let's make them configurable for different repos and branches.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -426,6 +426,8 @@ GENERATE_SSH_KEYS				 ?= true
 DATAPLANE_EXTRA_NOVA_CONFIG_FILE                 ?= /dev/null
 DATAPLANE_SERVER_ROLE                            ?= compute
 DATAPLANE_TLS_ENABLED                            ?= true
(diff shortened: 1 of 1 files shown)


rabi:
recheck

Makefile
```diff
@@ -967,7 +969,7 @@ edpm_deploy_prep: edpm_deploy_cleanup openstack_repo ## prepares the CR to insta
 ifeq ($(GENERATE_SSH_KEYS), true)
 	make edpm_deploy_generate_keys
 endif
-	oc apply -f devsetup/edpm/services
+	sed -e "s/repo-setup current-podified -b antelope/repo-setup ${DATAPLANE_REPO_SETUP_REPO} -b ${DATAPLANE_REPO_SETUP_BRANCH}/" devsetup/edpm/services/*.yaml | oc apply -f -
```
slagle, line 972: this is fine, but we could use kustomize here like we do for a lot of other env var inputs.

slagle (approved):

PR #1147, opened 2026-05-13
Add $EDPM_COMPUTE_NETDEV for setting edpm interface name

slagle:
RHEL 10 has done away with the eth* naming, so a way to override the
device name is needed.

network-scripts is also removed from RHEL 10, so configuring those files
is made conditional on the existence of the
/etc/sysconfig/network-scripts dir.


slagle:
Now contained with https://github.com/openstack-k8s-operators/install_yamls/pull/1184

PR #1150, opened 2026-06-15
Allow override of BMO image

steveb:
This allows a container image of baremetal-operator to be sourced from other than the quay.io/metal3-io/baremetal-operator.

This will allow local validation of fixes such as [1]

[1] https://github.com/metal3-io/baremetal-operator/pull/3310


abays (approved):
/lgtm

PR #1151, opened 2026-06-16
Wait only on Running cert-manager pods during OLM install

danpawlik:
During cert-manager upgrades, stale Completed pods still match app=cert-manager and cause oc wait to time out.


danpawlik:
recheck

stuggi (approved):
/lgtm

PR #1149, opened 2026-06-03
Bump BMO to release-0.12 and IPA/Ironic to matching versions

rabi:
BMO release-0.9 is EOL. Update to release-0.12 (supported) and align Ironic image and IPA branch accordingly:
- BMO_BRANCH: release-0.9 -> release-0.12
- IRONIC_IMAGE_TAG: release-24.1 -> release-33.0
- BMO_IPA_BRANCH: stable/2024.1 -> stable/2025.2

Also, add option to change deploy interface for bootc.

PR #1155, opened 2026-07-10
Bump ansible-lint version

sdatko:
Currently we use a very old version of ansible-lint and there occurs an error while preparing the pre-commit environment for checks:
```
...
      error: can't copy 'src/ansiblelint/_vendor/ansible_compat': doesn't exist or not a regular file
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for ansible-lint
error: failed-wheel-build-for-install
```

That was caused by distribution model and usage of git submodules that apparently was removed in favor of a regular PyPI dependency [1].

After bumping the version, the CI reported incorrect rule skip from ansible-lint conf:
```
Rule 'syntax-check' is unskippable, you cannot use it in 'skip_list'
or 'warn_list'. Still, you could exclude the file.
```

The problematic entry we had in config file was never valid [2], however it just was not enforced until v6.13.0 [3].

The second commit here removes it in favor of relevant skip and necessary minor change in code.

Assisted-by: Claude Sonnet 4.5

[1] https://github.com/ansible/ansible-lint/pull/3478
[2] https://github.com/ansible/ansible-lint/issues/1594
[3] https://github.com/ansible/ansible-lint/pull/3020

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,6 +1,6 @@
 ---
 warn_list:
-  - syntax-check[specific]  # couldn't resolve module/action "foo"
+  - var-naming[no-role-prefix]
 exclude_paths:
   - devsetup/standalone/network_data.yaml
   - devsetup/standalone/deployed_network.yaml
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
         types: [file, yaml]
         entry: yamllint --strict -f parsable
   - repo: https://github.com/ansible/ansible-lint.git
-    rev: v6.12.2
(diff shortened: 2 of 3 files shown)


stuggi:
maybe could test what @rabi did in https://github.com/openstack-k8s-operators/install_yamls/pull/1149/changes

sdatko:
Thanks @stuggi – I checked there and indeed noticed the similar changes to make the test passing, although there it is mixed together with some functional change, so I propose this here as a separate change only related to CI.

stuggi (approved):
/lgtm

PR #1157, opened 2026-07-14
[openstack_init] Wait for openstack CR to be available

karelyatin:
Required to avoid below issue seen in CI:-
oc wait openstack/openstack -n openstack-operators --for condition=Ready --timeout=1200s Error from server (NotFound): openstacks.operator.openstack.org "openstack" not found

Related-Issue: https://redhat.atlassian.net/browse/OSPCIX-1300


abays (approved):
/lgtm

PR #1158, opened 2026-07-14
Replace hardcoded passwords and secrets with dynamic generation

abays:
All hardcoded passwords, encryption keys, and credentials are removed and replaced with randomly generated per-service secrets. A new scripts/gen-secrets.sh produces a gitignored .secrets.env file on first use, loaded automatically by Make. A legacy PASSWORD override is preserved for backward compatibility.


devsetup/Makefile
```diff
@@ -214,7 +218,7 @@ nfs_cleanup: ## removes /etc/exports.d/osp.exports, reloads nfs exports and dele
 
 ##@ CRC
 .PHONY: crc
-crc: ## Deploys CRC using CRC_URL to download and install CRC, KUBEADMIN_PWD as the password which defaults to 12345678 and PULL_SECRET to specify the file containing the pull secret, defaults to ${PWD}/pull-secret.txt. To change the default memory and/or cpus for the VM use `CPUS=X MEMORY=Y DISK=Z make crc`.
+crc: ## Deploys CRC using CRC_URL to download and install CRC, KUBEADMIN_PWD as the password (auto-generated via .secrets.env) and PULL_SECRET to specify the file containing the pull secret, defaults to ${PWD}/pull-secret.txt. To change the default memory and/or cpus for the VM use `CPUS=X MEMORY=Y DISK=Z make crc`.
```
stuggi: do we have to depend this in `../.secrets.env` to call the gen-secrets if it was not run before? should do it in other make targets, too, to make sure the secrets file got created? for crc I think one might call it with an empty kubeadmin pwd?

scripts/gen-secrets.sh
```diff
@@ -0,0 +1,105 @@
...
+IPA_ADMIN_PASSWORD ?= $(gen_short_hex 16)
+
+# MetalLB BGP peer password
+BGP_PEER_PASSWORD ?= $(gen_short_hex 8)
+
+# Loki S3 credentials
+LOKI_S3_ACCESS_KEY_ID ?= $(gen_password)
+LOKI_S3_ACCESS_KEY_SECRET ?= $(gen_password)
+EOF
+
```
stuggi, line 104: maybe it should restrict the permissions to the pwd file? `chmod 600 "${SECRETS_FILE}"`

devsetup/Makefile
abays: Yes, fair point.

scripts/gen-secrets.sh
abays: Good idea.

stuggi (approved):
/lgtm

PR #1161, opened 2026-07-20
Harden CRC storage Job: narrow hostPath and trim RBAC

abays:
Scope the `hostPath` mount from `/mnt` (entire mount tree) down to `/mnt/openstack` (the only subtree the `Job` operates on), and use `DirectoryOrCreate` to handle bootstrapping. Update the `ConfigMap` scripts to reflect the narrower mount point.

Remove pods/jobs CRUD from the `crc-storage` `Role` — the `SA` never makes Kubernetes API calls; all `oc` commands run from the user's shell, and the `Job` controller uses its own cluster-level permissions to manage pods.

### DIFF
--- a/scripts/storage_apply.sh
+++ b/scripts/storage_apply.sh
@@ -69,7 +69,7 @@ spec:
           name: crc-storage
         - name: node-mnt
           hostPath:
-            path: /mnt
-            type: Directory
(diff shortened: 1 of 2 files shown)


stuggi (approved):
/lgtm

PR #1163, opened 2026-07-20
Remove hardcoded PSA privileged enforcement and label-sync opt-out

abays:
Remove the `pod-security.kubernetes.io/enforce: privileged` label and the `security.openshift.io/scc.podSecurityLabelSync: "false"` label from generated namespaces.

OpenShift's label syncer automatically sets the PSA enforcement level based on the SCCs actually granted to `ServiceAccounts` in the namespace. Hardcoding `privileged` and disabling label-sync prevents the cluster from tightening enforcement if SCC requirements change. Letting the label syncer manage the PSA label aligns with upstream Kubernetes and OpenShift guidance.

### DIFF
--- a/scripts/gen-namespace.sh
+++ b/scripts/gen-namespace.sh
@@ -35,7 +35,4 @@ apiVersion: v1
 kind: Namespace
 metadata:
     name: ${NAMESPACE}
-    labels:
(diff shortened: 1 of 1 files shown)


stuggi (approved):
/lgtm

PR #1164, opened 2026-07-21
ipv6 SNO lab - Verify SSL when downloading openshift installer

hjensas:
The script used the curl -k (--insecure) flag in the command when downloading the openshift installer archive. This change removes this flag so that the certificate is verified.

Also, change the default Fedora cloud image, F39 has been archived - switch to F44.
    
Jira: [OSPRH-33058](https://redhat.atlassian.net/browse/OSPRH-33058)


hjensas:
recheck

rabi (approved):

abays (approved):
/lgtm

PR #1166, opened 2026-07-21
Hide MySQL root password from process table and Make output

abays:
Source the mariadb-operator's `mysql_root_auth.sh` script instead of passing `--password=` on the `mysql` command line in all `*_deploy_cleanup` targets. The script sets up authentication from the corresponding secret, keeping the password out of the mysql process's `/proc/*/cmdline` and Make's stdout.


dciabrin (dismissed):
I would prefer if we use:
. /var/lib/operator-scripts/mysql_root_auth.sh

Instead of setting MYSQL_PWD directly.

This script is provided by the mariadb operator; it fetches the up-to-date root password from the generated .my.cnf file in the container, and it sets MYSQL_PWD automatically.

It achieves the same goal of keeping the password out from the command line.

dciabrin:
recheck

stuggi (approved):
/lgtm

PR #1171, opened 2026-08-05
fix: default to TLS certificate verification in scripts

slagle:
Remove unconditional curl -k and --tls-verify=false from base OS
image downloads, mirror registry operations, and health checks.
All insecure patterns now default to secure behavior with opt-in
variables (CURL_INSECURE, MIRROR_INSECURE) for environments that
require self-signed certificate support.

- gen-edpm-node.sh, gen-edpm-node-bgp.sh: gate -k behind
  CURL_INSECURE (default: false)
- mirror-registry-setup.sh: try secure skopeo inspect first,
  gate --dest-tls-verify behind MIRROR_INSECURE
- mirror-registry-ca.sh: use --cacert with generated CA cert
  instead of -k
- gen-lvms-kustomize.sh: document why -k is acceptable for
  CRC liveness check

Jira: [OSPRH-32813](https://redhat.atlassian.net/browse/OSPRH-32813)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/scripts/gen-lvms-kustomize.sh
+++ b/scripts/gen-lvms-kustomize.sh
@@ -59,6 +59,9 @@ function restart_crc {
     sudo virsh start crc
     status=""
     while [ -z $status ]; do
+        # -k is acceptable here: this is a liveness check against the local CRC
+        # OCP API which uses a self-signed certificate. No content is downloaded
+        # or acted upon — we only check whether the API is reachable.
         if curl -k https://"${OCP_API}":6443; then
             status="ok";
         else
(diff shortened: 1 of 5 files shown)


scripts/gen-lvms-kustomize.sh
```diff
@@ -59,6 +59,9 @@ function restart_crc {
     sudo virsh start crc
     status=""
     while [ -z $status ]; do
+        # -k is acceptable here: this is a liveness check against the local CRC
```
cjeanner, line 62: Would `curl -kI https://...` be possible? If it's a liveness check, a HEAD should be sufficient. And adding `-L` would also follow redirect, if needed.

slagle: it might be, but that's unrelated to this patch

abays (approved):
/lgtm

PR #1173, opened 2026-08-13
Use the relative path for `.secrets.env`

amartyasinha:
oc login command is being executed from `install_yamls/devsetup` dir, while `.secrets.env` file is in `install_yamls` dir.

Thus it was failing with:

```
grep: .secrets.env: No such file or directory
Login failed (401 Unauthorized)
Verify you have provided the correct credentials.
```


abays (approved):
/lgtm

amartyasinha:
recheck

abays:
I think we can skip Zuul since it's just a doc change

PR #1176, opened 2026-08-21
Use oc with /readyz endpoint for lightweight readiness check

fultonj:
When CRC needs to be restarted `scripts/gen-lvms-kustomize.sh` uses `curl -k` for a livness check against the OCP API. This patch replaces curl with `oc get --raw /readyz`.

Jira: OSPRH-33099, [OSPRH-35191](https://redhat.atlassian.net/browse/OSPRH-35191)

Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/scripts/gen-lvms-kustomize.sh
+++ b/scripts/gen-lvms-kustomize.sh
@@ -59,10 +59,8 @@ function restart_crc {
     sudo virsh start crc
     status=""
     while [ -z $status ]; do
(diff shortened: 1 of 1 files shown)


fultonj:
This has been tested additional details in https://redhat.atlassian.net/browse/OSPRH-35191

fmount (approved):
/lgtm

PR #1181, opened 2026-09-08
Add a new openstack_versions_master make target

elfiesmelfie:
This make target fetches the latest tags for the s2i images and creates an openstackversions CR to deploy them.

It can be run before openstack deploy so that the control plane and dataplane services use the new containers

Assisted-By: Claude


abays:
Thanks for working on this! I was looking at the implementation and noticed there might be an alternative approach that leverages the existing `openstack_deploy` infrastructure.

## Alternative Approach: Integrate with openstack_deploy

The existing `openstack_deploy` flow uses kustomize to deploy multiple CRs together from `${DEPLOY_DIR}`. This could potentially be used for OpenStackVersion CRs as well.

### Implementation

**1. Add Makefile Variables (top of Makefile)**
```makefile
# OpenStackVersion
OPENSTACK_VERSION_CR             ?= config/samples/core_v1beta1_openstackversion_empty.yaml
CUSTOM_CONTAINER_IMAGES          ?= false
S2I_IMAGE_MAPPINGS_URL           ?= https://raw.githubusercontent.com/openstack-k8s-operators/s2i-openstack-containers/refs/heads/main/containers/image-mappings.yaml
S2I_IMAGE_MAPPINGS_BRANCH        ?= main
```

**2. Extend `openstack_deploy_prep` target**
```makefile
openstack_deploy_prep: openstack_deploy_cleanup openstack_repo
	cp ${OPENSTACK_CR} ${DEPLOY_DIR}
ifneq (${OPENSTACK_NEUTRON_CUSTOM_CONF},)
	cp ${OPENSTACK_NEUTRON_CUSTOM_CONF} ${NEUTRON_CUSTOM_CONF}
endif
ifeq ($(CUSTOM_CONTAINER_IMAGES), true)
	wget -O ${DEPLOY_DIR}/image-mappings.yaml ${S2I_IMAGE_MAPPINGS_URL}
	cp ${OPERATOR_BASE_DIR}/openstack-operator/${OPENSTACK_VERSION_CR} ${DEPLOY_DIR}
	DEPLOY_DIR=${DEPLOY_DIR} bash scripts/fetch-image-digests.sh ${DEPLOY_DIR}/image-mappings.yaml
endif
ifeq ($(NETWORK_ISOLATION_USE_DEFAULT_NETWORK), false)
	sed -i 's/192.168.122/${NNCP_CTLPLANE_IP_ADDRESS_PREFIX}/g' ${DEPLOY_DIR}/$(notdir ${OPENSTACK_CR})
endif
```

**3. Modify `fetch-image-digests.sh`**
- Populate the OpenStackVersion CR in `${DEPLOY_DIR}` with `customContainerImages`
- Remove the `oc apply` at the end (kustomize handles deployment)

Since `kustomization_add_resources()` auto-discovers all `.yaml` files in `${DEPLOY_DIR}`, the OpenStackVersion CR would be deployed alongside OpenStackControlPlane.

### Usage
```bash
# Deploy with custom s2i master images
CUSTOM_CONTAINER_IMAGES=true make openstack_deploy

# Or standalone with different branch
CUSTOM_CONTAINER_IMAGES=true S2I_IMAGE_MAPPINGS_BRANCH=stable/2024.2 make openstack_deploy
```

### Trade-offs

**Integrated Approach:**
- ✅ Follows existing `_prep`/`_deploy`/`_cleanup` patterns  
- ✅ Automatic cleanup via `openstack_deploy_cleanup`
- ✅ Works with kustomize infrastructure (namespacing, labels)
- ✅ Parameterizable via Make variables
- ⚠️ More complex - coupled with openstack deployment
- ⚠️ Can't apply images without redeploying control plane

**Standalone Target (current PR):**
- ✅ Simple, focused, single-purpose
- ✅ Can be run independently anytime
- ✅ Easy to understand what it does
- ⚠️ Doesn't follow existing patterns
- ⚠️ Manual cleanup needed
- ⚠️ Direct `oc apply` bypasses kustomize

Both approaches are valid - just wanted to share the alternative in case it's useful. Let me know your thoughts!

scripts/fetch-image-digests.sh
```diff
@@ -0,0 +1,158 @@
...
+
+echo ""
+
+if [ ${success_count} -eq 0 ]; then
+    echo "ERROR: No images were successfully fetched!"
+    exit 1
+fi
+
+# Save digests to a file
+DIGEST_FILE="/home/zuul/install_yamls/image-digests.env"
```
slagle, line 108: this can't be hardcoded for /home/zuul

Makefile
```diff
@@ -3050,3 +3050,12 @@ set_slower_etcd_profile:  ## that is a helper for the CI jobs, where OpenShift A
 	# need to wait until the etcd pod would apply new rules
 	sleep 60
 	timeout $(TIMEOUT) bash -c "while ! (timeout 5 oc get pods -n openshift-etcd -o jsonpath='{.items[*].status.phase}' | grep -qE '^Running'); do sleep 10; done"
+
+# update openstackversions for 19
+.PHONY: openstack_versions_master
+openstack_versions_master: export OV_MASTER_PATCHFILE=openstack_versions_master
+openstack_versions_master:
+	wget -O image-mappings.yaml https://raw.githubusercontent.com/openstack-k8s-operators/s2i-openstack-containers/refs/heads/main/containers/image-mappings.yaml
```
slagle, line 3058: Can we use curl instead?

wget isn't installed by default on most base/minimal installs. If it's required, I suggest adding it to the list of packages in devsetup/roles/download_tools/tasks/main.yaml

scripts/fetch-image-digests.sh
elfiesmelfie: I missed this when cleaning up the vibecoded starter script.

I'll remove this function, as it is not used.

Makefile
elfiesmelfie: wget is my go-to, but replacing with curl is fine.

elfiesmelfie:
> **Integrated Approach:**
> 
>     * ✅ Follows existing `_prep`/`_deploy`/`_cleanup` patterns
> 
>     * ✅ Automatic cleanup via `openstack_deploy_cleanup`
> 
>     * ✅ Works with kustomize infrastructure (namespacing, labels)
> 
>     * ✅ Parameterizable via Make variables
> 
>     * ⚠️ More complex - coupled with openstack deployment
> 
>     * ⚠️ Can't apply images without redeploying control plane
> 
> 
> **Standalone Target (current PR):**
> 
>     * ✅ Simple, focused, single-purpose
> 
>     * ✅ Can be run independently anytime
> 
>     * ✅ Easy to understand what it does
> 
>     * ⚠️ Doesn't follow existing patterns
> 
>     * ⚠️ Manual cleanup needed
> 
>     * ⚠️ Direct `oc apply` bypasses kustomize
> 
> 
> Both approaches are valid - just wanted to share the alternative in case it's useful. Let me know your thoughts!

Thank you for the detailed analysis. I think there's merit to what you're saying, but both methods have their place.
The current implementation I have can also be used standalone to bump versions during development, and IMHO is more intuitive to use.
However, integrating with the openstack_deploy is also useful, and makes sense.
I think the fetch-image-data.sh script can be modified to create either the kustomise or full patchfile, but that would take a bit more testing. I am not reallly familiar with kustomize, but I think the difference is in some of the top-matter in the patch-opensntack-versions.yaml, right?

There is also space to add additional params to the digest fetch script that would allow selecting the tags that get used.

I'll update this PR and add some cleanup, to burn down one of the disadvantages of standalone.
Then I can create a follow-up that modifies the script to allow using it in the standard workflow.
WDYT?

abays:
> > **Integrated Approach:**
> > ```
> > * ✅ Follows existing `_prep`/`_deploy`/`_cleanup` patterns
> > 
> > * ✅ Automatic cleanup via `openstack_deploy_cleanup`
> > 
> > * ✅ Works with kustomize infrastructure (namespacing, labels)
> > 
> > * ✅ Parameterizable via Make variables
> > 
> > * ⚠️ More complex - coupled with openstack deployment
> > 
> > * ⚠️ Can't apply images without redeploying control plane
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **Standalone Target (current PR):**
> > ```
> > * ✅ Simple, focused, single-purpose
> > 
> > * ✅ Can be run independently anytime
> > 
> > * ✅ Easy to understand what it does
> > 
> > * ⚠️ Doesn't follow existing patterns
> > 
> > * ⚠️ Manual cleanup needed
> > 
> > * ⚠️ Direct `oc apply` bypasses kustomize
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Both approaches are valid - just wanted to share the alternative in case it's useful. Let me know your thoughts!
> 
> Thank you for the detailed analysis. I think there's merit to what you're saying, but both methods have their place. The current implementation I have can also be used standalone to bump versions during development, and IMHO is more intuitive to use. However, integrating with the openstack_deploy is also useful, and makes sense. I think the fetch-image-data.sh script can be modified to create either the kustomise or full patchfile, but that would take a bit more testing. I am not reallly familiar with kustomize, but I think the difference is in some of the top-matter in the patch-opensntack-versions.yaml, right?
> 
> There is also space to add additional params to the digest fetch script that would allow selecting the tags that get used.
> 
> I'll update this PR and add some cleanup, to burn down one of the disadvantages of standalone. Then I can create a follow-up that modifies the script to allow using it in the standard workflow. WDYT?

Sorry for the delayed response.  Your proposed approach sounds good!

PR #1183, opened 2026-09-10
Pin RabbitMq PVC storage class via spec.persistence.storageClassName

lmiccini:
## Problem

The RabbitMq CR no longer exposes a top-level `spec.storageClass`; the storage class now lives under **`spec.persistence.storageClassName`** (infra-operator `apis/rabbitmq/v1beta1`, `DeprecatedPersistenceSpec.StorageClassName`). The generic `/spec/storageClass` patch in `scripts/gen-service-kustomize.sh` is therefore an unknown field that the API server **prunes** for `kind: RabbitMq`, so the RabbitMQ PVC silently falls back to the cluster **default** StorageClass instead of `${STORAGE_CLASS}`.

On OpenShift-on-OpenStack CI clusters the default is `standard-csi` (Cinder). Every other stateful component (galera, OVN db clusters, service backends) is pinned to `local-storage`, so RabbitMQ ends up as the **only** workload on Cinder CSI — the only one requiring a CSI `ControllerPublishVolume` attach step.

## Impact

A transient Cinder attach timeout is enough to wedge the whole run:

```
Warning FailedAttachVolume pod/rabbitmq-server-0
  AttachVolume.Attach failed for volume "pvc-...": rpc error: code = Internal
  desc = [ControllerPublishVolume] failed to attach volume: ... failed to be
  attached within the allotted time
```

→ `rabbitmq-server-0` stuck `Init:0/1` → RabbitMq/TransportURL never Ready → consumer operators (keystone, designate, …) loop forever `Waiting for TransportURL <svc>-transport secret to be created` → their `build-deploy-kuttl` suites fail (e.g. `designateBackendbind9ReadyCount: key is missing from map`, missing StatefulSets/Services/ConfigMaps). Observed across multiple operators' runs in the same window. The galera/OVN pods on `local-storage` are node-local and skip the attach path, so they were unaffected.

## Fix

Add a RabbitMq-specific kustomize patch that sets `spec.persistence.storageClassName=${STORAGE_CLASS}`, so RabbitMQ uses the same storage class as its siblings (e.g. `local-storage`) rather than the cluster default.

Verified locally with `kustomize build` against the infra-operator `rabbitmq_v1beta1_rabbitmq.yaml` sample — the rendered CR now contains:

```yaml
spec:
  persistence:
    storageClassName: local-storage
```

## Notes

- The generic `/spec/storageClass` patch is intentionally left unchanged — Galera, OVNDBCluster, Designate and other CRs still use top-level `spec.storageClass` legitimately.
- infra-operator's own sample (`config/samples/rabbitmq_v1beta1_rabbitmq.yaml`) still uses the removed top-level `storageClass` and should be updated separately to `spec.persistence.storageClassName`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts/gen-service-kustomize.sh
+++ b/scripts/gen-service-kustomize.sh
@@ -131,6 +131,20 @@ if [ "${KIND}" == "Galera" ]; then
 EOF
 fi
 
+# RabbitMq no longer exposes a top-level spec.storageClass; the storage class
+# now lives under spec.persistence.storageClassName. The generic
+# /spec/storageClass patch above is pruned by the API server for RabbitMq, so
+# without pinning it here the PVC silently falls back to the cluster default
+# StorageClass (e.g. Cinder standard-csi) instead of ${STORAGE_CLASS}.
+if [ "${KIND}" == "RabbitMq" ]; then
+   cat <<EOF>>kustomization.yaml
+    - op: add
+      path: /spec/persistence
+      value:
+        storageClassName: ${STORAGE_CLASS}
+EOF
+fi
+
 if [ "${KIND}" == "NetConfig" ]; then
 
     if [ -z "${IPV4_ENABLED}" ]; then
--- a/scripts/gen-service-kustomize.sh
+++ b/scripts/gen-service-kustomize.sh
@@ -142,6 +142,7 @@ if [ "${KIND}" == "RabbitMq" ]; then
       path: /spec/persistence
       value:
         storageClassName: ${STORAGE_CLASS}
+        storage: ${STORAGE_REQUEST}
 EOF
 fi
 


scripts/gen-service-kustomize.sh
```diff
@@ -131,6 +131,20 @@ if [ "${KIND}" == "Galera" ]; then
...
 
+# RabbitMq no longer exposes a top-level spec.storageClass; the storage class
+# now lives under spec.persistence.storageClassName. The generic
+# /spec/storageClass patch above is pruned by the API server for RabbitMq, so
+# without pinning it here the PVC silently falls back to the cluster default
+# StorageClass (e.g. Cinder standard-csi) instead of ${STORAGE_CLASS}.
+if [ "${KIND}" == "RabbitMq" ]; then
+   cat <<EOF>>kustomization.yaml
+    - op: add
+      path: /spec/persistence
```
stuggi, line 142: should also set storageRequest ?

lmiccini: Good catch — added `storage: ${STORAGE_REQUEST}` under `spec/persistence` in 51fa719, mirroring what Galera and the OpenStackControlPlane galera templates already do. Without it the PVC size fell back to the CRD default (10Gi) instead of the deployment-wide `STORAGE_REQUEST`. Verified the rendered CR now has both `persistence.storageClassName` and `persistence.storage`.

stuggi (approved):
/lgtm

danpawlik (approved):

PR #1184, opened 2026-09-10
Default to Centos Stream 10 for devsetup edpm nodes

slagle:
- **Default to Centos Stream 10 for devsetup edpm nodes**
- **Add $EDPM_COMPUTE_NETDEV for setting edpm interface name**


devsetup/scripts/gen-edpm-node-common.sh
```diff
@@ -285,8 +286,12 @@ fi
 
 # Set network variables for firstboot script
 IP=${IP:-"${EDPM_COMPUTE_NETWORK_IP%.*}.${IP_ADRESS_SUFFIX}"}
-NETDEV=eth0
-NETSCRIPT="/etc/sysconfig/network-scripts/ifcfg-${NETDEV}"
+if [ "${CENTOS_STREAM_VERSION}" = "9" ]; then
+    _netdev=eth0
+fi
+NETDEV=${EDPM_COMPUTE_NETDEV:-"${_netdev:-"enp2s0"}"}
```
SeanMooney: ah this is the biosdevname vs systemd stable devcie names vs legacy device name changes between 9 and 10

makes sense

slagle:
The rdo CI here is still using CS9 since that is configured in cifmw, and that is being updated here: https://github.com/openstack-k8s-operators/ci-framework/pull/4163

The install_yamls edpm devsetup isn't tested in CI but I have successfully tested this locally.

devsetup/scripts/gen-edpm-node-common.sh
abays, line 324: Does the device name here need to be something like `$NETDEV` instead?

slagle: Yes it does, thanks for catching that. It doesn't seem to have caused any problem, but I do see an error from this line in my firstboot script log. It's fixed in the next revision.

abays (approved):
/lgtm

slagle:
recheck

slagle:
recheck

tempest test failure:
https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//036/rdoproject.org/0364eb01177b43699ad34ae78adc7479/controller/ci-framework-data/tests/test_operator/tempest-tests-tempest/

PR #1186, opened 2026-09-18
Add Aetos password to osp-secret

vyzigold:
Also remove .secrets.env on crc_cleanup to regenerate the secrets on new deploy to get the new Aetos password.


devsetup/Makefile
```diff
@@ -227,6 +227,7 @@ crc_cleanup: ## Destroys the CRC env, but does NOT clear ( --clear-cache ) the c
 	crc cleanup
 	sudo ${CLEANUP_DIR_CMD} /etc/pki/ca-trust/source/anchors/crc-router-ca.pem
 	sudo update-ca-trust
+	rm -f ../.secrets.env
```
vyzigold: Are there any objections against this? Any reason to need to keep the .secrets.env across CRC deployments?

abays: Seems fine, but maybe we could do `$(MAKE) -C .. secrets_clean` (or something like that, whatever the exact syntax is) instead just to preserve a single source of truth for the secrets cleanup logic?

abays (approved):
/lgtm

PR #1187, opened 2026-09-18
Add DATAPLANE_USE_PARALLEL_EXECUTION to run EDPM services in parallel

rabi:
openstack-operator adds a useParallelExecution field on the OpenStackDataPlaneDeployment CR. When set, the data plane controller resolves the per-service dependsOn graph and deploys services level by level, running each level concurrently instead of strictly in list order.

ex.  DATAPLANE_USE_PARALLEL_EXECUTION=true make edpm_deploy

It defaults to false, so existing deployments keep the serial behaviour until the flag is set explicitly.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1918


rabi:
recheck

abays (approved):
/lgtm

